PR #296, opened 2022-02-09
295 cli secure string support

### LINKED ISSUE #295: CLI SecureString support
Lets support secure strings in the CLI user input interface.

This is amongst other things needed by process elevation on Linux (sudo).

### LINKED ISSUE #589: avoid downloading and installing tar.exe during Linux-Installer
The following script downloads tar from sourceforge. This seems to be a bit unstable. It would be better to create a repo with the exe file or somehow include it in the repository when we need to tar.
```
        run: |
          Invoke-WebRequest https://netcologne.dl.sourceforge.net/project/gnuwin32/tar/1.13-1/tar-1.13-1-bin.exe -Outfile tarsetup.exe
          Start-process .\tarsetup.exe -NoNewWindow -Wait -ArgumentList "/SILENT"

          $path=(get-item -path "./").Fullname
          $str=[IO.File]::ReadAllText($path + "/LinuxInstall/INSTALL.sh") -replace "`r`n", "`n"
          ([IO.File]::WriteAllText($path + "/INSTALL.sh", $str))

          $str=[IO.File]::ReadAllText($path + "/LinuxInstall/README") -replace "`r`n", "`n"
          ([IO.File]::WriteAllText($path + "/README", $str))

          cmd /C '"C:\Program Files (x86)\GnuWin32\bin\tar.exe" -cf "OpenTAP.tar" OpenTAP.*.Linux.TapPackage INSTALL.sh README'
      ```

### PR DESCRIPTION
**Pull request by rmadsen-ks**
Closes #295 

Closes #589

#589: Due to issues during build, I included a fix for this in that last commit. Seems win10 already has tar.exe installed!

### DIFF
--- a/Engine/UserInput.cs
+++ b/Engine/UserInput.cs
@@ -178,7 +178,7 @@ namespace OpenTap
                     }
                 }
             }
-            TapThread.Sleep(1000);
+            
             DateTime timeoutAt;
             if (timeout == TimeSpan.MaxValue)
                 timeoutAt = DateTime.MaxValue;
@@ -216,8 +216,8 @@ namespace OpenTap
                 if (display is DisplayAttribute attr && attr.IsDefaultAttribute() == false)
                     title = display.Name;
                 
-                // flush and make sure that there is no new log messages coming in (0.1s span).
-                Log.Flush(TimeSpan.FromSeconds(0.1));    
+                // flush and make sure that there is no new log messages coming in before starting to message the user.
+                Log.Flush();    
                 if (string.IsNullOrWhiteSpace(title))
                     // fallback magic
                     title = TypeData.GetTypeData(dataObject)?.GetMember("Name")?.GetValue(dataObject) as string;
--- a/Engine/UserInput.cs
+++ b/Engine/UserInput.cs
@@ -124,6 +124,7 @@ namespace OpenTap
     {
         readonly Mutex userInputMutex = new Mutex();
         readonly object readerLock = new object();
+
         void IUserInputInterface.RequestUserInput(object dataObject, TimeSpan timeout, bool modal)
         {
             if(readerThread == null)
@@ -137,6 +138,7 @@ namespace OpenTap
                             try
                             {
                                 var sb = new StringBuilder();
+                                
                                 while (true)
                                 {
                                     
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Review by alnlarsen — DISMISSED** [review]
This looks good! And it works like a charm on Linux. Let's see if we can get a similar implementation working on the TUI

**Comment by alnlarsen** [comment]
This works like a charm, however.. it is possible to delete the help text preceding the password, e.g. `Please enter Normal Input ():`

File: Engine/UserInput.cs [hunk 1]
```diff
@@ -132,12 +134,49 @@ void IUserInputInterface.RequestUserInput(object dataObject, TimeSpan timeout, b
...
+                                        Console.WriteLine();
+                                    }
+
+                                    if (chr.Key == ConsoleKey.Backspace)
+                                    {
+                                        if(sb.Length > 0)
+                                            sb.Remove(sb.Length - 1, 1);
+                                        
+                                        // delete current char and move back.
+                                        Console.Write("\b \b");
```
alnlarsen, line 162: It is possible to delete characters from the line which are not part of the current input. This doesn't affect the input, but it is a bit weird in terms of usability.

alnlarsen (changes requested):

alnlarsen (dismissed):
The new usability fixes are great! Let's merge!

Engine/UserInput.cs
```diff
@@ -175,6 +216,8 @@ void IUserInputInterface.RequestUserInput(object dataObject, TimeSpan timeout, b
                 if (display is DisplayAttribute attr && attr.IsDefaultAttribute() == false)
                     title = display.Name;
                 
+                // flush and make sure that there is no new log messages coming in (0.1s span).
+                Log.Flush(TimeSpan.FromSeconds(0.1));
```
AsgerIversen, line 220: I'm not sure I understand why a normal flush is not enough. I'm guessing we do this to avoid getting update info printed while we are waiting for user input. It looks to me like that would be covered by the userInputMutex (and a regular flush).

rmadsen-ks: Done! I don't think we actually needed that.

AsgerIversen (approved):

PR #303, opened 2022-02-10
Enhance TypeData logic to construct types with default constructors

### LINKED ISSUE #302: Allow TypeData to construct types with constructors that have default values for all parameters
```csharp
public class ConstructorWithDefaultValues
{
    public ConstructorWithDefaultValues(int a = 1)
    {
    }
}
```
There is no reason we should refuse to construct a type like this, and it is not that complicated to support.

### LINKED ISSUE #304: CanCreateInstance is set to true for types with private default constructors
e.g. this type:

```csharp
public class SomePluginClass
{
	private SomePluginClass(){}
}
```
will have its typedata's 'CanCreateInstance' return true because we don't check if constructors are public! This is an easy check to add and can save us some runtime exceptions.

alnlarsen:
This adds a check to the assembly reader so it will no longer consider private constructors when determining if an instance can be created from a typedata.

It also expands the type creation logic a bit so we now check if there is a constructor that has all default values. This should make the system a bit less restrictive to use, as you would previously have to create an empty constructor which called the other constructor with default values.

This is not a breaking change because we still prefer constructors with fewer parameters, so it should only affect types which did not previously have an applicable constructor.

Closes #302 
Closes #304

### DIFF
--- a/Engine.UnitTests/TestCreateInstance.cs
+++ b/Engine.UnitTests/TestCreateInstance.cs
@@ -1,0 +1,88 @@
+using System;
+using NUnit.Framework;
+
+namespace OpenTap.UnitTests
+{
+    public class ParameterlessConstructor
+    {
+    }
+    public class ConstructorWithDefaultValues
+    {
+        public ConstructorWithDefaultValues(int a = 1)
+        {
+            // This should be instantiated with the default value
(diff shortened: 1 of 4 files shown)


alnlarsen:
This feature should not be added

PR #360, opened 2022-02-21
359 reflection helper optimization

### LINKED ISSUE #359: Performance Optimization Opportunities in ReflectionHelper
```cs
IEnumerable<T> GetAttributes<T>(this IReflectionData mem)
```
Can be made faster, since the default case is that it only has one or zero elements of any given type. Currently it does an allocation every time it is called.

rmadsen-ks:
To test this, check out the first commit and run the profile cli action as described in #359.

Closes #359


alnlarsen (approved):
This looks good to me.

PR #430, opened 2022-03-07
Update TapEntry.cs

alnlarsen:


AsgerIversen (approved):

alnlarsen:
@AsgerIversen it looks like some pipeline stages are failing. We should probably configure what steps are required to pass for pull requests. I'm not sure how fine-grained control we have. If the worfkflow knows it isn't running within the OpenTAP organization, maybe we can skip steps that require organization secrets for example.

We should probably create an issue to at least discuss this.

g-galante:
In my understanding the workflow is running in the opentap Organization, as explained in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-events-for-forked-repositories-2:
> For pull requests from a forked repository to the base repository, GitHub sends the `pull_request`, `issue_comment`, `pull_request_review_comment`, `pull_request_review`, and `pull_request_target` events to the base repository. No pull request events occur on the forked repository.

And this is also confirmed by looking at the URL: https://github.com/opentap/opentap/actions/runs/2346130737 for this one.

It's just not being passed secrets because of https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories-2:
> With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository. The `GITHUB_TOKEN` has read-only permissions in forked repositories.

Check also https://github.com/opentap/opentap/issues/585#issuecomment-1129709011, which you have already noticed, looking at the reaction :)

alnlarsen:
Thanks for the input @g-galante. I guess #585 is fine for tracking this pipeline issue for now.

g-galante:
I ended up opening issue #595 and submitting PR #596 since I cannot work on #585, because I don't have write access to `opentap/opentap`.

PR #437, opened 2022-03-08
Rewrite assembly dependency resolution logic as CustomPackageAction

### LINKED ISSUE #436: Make assembly dependency resolution run later during package create
Currently, the assembly-level dependencies of plugins in a package are resolved before any ICustomPackageActions are run. This can cause issues because ICustomPackageActions are allowed to modify the DLLs directly (e.g. licensing and signing plugins). 

I propose rewriting this assembly dependency resolution as an ICustomPackageAction and assign it an Order so it runs pretty late in the package creation process (But still before the FileHash action) to support the use case where a third party plugin injects new code.

alnlarsen:
This changes the assembly-level dependency resolution logic to run much later in the package creation process which allows us to correctly handle the cases where a CustomPackageAction from a plugin has modified the content of managed assemblies.

Closes #436


alnlarsen:
It doesn't really look like this solves a problem anyone has.

PR #438, opened 2022-03-08
Use netstandard1.1 compatible DotNet.Glob dll instead of .NET 6.0

alnlarsen:
It seems bumping to .NET 6 has caused a regression in the NuGet package because we are shipping a framework dependent version of DotNet.Glob.


Denrask (changes requested):
I can't test this as Nuget Package job fails

Denrask (approved):

PR #442, opened 2022-03-09
Added CNAME file to public folder that gets committed to the pages repo

AsgerIversen:
Github pages can be configured to use a custom domain (get a cert for that domain):

image

However, that setting just results a commit adding a CNAME file to the root of the git repo. Next time the CI pipeline publishes the pages, this file gets overwritten (see [here](https://github.com/opentap/opentap.github.io/commit/60cf887937615db27f47dfea8027aa7eb22d5e57#diff-76b26edb75578663ca0fd6515ec7880ca04046bdbab298938981c9f0bd3907dd)).

Instead of using that setting in the opentap/opentap.github.io repo, manually create the CNAME file in this repo, before committing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,8 +130,6 @@ jobs:
         working-directory: doc
       - name: Copy API
         run: cp -r API public/api
-      - name: Add CNAME
-        run: echo doc.opentap.io > ./public/CNAME
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         if: startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -130,6 +130,8 @@ jobs:
         working-directory: doc
       - name: Copy API
         run: cp -r API public/api
+      - name: Add CNAME
```
StefanHolst, line 133: I think we could add this file to the .vuepress/public folder and then it would automatically be copied.

StefanHolst (changes requested):
Just a minor comment

PR #448, opened 2022-03-10
Use OpenTAP dll from nuget dir instead of output folder

### LINKED ISSUE #447: Unable to rebuild projects: Files are in use
We accidentally merged some code to the .csproj image installer which causes it to load OpenTap.dll and potentially other plugin DLLs from the output folder. This is an issue because C# IDEs keep MSBuild.exe and dotnet.exe open in the background, but these processes have now loaded OpenTAP assemblies preventing them from being overwritten! This causes builds to fail and can generally cause unexpected behavior when projects being built contain ICustomPackageAction implementations themselves.

alnlarsen:
This fixes a bug preventing projects from building when using the OpenTAP 9.17 nuget package.

CLoses #447


AsgerIversen (approved):
Looks good.

PR #450, opened 2022-03-11
449 Avoid Input to string blocking: fix

rmadsen-ks:
Fixed that input.ToString could block and while this is fine for Value, we should avoid doing time consuming work inside a ToString implementation.


alnlarsen:
Looks good to me.

alnlarsen (approved):

razvancoste:
Fixed in 9.17.3-rc.2

Issue #284, opened 2022-02-08
Downloading OpenTAP from documentation page doesn’t work  


ghost:
**Environment:**

Google Chrome, 97.0.4692.99 64-bit

https://docs.opentap.io/

[https://doc.opentap.io/](https://opentap.io/)

**Steps to reproduce:**

1. Access the docs.opentap.io
2. Go to the “Install OpenTAP”
3. Click on the “here” button to download OpenTAP

**Actual result:**

The link redirects to a 404 page

![2022-02-08 14_08_12-Getting Started _ OpenTAP](https://user-images.githubusercontent.com/98388827/153012561-1fcf76c9-cdc4-4595-83f4-a35015b6ce6c.png)

**Expected result:**

The link must work

**Other remarks:**

The correct download link is: https://opentap.io/downloads

The issue also reproduces for the pages:

https://doc.opentap.io/ 

https://forum.opentap.io/

![2022-02-08 17_07_42-OpenTAP Community Forum - Community discussion forums for OpenTAP](https://user-images.githubusercontent.com/98388827/153015060-7c153c2b-a8d1-44b4-bb44-8dacd66cd904.png)

StefanHolst:
@bdirenzo-keysight yes: https://github.com/opentap/opentap/pull/453

alnlarsen:
It looks like this is fixed both in the docs and the forums. Good work everyone.

alnlarsen:
This was fixed by #475

PR #453, opened 2022-03-14
#284 Fixed download link in docs

StefanHolst:
Updated link to old downloads page.


bdirenzo-keysight:
@StefanHolst there is a Linux download link that also points to the old link.

alnlarsen (approved):
Looks good

PR #457, opened 2022-03-14
Trim OS from Package XML to compare

### LINKED ISSUE #454: Packages with whitespace in OS is not properly being compared in OpenTAP
Given a package in PackageRepository that has OS="Windows, Linux", this is currently not OS Compatible with Linux.

OpenTAP needs to trim the entries here: https://github.com/opentap/opentap/blob/7b9de381877eff3bc161dd11e2834c347ebc1fcc/Package/IPackageIdentifier.cs#L67

Denrask:
Closes #454


AsgerIversen (approved):
Looks good

PR #459, opened 2022-03-14
PluginSearcher to respect .OpenTapIgnore

Denrask:
Ignore files that origins from a directory or subdirectory of a ".OpenTapIgnore" file entry.


Engine/AssemblyFinder.cs
```diff
@@ -0,0 +1,139 @@
...
+    class AssemblyFinder
+    {
+        static TraceSource log = Log.CreateSource("AssemblyFinder");
+        public void Invalidate()
+        {
+            lastSearch = DateTime.MinValue;
+        }
+
+        public bool IncludeDependencies = false;
+        public bool OutputLog = true;
```
rmadsen-ks, line 23: could this just be 'Quiet'? I think boolean properties should be false by default (when possible).

Shared/PathUtils.cs
```diff
@@ -131,5 +131,24 @@ public static string GetTempFileName(string extension)
...
+        /// <returns>The precense of OpenTapIgnore in parent chain</returns>
+        internal static bool DecendsFromOpenTapIgnore(string location)
+        {
+            string dir = Path.GetDirectoryName(location);
+            if (File.Exists(Path.Combine(dir, ".OpenTapIgnore")))
+                return true;
+            if (string.IsNullOrWhiteSpace(dir))
+                return false;
+            DirectoryInfo d = new DirectoryInfo(dir);
+            if (d.Parent == null)
```
rmadsen-ks, line 148: I dont think we need to add this check, isn't the final condition catched by string.IsNUllOrWhiteSpace(dir)?

maybe just add: if (string.IsNullOrWhiteSpace(dir) || dir == "/" || dir == location) return fals

Shared/PathUtils.cs
```diff
@@ -131,5 +131,24 @@ public static string GetTempFileName(string extension)
             //use instead of Path.GetTempFileName()
             return Path.Combine(Path.GetTempPath(), Path.GetRandomFileName() + extension);
         }
+
+        /// <summary>
+        /// Checks if the relative path has any OpenTapIgnore in parent chain
+        /// </summary>
+        /// <param name="location">Relative path of file</param>
+        /// <returns>The precense of OpenTapIgnore in parent chain</returns>
```
rmadsen-ks, line 139:
```suggestion
        /// <returns>Whether an .OpenTapIgnore file exists in folders.</returns>
```

rmadsen-ks (dismissed):
Minor change requests

Shared/PathUtils.cs
```diff
@@ -131,5 +131,21 @@ public static string GetTempFileName(string extension)
             //use instead of Path.GetTempFileName()
             return Path.Combine(Path.GetTempPath(), Path.GetRandomFileName() + extension);
         }
+
+        /// <summary>
+        /// Checks if the relative path has any ".OpenTapIgnore" file in parent directory chain. Throws argument error eventually if input is an absolute path and no ".OpenTapIgnore" file is present in parent directory chain.
+        /// </summary>
+        /// <param name="location">Relative path of file</param>
+        /// <returns>Whether an .OpenTapIgnore file exists in folders.</returns>
+        internal static bool DecendsFromOpenTapIgnore(string location)
```
alnlarsen, line 140: Should we prevent this method from recursing out of the install directory? `touch C:\.OpenTapIgnore` is going to brick everything ever and is difficult to debug

Denrask: This method explicitly states that it takes a relative path, so this should not be a problem. Examples of inputs: "Packages/REST-API/xxx.dll"

alnlarsen (changes requested):
I have one minor complaint, otherwise this looks fine to me.

alnlarsen (approved):

Shared/PathUtils.cs
Denrask: One thing I could do to make this clearer is the input argument to actually be named "relativePath" instead of "location". Would this help?

rmadsen-ks (approved):
Did a small amount of testing.  Looks good to me!

PR #463, opened 2022-03-15
464 Picture Annotation Improvement

### LINKED ISSUE #464: PictureAnnotation causes error when parameterizing
**Steps to reproduce:**

1. Add Dialog and Log Output steps
2. Paramterize Title and Message on Test Plan
3. Also parameterize Log Message on Test Plan
4. Merge Message and Log Message with Title so that Title will be the only one parameter displayed
5. Change Title value

rmadsen-ks:
Fixed issue occuring during parameterization , where the object source is not necessarily the right object type. It is better to just let the IObjectValueAnnotation handle getting and setting the object value.

Also enabled multi select for Pictures by overriding the Equals method.

Closes #464

### DIFF
--- a/Engine.UnitTests/CliCommandTest.cs
+++ b/Engine.UnitTests/CliCommandTest.cs
@@ -15,7 +15,7 @@ namespace OpenTap.UnitTests
     public class CliCommandTest
     {
         [Test]
-        public void test()
+        public void CliActionTreeTest()
         {
             var root = new CliActionTree();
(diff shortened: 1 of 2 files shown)


alnlarsen:
@rmadsen-ks it looks like tests are failing. Also, there are missing XML comments :)

rmadsen-ks:
Fixed the issue by adding comments.

alnlarsen (approved):
It looks like this works beautifully now ⚡ ⚡ ⚡

PR #465, opened 2022-03-16
Doc: .NET dependency reference file and IgnoreDependency

rmadsen-ks:
Update doc on .NET dll dependencies.


alnlarsen (approved):
Awesome! I love expanding our documentation!

PR #466, opened 2022-03-16
314 Change casing of UserID dir

### LINKED ISSUE #314: local/share on Linux has 2 folders named OpenTAP
**Environment:**

RHEL8

OpenTAP 9.17.0-beta.61

**Steps to reproduce:**

1. tap package install Demonstration
2. Check .local/share folder

**Actual result:**

image

(Please also attach a log file from %TAP_PATH%\SessionLogs\ that shows the issue when possible/relevant)

**Expected result:**

**Other remarks:**

alnlarsen:
This moves the 'OpenTapGeneratedId' into the same folder as the user-wide package cache on Linux, so the behavior is the same as on Windows.

This will cause Linux installations to get a new UserID, but this is not a huge issue. Runners and dockers get new images constantly, so this will make no difference.

The issue never existed on Windows because paths are not case sensitive. 

Closes #314

### DIFF
--- a/Package.UnitTests/PackageManagerTests.cs
+++ b/Package.UnitTests/PackageManagerTests.cs
@@ -61,7 +61,7 @@ namespace OpenTap.Package.UnitTests
         [Test]
         public void TestUserId()
         {
-            var idPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create), "OpenTAP", "OpenTapGeneratedId");
+            var idPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create), "OpenTap", "OpenTapGeneratedId");
             string orgId = null;
             if (File.Exists(idPath))
                 orgId = File.ReadAllText(idPath);
--- a/Package/Repositories/HttpPackageRepository.cs
+++ b/Package/Repositories/HttpPackageRepository.cs
@@ -96,7 +96,7 @@ namespace OpenTap.Package
         Action<string, long, long> IPackageDownloadProgress.OnProgressUpdate { get; set; }
         internal static string GetUserId()
         {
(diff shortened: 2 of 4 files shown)


rmadsen-ks (changes requested):
It seems that there are other places where this folder is being used.
```
./docker/Linux/Dockerfile:26:RUN echo 11111111-1111-1111-1111-111111111111 > /root/.local/share/OpenTAP/OpenTapGeneratedId
./docker/Windows/Dockerfile:10:RUN echo 11111111-1111-1111-1111-111111111111 > C:\Users\ContainerAdministrator\AppData\Local\OpenTAP\OpenTapGeneratedId
./Package/Repositories/HttpPackageRepository.cs:100:            var idPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create), "OpenTap", "OpenTapGeneratedId");
./Package.UnitTests/PackageManagerTests.cs:64:            var idPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create), "OpenTAP", "OpenTapGeneratedId");
```
maybe we should consolidate it?

rmadsen-ks (changes requested):
There are multiple references to OpenTapGeneratedId. Should these also be changed?

alnlarsen:
Yes!

rmadsen-ks (approved):
Looks good!

rmadsen-ks:
For the record, the reason we change OpenTAP -> OpenTap is that OpenTAP is only used for an installation id, whicle OpenTap contains a PackageCache. Hence if we change that we risk that the users leak that amount of harddrive space.

rmadsen-ks (approved):
Looks great!

rmadsen-ks:
Docker does no like this:

```
INFO[0043] args: [-c echo 11111111-1111-1111-1111-111111111111 > /root/.local/share/OpenTap/OpenTapGeneratedId] 
INFO[0043] Running: [/bin/sh -c echo 11111111-1111-1111-1111-111111111111 > /root/.local/share/OpenTap/OpenTapGeneratedId] 
/bin/sh: 1: cannot create /root/.local/share/OpenTap/OpenTapGeneratedId: Directory nonexistent
```

rmadsen-ks (approved):
Looks great.

PR #479, opened 2022-03-21
Add a new csproj target to verify installed packages

### LINKED ISSUE #470: Visual studio: Misbehaving failing image install
When an image install fails, it keeps retrying in an infinite loop. This is because the image install task checks if all the required package.xml files exist on the disk as a condition for running, but obviously the package.xml file does not exist if the image failed to install.

This can be pretty annoying for anyone using Visual Studio, as the errors will keep popping up until the image successfully resolves.

alnlarsen:
This fixes an issue where Visual Studio will run the install task in an infinite loop if it fails

This can be tested by downloading the nuget package and playing around with a .csproj file

Try adding and removing some 'AdditionalOpenTapPackage' or 'OpenTapPackageReference' elements and see if it behaves as expected.

Preferably test this in Visual Studio as this editor had serious issues before when packages failed to install.

Closes #470


Denrask (approved):
Tested and verified fix. Also experience performance optimised!

rmadsen-ks (approved):
Looks great! and works!

PR #503, opened 2022-04-06
Added Authentication CLI action

rmadsen-ks:
- Login - login to oauth instance
- Logout - logout from oauth instance. Including stopping the refresh token session.
- Refresh - refreshes the refresh token.

### DIFF
--- a/Engine/Authentication/AuthenticationSettings.cs
+++ b/Engine/Authentication/AuthenticationSettings.cs
@@ -123,7 +123,7 @@ namespace OpenTap.Authentication
             RefreshTokens.RemoveIf(x => x.Domain == domain);
         }
 
-        TokenInfo GetTokenInfoForDomain(string domain)
+        TokenInfo GetAccessTokenForDomain(string domain)
         {
             return AccessTokens.OrderByDescending(x => x.Domain.Length).FirstOrDefault(x => domain.EndsWith(x.Domain));
         }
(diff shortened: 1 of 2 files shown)


Denrask (changes requested):
I have a few comments. However, I still must express my deepest resistance against treating ICliActions as an acceptable C# API.

I'm going to be using the Login and Refresh functionality and even though exceptions are now thrown I can't disregard that this returns an integer, which I must always consider in my code. Furthermore, these CliActions are outputting log messages, which I would like to have more control over. I'm not keen on getting x amount of "Token refreshed." messages every minute.

Engine/Cli/LogoutAction.cs
```diff
@@ -0,0 +1,52 @@
...
+        
+        /// <summary> The domain to log out of.</summary>
+        [CommandLineArgument("domain")] public string Domain { get; set; }
+        
+        /// <summary> Executes the action. </summary>
+        public int Execute(CancellationToken cancellationToken)
+        {
+            this.MustBeDefined(nameof(Domain));
+            var token = AuthenticationSettings.Current.RefreshTokens.FirstOrDefault(x => x.Domain == Domain);
+            if(token != null){
```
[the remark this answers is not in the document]
rmadsen-ks: If it is null, it just meant that only an access token is available. Maybe we should throw an exception if there's neither?

Engine/Authentication/AuthenticationSettings.cs
```diff
@@ -122,14 +123,21 @@ public void UnregisterRefreshToken(string domain)
             RefreshTokens.RemoveIf(x => x.Domain == domain);
         }
 
+        TokenInfo GetAccessTokenForDomain(string domain)
+        {
+            return AccessTokens.OrderByDescending(x => x.Domain.Length).FirstOrDefault(x => domain.EndsWith(x.Domain));
```
alnlarsen, line 128: Why are we accepting any domains with the same ending?

Engine/Authentication/AuthenticationSettings.cs
```diff
@@ -122,14 +123,21 @@ public void UnregisterRefreshToken(string domain)
...
         {
-            cancel.ThrowIfCancellationRequested();
-
-            // Try refresh
-            
-            return AccessTokens.FirstOrDefault(x => x.Domain == domain);
+            var access = GetAccessTokenForDomain(domain);
+            if (access?.Expired == true)
+            {
+                var refresh = new RefreshTokenAction { Domain = domain };
```
alnlarsen, line 137: This looks like a weird place to refresh a token?

alnlarsen (approved):
I have a few questions, but I think @Denrask is more suited to review this API.

PR #514, opened 2022-04-19
Add progress update when system-wide packages start installing

### LINKED ISSUE #513: Package Installation status is "Gathering dependencies" while installing system-wide packages
The code which updates the current installation status to "Installing" is fired after system-wide packages have finished installing, and the installation status is therefore currently misleading.

alnlarsen:
Raise an update event when system-wide packages start installing

Closes #513


rmadsen-ks (approved):
Looks good!

PR #524, opened 2022-04-21
508-externalParameterDeserializeError

frederikja163:
Added test to prove the issue
Added other tests to TypeData, aswell as ObjectCloner.Clone
ObjectCloner.Clone has been updated to throw an InvalidCastException if TryClone returns false.
ObjectCloner.TryClone has been fixed so it can skip serialization and still succeed the cloning if the value is assignable to the targetType.

### DIFF
--- a/Engine.UnitTests/SerializerTest.cs
+++ b/Engine.UnitTests/SerializerTest.cs
@@ -1669,6 +1669,7 @@ namespace OpenTap.Engine.UnitTests
 
             public override void Run()
             {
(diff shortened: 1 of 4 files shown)


rmadsen-ks (dismissed):
This is perfect!

rmadsen-ks:
Unfortunately it seems like RunParseTest is failing. I guess we need to fix that.

rmadsen-ks (approved):
Things seems to work well now. Lets merge it.

PR #526, opened 2022-04-28
209-parallelDeadlock

### LINKED ISSUE #209: If verdict step should not be allowed to bind to it's parent
Originally filed January 12 2022 by Mikko Ala-Kojola on [GitLab](https://gitlab.com/OpenTAP/opentap/-/issues/826)

**Environment:**
OpenTAP 9.16.2

**Steps to reproduce:**
1. Add Parallel step into a plan
2. Add If verdict step as Parallel steps child
3. Set Parallel steps verdict into if verdict steps If parameter
4. Run the plan

**Actual result:**
Dead lock is produced.

**Expected result:**
If verdict should not allow verdict of it's own parent be placed as If parameter value. OpenTAP should not let to proceed into step 4.

**Other remarks:**
I raising this issue on behalf of colleague Teemu Luukkonen.
I used Parallel step as an example here as it's verdict is allowed in If verdicts If parameter. Not all OpenTAP test steps do.
But wanting to extend the limitation to all parents. Parent is not finished until all of it's childs are finished.

@witold wanted to be tagged into this. So that is now covered.

frederikja163:
Added deadlock detection for when a childstep is waiting for one of its parent steps.
Added unit tests with 1,2 and 10 levels of parallel nesting.

Closes #209

### DIFF
--- a/Engine/TestStep.cs
+++ b/Engine/TestStep.cs
@@ -844,6 +844,7 @@ namespace OpenTap
                 step.Verdict = newVerdict;
         }
 
+        /// <summary> This is the currently executing test step or null, used to detect deadlock when a step is waiting for its parent. </summary>
         [ThreadStatic]
         internal static ITestStep currentlyExecutingTestStep = null;
 
--- a/Engine/TestStep.cs
+++ b/Engine/TestStep.cs
@@ -844,6 +844,9 @@ namespace OpenTap
                 step.Verdict = newVerdict;
         }
 
+        [ThreadStatic]
+        internal static ITestStep currentlyExecutingTestStep = null;
+
         internal static TestStepRun DoRun(this ITestStep Step, TestPlanRun planRun, TestRun parentRun, IEnumerable<ResultParameter> attachedParameters = null)
         {
             {
@@ -868,6 +871,8 @@ namespace OpenTap
                 TestStepPath = Step.GetStepPath(),
             };
 
+            var previouslyExecutingTestStep = currentlyExecutingTestStep;
+            currentlyExecutingTestStep = Step;
             var stepPath = stepRun.TestStepPath;
             //Raise an event prior to starting the actual run of the TestStep. 
             Step.OfferBreak(stepRun, true);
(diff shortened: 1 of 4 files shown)


Engine/TestStepRun.cs
```diff
@@ -330,20 +330,31 @@ internal void ThrowDueToBreakConditions()
...
+        {
+            ITestStep currentStep = TestStepExtensions.currentlyExecutingTestStep;
             var currentThread = TapThread.Current;
             switch (mode)
             {
                 case OutputAvailability.BeforeRun: 
                     return;
                 case OutputAvailability.AfterDefer:
-                    if (StepThread == currentThread && runDone.Wait(0) == false)
+                    if ((StepThread == currentThread && runDone.Wait(0) == false) || IsStepChildOf(currentStep, waitingFor))
```
rmadsen-ks, line 352: We need to check if currentStep is null.

Engine/TestStep.cs
```diff
@@ -844,6 +844,9 @@ internal static void UpgradeVerdict(this ITestStep step, Verdict newVerdict)
                 step.Verdict = newVerdict;
         }
 
+        [ThreadStatic]
+        internal static ITestStep currentlyExecutingTestStep = null;
```
rmadsen-ks, line 848: Let's add a comment stating what this is for.

rmadsen-ks (changes requested):
Looks great! Just a couple of comments.

Engine/TestStepRun.cs
frederikja163: I'm not sure how it is possible for currentStep to be null here, `WaitForOutput` only gets called from `DoRun`, where it is always set to a non null value before it is being run.
Even if we add a null check for the future im not even sure what should be done, how can a step wait for anything if the currently executing step is null?

rmadsen-ks: I think you have a good point. I can some up with two cases.

What if the test step starts a new thread, which waits for the input? Couldn't the test plan complete before the thread does it thing? 

Also user interfaces may read these values asynchronously, which could occur during the ru, but the thread static value is always null seen from the UI thread.

frederikja163: Added a null check as requested

Engine/TestStep.cs
frederikja163: Added comments as requested

rmadsen-ks (approved):
Looks great-

PR #530, opened 2022-05-02
Create lock directory if it is missing.

### LINKED ISSUE #528: Cannot install a system wide package on Ubuntu
**Describe the bug**
Install fails for "PathWave License Manager" on Ubuntu

**To Reproduce**
1. `tap package install "PathWave License Manager"`

**Actual  behavior**
Installation fails.
I am being asked for sudo password

**Expected behavior**
PWLM should be installed

**Screenshots**
Log: [SessionLog 2022-04-29 10-49-04.txt](https://github.com/opentap/opentap/files/8589120/SessionLog.2022-04-29.10-49-04.txt)

 - OS: Ubuntu 20.04
 - OpenTAP Version 9.17.4

alnlarsen:
flock would throw an exception if the directory containing the lock file was missing.

Closes #528

### DIFF
--- a/Package/PackageActions/LockingPackageAction.cs
+++ b/Package/PackageActions/LockingPackageAction.cs
@@ -94,7 +94,9 @@ namespace OpenTap.Package
                 FileSystemHelper.EnsureDirectory(Target);
             }
 
-            using var fileLock = FileLock.Create(Path.Combine(Target, ".lock"));
+            var lockfile = Path.Combine(Target, ".lock");
+            FileSystemHelper.EnsureDirectory(lockfile);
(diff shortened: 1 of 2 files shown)


rmadsen-ks (changes requested):
I can see that we use this inside  LockingPackageAction when setting --target. Maybe it would be better to ensure that the directory exists there and throw an exception if it is not the case inside FileLock?

I think it might be a bit surprising if FileLock also creates directories.

alnlarsen:
Hmm, I agree, an error would be expected in this case.

rmadsen-ks (approved):
this looks good

rmadsen-ks (approved):
Looks great!

PR #539, opened 2022-05-06
538 don't log errors when typedata providers call Installation.GetPackages

### LINKED ISSUE #538: Error in log when calling Installation.GetPackages from typedata providers
In 9.17, we added a Memorizer to the Installation class. This results in error messages in the logs if an `IStackedTypeDataProvider` implementation tries to check which packages are installed because the memorizer detects the cycle and throws an exception.

This is not super critical because it will be resolved once installed packages are cached, but it is still annoying and potentially confusing why this error occurs.

alnlarsen:
This adds a virtual method to the Memorizer which allow inheritors to define how cycles should be handled.

Closes #538

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,6 +388,7 @@ jobs:
     needs:
       - Build-Win
       - Build-ApiDoc
+      - GetVersion
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


rmadsen-ks (approved):
Looks great!

AsgerIversen (commented):
I'm good with the Memorizer change, but have a question regarding the build script change..

AsgerIversen (approved):

PR #540, opened 2022-05-06
537 type data unbrowsable bug

### LINKED ISSUE #537: TypeData made with FromType are sometimes unbrowsable
This causes issues with plugins that provide their own type discovery.

The reason for this issue is that inside the TypeData class there are two constructors where one sets Browsable = True and the other does not.

rmadsen-ks:
Closes #537 

This PR contains a fix for #537, but also a refactoring which maybe confusing if you view the full PR.

The first refactoring commit has NO functional changes. The following commits is a unit test. the only functional change is the last commit.

The use case is hard to test and mostly shows when the type is from a dynamic assembly which has significance when using TypeData.FromType as FromType falls back to using a different constructor when a dynamic assembly is detected.


alnlarsen:
This looks awesome! Can we add a case to the unittest to also set when the Browsable attribute is missing or false?

alnlarsen (changes requested):
See above

alnlarsen (approved):
Awesome!

PR #542, opened 2022-05-06
541 type data attributes gets unintentionally inherited

### LINKED ISSUE #541: TypeData Attributes gets unintentionally inherited
TypeData.Attributes specifies to use un-inherited attributes, but actually the code in ReflectionHelper does exactly the opposite.

```cs
// the code from typedata
public IEnumerable<object> Attributes => attributes ?? (attributes = Load()?.GetAllCustomAttributes(false)) ?? Array.Empty<object>();

// GetAllCustomAttributes(bool):
public static object[] GetAllCustomAttributes(this MemberInfo prop, bool inherit)
{
   if(inherit) // <-- this should be negated, otherwise no inherit will be used when inherit is true.
        return attrslookupNoInherit.GetValue(prop, getAttributesNoInherit);
   return GetAllCustomAttributes(prop);
 }
```

rmadsen-ks:
Closes #541

Note, this maybe a breaking change since we change previously actual behavior the intended behavior.

### DIFF
--- a/Engine.UnitTests/SerializerTest.cs
+++ b/Engine.UnitTests/SerializerTest.cs
@@ -2580,6 +2580,7 @@ namespace OpenTap.Engine.UnitTests
         }
 
         /// <summary> This class just inhertis from test plan </summary>
+        [AllowAnyChild] 
         public class TestPlanTest2 : TestPlan
         {
             
--- a/Engine.UnitTests/TypeDataTest.cs
+++ b/Engine.UnitTests/TypeDataTest.cs
@@ -1,12 +1,8 @@
 ﻿using NUnit.Framework;
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
-using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace OpenTap.UnitTests
 {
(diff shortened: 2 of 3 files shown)


alnlarsen (approved):
Looks good.

AsgerIversen (changes requested):
Looks like there is an issue with a unit test.

```
Failed TestSerializeInheritTestPlan [4 ms]
  Error Message:
   System.ArgumentException : Cannot add test step of type DelayStep as a root test step (must be nested).
  Stack Trace:
     at OpenTap.TestStepList.InsertItem(Int32 index, ITestStep item) in D:\a\opentap\opentap\Engine\TestStepList.cs:line 271
   at System.Collections.ObjectModel.Collection`1.Add(T item)
   at OpenTap.Engine.UnitTests.XmlTextAttributeTest.TestSerializeInheritTestPlan() in D:\a\opentap\opentap\Engine.UnitTests\SerializerTest.cs:line 2597
  Passed TestSerializeProcessStepWithNullDefaultValue [5 ms]
```

rmadsen-ks:
Fixed the unit test. I think this proves that this change is slightly breaking as people who have previously been relying on inheriting AllowAnyChild may get into trouble now.

AsgerIversen (approved):

PR #545, opened 2022-05-09
Add SessionLog initialization option to not lock session log files

### LINKED ISSUE #518: OpenTap build improvement: log files are locked
Log files are locked (blocking e.g. git clean -dfx) - OpenTAP tasks should release these text files

This is an issue because dotnet reuses processes and tasks between builds, and the msbuild tasks shipped by OpenTAP do not release their held resources after a task finishes -- this is usually done when the process exits, which would usually be after a CLI action ends, but this is not how OpenTAP is used in build tasks. We should add setup and cleanup code to the tasks.

alnlarsen:
This fixes an issue that prevented `git clean -dfx` from working because OpenTAP keeps log files open.

Closes #518

### DIFF
--- a/Engine/SessionLogs.cs
+++ b/Engine/SessionLogs.cs
@@ -44,14 +44,15 @@ namespace OpenTap
 
         static string currentLogFile;
 
+        // This controls whether or not session logs should keep files locked
+        private static bool canDeleteLog = false;
+
         /// <summary>
         /// Initializes the logging. Uses the following file name formatting: SessionLogs\\[Application Name]\\[Application Name] [yyyy-MM-dd HH-mm-ss].txt.
         /// </summary>
         public static void Initialize()
         {
             if (currentLogFile != null) return;
-            
-           
 
             var timestamp = System.Diagnostics.Process.GetCurrentProcess().StartTime.ToString("yyyy-MM-dd HH-mm-ss");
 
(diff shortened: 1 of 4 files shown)


Engine/SessionLogs.cs
```diff
@@ -73,6 +74,17 @@ public static void Initialize()
...
         {
+            // We can't just add this as a parameter with a default value because
+            // it isn't backwards compatible with plugins compiled against older versions.
+            Initialize(tempLogFileName, canDeleteLog);
+        }
+        
+        /// <summary>
+        /// Initializes the logging. 
+        /// </summary>
+        public static void Initialize(string tempLogFileName, bool canDelete)
```
rmadsen-ks, line 85: Can we add to the xml comment what canDelete does and possibly improve the name maybe 'noWriteLock' or noExclusiveWriteLock?

Engine/SessionLogs.cs
```diff
@@ -73,6 +74,17 @@ public static void Initialize()
         /// </summary>
         public static void Initialize(string tempLogFileName)
         {
+            // We can't just add this as a parameter with a default value because
+            // it isn't backwards compatible with plugins compiled against older versions.
+            Initialize(tempLogFileName, canDeleteLog);
```
rmadsen-ks, line 79: I dont understand this comment. Why is it not compatible with older plugins?

rmadsen-ks (changes requested):

Engine/SessionLogs.cs
alnlarsen: Because on the IL level that would mean removing this overload and adding a method with a different signature

alnlarsen:
I agree with this feedback. Please have another look!

rmadsen-ks (changes requested):
It seems the build logs does not get flushed when the build ends.

image

alnlarsen:
> It seems the build logs does not get flushed when the build ends.
> 
> image

That's a good point! I have added a flush after the build task ends.

rmadsen-ks (approved):
Seems to work perfectly now!

rmadsen-ks:
Note, now the log files are separated for each msbuild process:
``` 
ls bin/Debug/SessionLogs/
MSBuild.23828 2022-05-12 13-36-31 (1).txt  MSBuild.23828 2022-05-12 13-36-31.txt  MSBuild.26736 2022-05-12 13-36-31.txt
```

AsgerIversen (approved):
Thanks @alnlarsen. Looks good now

PR #549, opened 2022-05-10
Fix warnings when installing system-wide packages

### LINKED ISSUE #509: Warnings when installing a system wide package
Steps:

`tap package install "Keysight Fixed License Manager"`  

- Log: [SessionLog 2022-04-10 23-56-30.txt](https://github.com/opentap/opentap/files/8461720/SessionLog.2022-04-10.23-56-30.txt)
- Warning     ; Could not find OpenTAP in C:\ProgramData\Keysight\Test Automation.

`tap package install "PathWave License Manager"`  

- Log: [SessionLog 2022-04-10 23-58-11.txt](https://github.com/opentap/opentap/files/8461732/SessionLog.2022-04-10.23-58-11.txt)
- Warning     ; Selected package PathWave License Manager for Windows, x64 is incompatible with the host platform Windows, AnyCPU.
- Warning     ; Package 'PathWave License Manager' depends on 'OpenTAP' which will not be installed.

All these `subprocess` messages, are they necessary to be displayed? Maybe these should be debug messages

image

alnlarsen:
This removes a few warnings that were simply incorrect

Other warnings have been demoted to 'info' when the user has explicitly set the '--no-dependencies' option

Bogus 'Duplicate packages' warning has been removed when an Installation has the systemwide dir as a target

This also includes a minor bugfix to the repositories used in elevated processes, and a butfix to the default Architecture in Installations

Closes #509

### DIFF
--- a/Shared/SubProcessHost.cs
+++ b/Shared/SubProcessHost.cs
@@ -21,8 +21,7 @@ namespace OpenTap
     {
         public bool ForwardLogs { get; set; } 
         public string LogHeader { get; set; } = "";
-        public string[] MutedSources { get; set; } = Array.Empty<string>();
-
+        public HashSet<string> MutedSources { get; } = new HashSet<string>();
 
         public static bool IsAdmin()
         {
--- a/Shared/SubProcessHost.cs
+++ b/Shared/SubProcessHost.cs
@@ -20,7 +20,8 @@ namespace OpenTap
     class SubProcessHost
     {
         public bool ForwardLogs { get; set; } 
-        public string LogHeader { get; set; } = "Subprocess";
+        public string LogHeader { get; set; } = "";
+        public string[] MutedSources { get; set; } = Array.Empty<string>();
 
 
         public static bool IsAdmin()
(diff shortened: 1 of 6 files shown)


Shared/SubProcessHost.cs
```diff
@@ -20,7 +20,8 @@ namespace OpenTap
     class SubProcessHost
     {
         public bool ForwardLogs { get; set; } 
-        public string LogHeader { get; set; } = "Subprocess";
+        public string LogHeader { get; set; } = "";
+        public string[] MutedSources { get; set; } = Array.Empty<string>();
```
rmadsen-ks, line 24: maybe this should be a `HashSet<string>`

rmadsen-ks (changes requested):
Otherwise things seems to work very well.

Shared/SubProcessHost.cs
alnlarsen: Hmm. I agree.

rmadsen-ks (approved):
Works perfectly/

PR #553, opened 2022-05-10
548 bump smartinstaller

### LINKED ISSUE #534: OpenTAP for Windows requires .NET Framework but does not install it
**Steps**

1. Install OpenTAP on a Windows 10 machine without .NET Framework 4.8
2. Run OpenTAP

**Actual Results**

Error, OpenTAP does not work

image

**Expected Result**

OpenTAP installer should also install the required .NET Framework. Currently it only installs .NET 6

Found by @witoldpiet

### LINKED ISSUE #548: Consistent installer failure (because of OS language?)
I tried installing OpenTAP 9.17.4 and 9.17.3 on multiple different windows machines using the the released *.exe installers. Installation of .NET 6 works as expected on all machines that didn't yet have it installed. All computers run fully patched Windows 10 in German. All attempts of installation fail with the same error message regarding permissions. The best translation of the error message I could find goes like this: "Some or all Identity references could not be translated". Since I am running the installer as Administrator with full privileges and have even tried different source- and target locations on each system, I am now assuming there is an issue with translated user names (e.g. "Everyone" equals "Jeder" in German).

Screenshot 2022-05-09 175114

alnlarsen:
From what I could tell, the current smartinstaller image we use was built February 9, and we merged some localization fixes in the smartinstaller on February 23. Let's try bumping the smart installer to the latest beta and see if it fixes the problem.

Closes #548
Closes #534

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -710,7 +710,7 @@ jobs:
       - GetVersion
       - Package-Win
     container:
-      image: registry.gitlab.com/opentap/buildrunners/smartinstaller:1.0.0-beta.3
+      image: registry.gitlab.com/opentap/buildrunners/smartinstaller:1.1.0-beta.7
       credentials:
         username: github
         password: ${{ secrets.GITLAB_REGISTRY_TOKEN}}


rmadsen-ks (approved):
Looks great

PR #554, opened 2022-05-10
Build Instabilities Fix

### LINKED ISSUE #527: Improve UnitTest RunParseTest
The current RunParseTest unit test starts up several sub processes in order to verify the behavior of external parameters.

It is slow and flaky and when a problem occurs it can be hard to understand what went wrong.

This test should be improved by testing in a simpler way and if it tests multiple things, maybe split it into more tests

### LINKED ISSUE #529: Retrying repo downloads 60 times in a loop is excessive
See the offending code here: https://github.com/opentap/opentap/blob/29081f8230af5ab025e9846871659c93622af3e3/Package/Repositories/HttpPackageRepository.cs#L132

This retry code is almost never executed. I experienced it in a local environment with an alpha build of the HTTP repository (I suspect) because the repository claimed to have a package, but an exception was triggered in the repo when I tried to download said package.

In any case, retrying 60 times here seems excessive, and while this behavior likely doesn't affect anyone, the code could use some cleanup.

### LINKED ISSUE #551: tap package install of file-like objects
Sometimes it is pretty obvious that the user is trying to install a tap package from a file:
```tap package install Test.TapPackage```

Still in this case we ask the package repository for a package named Test.TapPackage, which does not exist.

When a package name ends with TapPackage we should just assume that it is always a file and throw a FileNotFoundException instead of asking the package repositories.

### LINKED ISSUE #552: HttpPackageRepositories 'DoDownloadPacakage' does not support authentication
Most of the HttpPackageRepository API support authentication but not the part that downloads the package.

### LINKED ISSUE #555: tap package install cannot be stopped when running in a test plan
Run the following Run Program step:

image

Observer that the test plan never completes. If Abort is hit, the test plan will also not stop.

That is because the actually running process is a child-child process which is not reached with Process.Kill()

### LINKED ISSUE #558: Optimized ResultTables removes column parameters
An intermittent error occurs in ResultTableAndColumnParameters, it gets worse with more timeconsuming result listeners. 

Result tables gets merged even though their columns has different parameters. 

This is very hard to reproduce under normal circumstances, especially since column parameters are rarely used.

the following new unit test shows the issue:

```

        [Test]
        public void ResultOptimizerTest()
        {
            ResultColumn createColumn<T>(string name)
            {
                return new ResultColumn("x", new T[2]);
            }

            {
                var rc1x = createColumn<int>("X");
                var rc1y = createColumn<int>("Y");
                var rt1 = new ResultTable("XY", new[] {rc1x, rc1y});

                var rc2x =  createColumn<int>("X");
                var rc2y = createColumn<int>("X");
                var rt2 = new ResultTable("XY", new[] {rc2x, rc2y});
                Assert.IsTrue(ResultTableOptimizer.CanMerge(rt1, rt2));
                var merged = ResultTableOptimizer.MergeTables(new []{rt1, rt2});
                Assert.AreEqual(4, merged.Columns[0].Data.Length);
            }
            {
                var rc1x = createColumn<int>("X");
                var rc1y = createColumn<int>("Y");
                var rt1 = new ResultTable("XY", new[] {rc1x, rc1y});

                var rc2x =  createColumn<int>("X");
                var rc2y = createColumn<int>("X");
                rc2y = rc2y.AddParameters(new ResultParameter("Unit", "X"));
                var rt2 = new ResultTable("XY", new[] {rc2x, rc2y});
                Assert.IsTrue(ResultTableOptimizer.CanMerge(rt1, rt2) == false);
            }
        }
    }
```

### LINKED ISSUE #559: AssemblyVersion test fails intermittently
The test plan test NoAssemblyInfo.TapPlan fails intermittently with an error seemingly caused by the version not being properly injected into the builds in some situations.

### LINKED ISSUE #560: HttpPackageRepository: Improve robustness of http requests
We should implement a backoff / retry policy when doing Http requests to the package repository. This is important for builds which are otherwise sensitive to transient connection issues.

### LINKED ISSUE #561: Reduce the amount of logs when unable to connect to a package repository
This is what happens when downloading from an invalid package repository url:
```
 ./tap.exe package install ASD -r packageasd.asd.opentap.io
Unable to connect to 'http://packageasd.asd.opentap.io'.
Unable to connect to 'http://packageasd.asd.opentap.io'.
Unable to connect to 'http://packageasd.asd.opentap.io'.
Could not download one or more packages.
```

rmadsen-ks:
Various problems caused issues in the build pipelines. Lets ensure that builds always runs perfectly when there is no bugs.

Fixes #552
Fixes  #551
Fixes  #555 
Fixes  #558

Fixes #559 by using the debug version of OpenTap instead of the one installed by the project.
Fixes #527
Fixes #529
Fixes #560
Fixes #561

Note, the problematic "RunParseTest" was moved from a unit test and into the testing test plan.
Also note, a few of our jobs are now run in matrix mode, so that we have a better chance of catching issues that affect build stability.


alnlarsen (dismissed):
This looks awesome! Let's get this merged ASAP so we don't have to retry builds.

AsgerIversen (approved):

PR #557, opened 2022-05-10
Fix double pipelines

### LINKED ISSUE #556: ci.yml: on pull_request increases build costs by a factor of 2.
It seems that pushes now causes all build jobs to be run twice, causing an increased build cost.

alnlarsen:
Stop running double pipelines on pushes to PRs

Closes #556


rmadsen-ks (approved):
I strongly agree with this.

PR #564, opened 2022-05-11
563 improve missing dependency install error message

### LINKED ISSUE #563: Package install: Confusing error message when a package dependency is not in specified repositories
If you install e.g. `Editor` and specify a local file directory as the repository, and you don't have all of its dependencies in the same repository, the error message you get is a null reference exception with no mention of the dependency conflict.

The actual error message should say that the dependency could not be resolved from the specified repositories.

alnlarsen:
Move the early return condition earlier in the method to avoid a null reference exception.

Closes #563


Denrask (approved):
No longer getting the null reference exception! Thanks

PR #565, opened 2022-05-11
Re-included System.Security.Principal.Windows

rmadsen-ks:
This was needed to do elevation.


alnlarsen (approved):
This PR has everything that I am looking for in a fix. Thank you, Rolf.

PR #566, opened 2022-05-11
Startup Performance Improvements

rmadsen-ks:
- SimpleTapAssemblyResolver is no longer needed as we dont deploy self-contained installations.
- Added minor plugin searcher performance improvements

### DIFF
--- a/Cli/TapInitializer.cs
+++ b/Cli/TapInitializer.cs
@@ -28,8 +28,22 @@ namespace OpenTap
             public static readonly InitTraceListener Instance = new InitTraceListener();  
         }
 
+        static string OpenTapLocation
+        {
+            get
+            {
+                var loc = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
+                return Path.Combine(loc, "OpenTap.dll");
+            }
+        }
+
         internal static void Initialize()
         {
+            // This current assembly looks for the opentap DLL in the wrong location.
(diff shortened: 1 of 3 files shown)


alnlarsen (changes requested):
This is a nice cleanup, but it looks like it breaks on MacOS / Linux

alnlarsen (approved):
This doesn't seem to have a measurable impact on performance, but it's a nice cleanup. Having a throwaway assembly resolver is an insane way to load OpenTap.dll.

Cli/TapInitializer.cs
```diff
@@ -26,63 +25,37 @@ public void EventsLogged(IEnumerable<Event> events)
...
+            public static readonly InitTraceListener Instance = new InitTraceListener();  
+        }
+
+        static IEnumerable<string> OpenTapLocations
+        {
+            get
+            {
+                var loc = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
+                yield return Path.Combine(loc, "OpenTap.dll");
+            }
```
AsgerIversen, line 37: Isn't there always only one item in this enumerable?

AsgerIversen (changes requested):
One small comment. Otherwise looks good.

Cli/TapInitializer.cs
rmadsen-ks: I've simplified it.

AsgerIversen (approved):

PR #569, opened 2022-05-12
520 ITypeDataSource Implementation

### LINKED ISSUE #520: ITypeDataSource
The class TypeData has an Assembly property, but types coming from other places does not have a specifier showing where they come from.

rmadsen-ks:
Added ITypeDataSource and ITypeDataSourceProvider

The first implementor of ITypeDataSource is AssemblyData and DotNetTypeDataSearcher, which are classes already existing inside OpenTAP. The added capability is to be able to query the ITypeDataSource of a Type using TypeData.GetTypeDataSource.

Additional changes are related to ITypeDataSource.Version which is a string, that can be a semantic version but is not necessesarily.

Added PluginOrderAttribute to be able to determine the orders of ITypeDataSourceProviders.

Closes #520


Engine/Reflection/PluginOrderAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+        
+        /// <summary>
+        /// This comparer can be used for sorting a list of objects based on their plugin order attributes.
+        /// </summary>
+        internal static Comparison<object> Comparer
+        {
+            get
+            {
+                var lut = new Dictionary<object, PluginOrderAttribute>();
```
alnlarsen, line 39: lut == lookup table? can we get a better name?

Engine/TestStep.cs
```diff
@@ -192,7 +190,7 @@ public string Version
                 }
                 else
                 {
-                    Log.Warning("Could not parse test plan file specified version {0} of step '{1}' as a semantic version, but version {2} is installed, compatibility issues may occur.", value, Name, installedVersion);
+                    //Log.Warning("Could not parse test plan file specified version {0} of step '{1}' as a semantic version, but version {2} is installed, compatibility issues may occur.", value, Name, installedVersion);
```
alnlarsen, line 193: why was this commented?

alnlarsen (changes requested):

alnlarsen (approved):
I think this looks fine now, though I didn't actually test it.

AsgerIversen (changes requested):
APIs looks good. I have one small comment.

@rmadsen-ks can you test a bit by using this in Python plugin/KS8400 and then merge if that works out?

Package/CreatePackage/Package.cs
```diff
@@ -450,6 +450,10 @@ internal static void findDependencies(this PackageDef pkg, List<string> excludeA
                             var requiredAsm = dependentAssemblyNames.FirstOrDefault(dep => dep.Name == candidateAsm.Name);
                             if (requiredAsm != null)
                             {
+                                if (candidateAsm.Version == null)
+                                    throw new Exception($"Assembly {candidateAsm} version is null");
```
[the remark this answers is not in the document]
rmadsen-ks: I think it is technically impossible for the version to be null under normal circumstances. The only reason I added this here is that OpenTap.Utils.Compatible throws a NullReferenceExeption exception if Version is null and I wanted to be easier to spot why this issue occurs.

AsgerIversen (approved):

PR #571, opened 2022-05-12
332 Expand/Collapse TestSteps

### LINKED ISSUE #332: API for expanding/collapsing test steps heirachies
Based on user request. It would be helpful to be able to control the expand/collapse state from an API, and also save the state as a part of the test plan.

### LINKED ISSUE #570: XmlAttributeAttribute does not work for dynamic IMemberData
Dynamically added IMemberData implementations which has the XmlAttributeAttribute are able to serialize, but not deserialize.

Needed for #332

rmadsen-ks:
- Implemented the expanded/collasped state as a dynamic member property.
- Fixed bug in XML deserialization that caused XmlAttributeAttribute not work for dynamically created properties.
- Added a unit test that tests the UI folding state.

Closes #332 
Closes #570


Engine/UiFolding.cs
```diff
@@ -0,0 +1,52 @@
...
+        public static void SetFolding(object obj, bool? expanded)
+        {
+            UiFoldingProperty.SetValue(obj, expanded switch
+            {
+                null => null,
+                true => "expand",
+                false => "collapse"
+            });
+        }
+        /// <summary>  Sets the current collapsed or folding state. If not set it will return null. </summary>
```
alnlarsen, line 39:
```suggestion
        /// <summary>  Gets the current folding state. If not set it will return null. </summary>
```

alnlarsen (changes requested):
The implementation looks good and seems to work as intended, but it looks like one of the doc comments are incorrect.

rmadsen-ks:
@AsgerIversen, should we create a new namespace for this?

AsgerIversen:
> @AsgerIversen, should we create a new namespace for this?

If we were doing this from scratch, we probably should have a UI namespace. If we create one now, we will not be able to move things like UserInput, IUserInterface, IMenuModel, IconNames, ColumnDisplayName, LayoutAttribute and maybe the entire Annotations system. With these UI related types left outside the UI namespace, I think introducing such a namespace does not really give us a more intuitive structure.

This warning also has some guidance on the use of namespaces in .NET: https://docs.microsoft.com/en-gb/previous-versions/visualstudio/visual-studio-2015/code-quality/ca1020-avoid-namespaces-with-few-types?view=vs-2015

Engine/ChildItemVisibility.cs
```diff
@@ -0,0 +1,53 @@
...
+            VisibilityProperty.SetValue(obj, expanded switch
+            {
+                null => null,
+                true => "expand",
+                false => "collapse"
+            });
+        }
+        /// <summary>  Gets the current visibility state. If not set it will return null. </summary>
+        /// <param name="obj">The object from which to get visibility.</param>
+        public static bool? GetVisibility(object obj)
```
AsgerIversen, line 42: Should we make an enum for this instead of a bool?. Similar to WPF vilibility. That should also make it show up nicer in xml without manually doing string stuff.

AsgerIversen (changes requested):
Looks good. One small thing...

Engine.UnitTests/ChildItemVisibilityTest.cs
```diff
@@ -0,0 +1,31 @@
+using NUnit.Framework;
+using OpenTap.Plugins.BasicSteps;
+
+namespace OpenTap.UnitTests
+{
+    [TestFixture]
+    public class ChildItemVisibilityTest
+    {
+        [Test]
+        public void TestSerializeUiFolding()
```
alnlarsen, line 10: TestSerializeChildItemVisibility?

Engine/ChildItemVisibility.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.ComponentModel;
+using System.Xml.Serialization;
+
+namespace OpenTap
+{
+    /// <summary>  Child item visibility implements the concept of showing or hiding a child item. This could for example be the child steps of a test step. </summary>
+    public static class ChildItemVisibility
+    {
+        // The UI Folding property can have 3 states, null, collapsed and expanded.
```
alnlarsen, line 10: This comment is no longer relevant

Engine/ChildItemVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            /// <summary> Child items are visible. This is the default state. </summary>
+            Visible, 
+            /// <summary> Child items are collapsed.</summary>
+            Collapsed
+        }
+        
+        internal static readonly DynamicMember VisibilityProperty = new DynamicMember
+        {
+            Name = "UI.Visibility",
```
alnlarsen, line 25: ChildItem.Visibility?

Engine/ChildItemVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>  Set visibility to collapsed or Visible. </summary>
+        /// <param name="obj">The object for which to set visibility.</param>
+        /// <param name="visibility"> </param>
+        public static void SetVisibility(object obj, Visibility visibility)
+        {
+            VisibilityProperty.SetValue(obj, visibility);
+        }
+        
+        /// <summary>  Sets the current collapsed or folding state. If not set it will return null. </summary>
```
alnlarsen, line 42: This cannot return null, right?

Engine/ChildItemVisibility.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="obj">The object from which to get folding.</param>
```
alnlarsen, line 43: folding -> visibility

alnlarsen (changes requested):
It seems there are a lot of renaming issues

alnlarsen (approved):
Looks good!

alnlarsen (approved):

AsgerIversen (approved):

PR #573, opened 2022-05-12
ProcessStep environment variables

### LINKED ISSUE #476: Process Step: Allow setting environment variables
This would be useful for testing application that depends on environment variables.

Tasks:
- in ProcessStep, add a List<EnvironmentVariable> property.
- Load those variables when running the sub process.
- Add a validation rule to ensure that the same environment variable is not added twice.

frederikja163:
Adds environment variables as a parameter to the ProcessStep, as a new feature requested by #476 
Duplicate environment variables will fail, however multiple different environment variables can be added.

Also added a new test CLI action to the engine unit tests to print environment variables this way it is possible to use a regex to prove the environment variables have been set.
Added 6 testcases to test the process step environmet variables
* One test for a single environment variable
* Three tests for multiple environment variables
* Two tests for duplicate environment variables
Closes #476

### DIFF
--- a/BasicSteps/ProcessStep.cs
+++ b/BasicSteps/ProcessStep.cs
@@ -13,12 +13,6 @@ using System.Threading;
 
 namespace OpenTap.Plugins.BasicSteps
 {
-    public class EnvironmentVariable
-    {
-        public string Name { get; set; }
-        public string Value { get; set; }
-    }
-
     [Display("Run Program", Group: "Basic Steps", Description: "Runs a program, and optionally applies regular expressions (regex) to the output.")]
     public class ProcessStep : RegexOutputStep
     {
@@ -39,7 +33,7 @@ namespace OpenTap.Plugins.BasicSteps
         public string WorkingDirectory { get; set; } = "";
 
         [Display("Environment Variables", Order: -2.25, Description: "The enironment variables passed to the program.")]
-        public VirtualCollection<EnvironmentVariable> EnvironmentVariables { get; set; } = new VirtualCollection<EnvironmentVariable>();
+        public VirtualCollection<KeyValuePair<string, string>> EnvironmentVariables { get; set; } = new VirtualCollection<KeyValuePair<string, string>>();
 
         [Display("Wait For Process to End", Order: -2.2,
             Description: "Wait for the process to terminate before continuing.")]
@@ -94,11 +88,10 @@ namespace OpenTap.Plugins.BasicSteps
             HashSet<string> seenVariables = new HashSet<string>();
             foreach (var variable in EnvironmentVariables)
             {
-                if (seenVariables.Contains(variable.Name))
+                if (!seenVariables.Add(variable.Key))
                 {
                     return false;
                 }
-                seenVariables.Add(variable.Name);
             }
             return true;
         }
@@ -132,11 +125,6 @@ namespace OpenTap.Plugins.BasicSteps
             Int32 timeout = Timeout <= 0 ? Int32.MaxValue : Timeout;
             prepend = string.IsNullOrEmpty(LogHeader) ? "" : LogHeader + " ";
 
-            foreach (var environmentVariable in EnvironmentVariables)
-            {
-                Environment.SetEnvironmentVariable(environmentVariable.Name, environmentVariable.Value);
-            }
-
             var process = new Process
             {
                 StartInfo =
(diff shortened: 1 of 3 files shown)


BasicSteps/ProcessStep.cs
```diff
@@ -73,9 +84,28 @@ public Int32 Timeout
...
+        private bool HasNoDuplicateEnvironmentVariables()
+        {
+            HashSet<string> seenVariables = new HashSet<string>();
+            foreach (var variable in EnvironmentVariables)
+            {
+                if (seenVariables.Contains(variable.Name))
+                {
+                    return false;
+                }
+                seenVariables.Add(variable.Name);
```
rmadsen-ks, line 101: HashSet.Add returns false if the value already exists in the collection, maybe that can be used here?

BasicSteps/ProcessStep.cs
```diff
@@ -102,6 +132,11 @@ public override void Run()
             Int32 timeout = Timeout <= 0 ? Int32.MaxValue : Timeout;
             prepend = string.IsNullOrEmpty(LogHeader) ? "" : LogHeader + " ";
 
+            foreach (var environmentVariable in EnvironmentVariables)
+            {
+                Environment.SetEnvironmentVariable(environmentVariable.Name, environmentVariable.Value);
```
rmadsen-ks, line 137: Let's not set the environment variable for the parent process. Use ProcessStartInfo.EnvironmentVariables to set environment variables in the new process directly.

BasicSteps/ProcessStep.cs
```diff
@@ -11,6 +13,12 @@
 
 namespace OpenTap.Plugins.BasicSteps
 {
+    public class EnvironmentVariable
```
rmadsen-ks, line 16: Should we make this a nested class inside ProcessStep? Or maybe we can just use `KeyValuePair<string,string>`?

rmadsen-ks (changes requested):
Really nice testing! Had a few comments though.

BasicSteps/ProcessStep.cs
frederikja163: Now it should set the environment variables only on the new process.

BasicSteps/ProcessStep.cs
frederikja163: Yup, that is totally possible, changed in the latest commit.

BasicSteps/ProcessStep.cs
```diff
@@ -73,9 +78,27 @@ public Int32 Timeout
         ManualResetEvent outputWaitHandle, errorWaitHandle;
         StringBuilder output;
 
+        public ProcessStep()
+        {
+            Rules.Add(HasNoDuplicateEnvironmentVariables, "Cannot have multiple environment variables with the same name.", nameof(EnvironmentVariables));
```
rmadsen-ks, line 83: What do you think about "Environment variable names must be unique."?

BasicSteps/ProcessStep.cs
```diff
@@ -30,6 +32,9 @@ public class ProcessStep : RegexOutputStep
         [DirectoryPath]
         public string WorkingDirectory { get; set; } = "";
 
+        [Display("Environment Variables", Order: -2.25, Description: "The enironment variables passed to the program.")]
```
rmadsen-ks, line 35:
```suggestion
        [Display("Environment Variables", Order: -2.25, Description: "The environment variables passed to the program.")]
```

BasicSteps/ProcessStep.cs
rmadsen-ks, line 83: Are there some names which are invalid for environment variables? Maybe we should check for those.
Maybe we need some rules like these?
- ```string.IsNullOrWhiteSpace(key)```
- ```key.Trim() == key```

rmadsen-ks (changes requested):
Generally seems great. I had a few comments though.

BasicSteps/ProcessStep.cs
frederikja163: Fixed the typo

BasicSteps/ProcessStep.cs
frederikja163: I will add the environment variables rules from this page:
Along with what you suggested.
https://docs.microsoft.com/en-us/windows/win32/procthread/environment-variables

Other than that i think we should be really careful about validating these too much, as different platforms might very well have different possibilities for what environment variables they allow or dont allow.

frederikja163: Yeah, that makes more sense. Fixed it

rmadsen-ks (changes requested):
Mostly looks good.

rmadsen-ks (approved):
Lets merge it.

PR #574, opened 2022-05-12
Waiting for ressource race condition

### LINKED ISSUE #121: Incomplete message in log when opening resources
Originally filed November 23 2021 by Wooi Po Chang on [GitLab](https://gitlab.com/OpenTAP/opentap/-/issues/811)

The log message `Waiting for resources to open:` seems to be incomplete
image

May consider improving `PrintWaitingMessage` function in [TestPlanExecution.cs](https://gitlab.com/OpenTAP/opentap/-/blob/master/Engine/TestPlanExecution.cs)

frederikja163:
Fixed race condition when ressources are waiting to open, by saving the disconnected ressources to a list before starting to print the message.

This should fix #121
Testing on this one is hard as it is hard to recreate the race condition reliably. Thus no testing has been done yet.

### DIFF
--- a/Engine/TestPlanExecution.cs
+++ b/Engine/TestPlanExecution.cs
@@ -95,10 +95,16 @@ namespace OpenTap
 
         internal static void PrintWaitingMessage(IEnumerable<IResource> resources)
         {
-            Log.Info("Waiting for resources to open:");
-            foreach (var resource in resources)
+            // Save disconnected ressources to avoid race conditions.
+            var waitingRessources = resources.Where(r => !r.IsConnected).ToList();
+            if (waitingRessources.Count == 0)
             {
-                if (resource.IsConnected) continue;
+                return;
+            }
+
+            Log.Info("Waiting for resources to open:");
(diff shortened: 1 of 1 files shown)


Engine/TestPlanExecution.cs
```diff
@@ -95,10 +95,16 @@ enum failState
 
         internal static void PrintWaitingMessage(IEnumerable<IResource> resources)
         {
+            // Save disconnected ressources to avoid race conditions.
+            var waitingRessources = resources.Where(r => !r.IsConnected).ToList();
```
rmadsen-ks, line 99: Let's use ToArray here - if the list is empty then an empty array gets us Array.Empty<Resource>(), which does not cause any allocation.

rmadsen-ks (changes requested):

Engine/TestPlanExecution.cs
frederikja163: Now uses .ToArray()

rmadsen-ks (approved):
Looks great!

PR #576, opened 2022-05-12
230 improve handling of unloadable plugins

### LINKED ISSUE #230: The `PluginManager` loads assemblies that cannot be loaded
When loading all plugins and trying to get custom attributes it throws a FileNotFoundException

```cs
var assemblies = PluginManager.GetAllPlugins()
    .Select(x => x.Type.Assembly)
    .Distinct()
     .Where(x => x.GetCustomAttributes<MyExtension>().Any());
```

alnlarsen:
Make GetAllPlugins not return plugins which failed to load.

In general, it looks like we currently handle unloadable plugins okay. `GetAllPlugins` should be easier to use now that it doesn't return incompatible plugins.

`TypeData.GetDerivedTypes` does not need to be modified because all the `DescendsTo` logic already returns false if a candidate for a derived type fails to load.

Closes #230

### DIFF
--- a/Engine/PluginManager.cs
+++ b/Engine/PluginManager.cs
@@ -189,7 +189,10 @@ namespace OpenTap
         {
             PluginSearcher searcher = GetSearcher();
             return searcher.PluginTypes
-                .Where(st => !st.TypeAttributes.HasFlag(TypeAttributes.Interface) && !st.TypeAttributes.HasFlag(TypeAttributes.Abstract))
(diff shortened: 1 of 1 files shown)


rmadsen-ks (changes requested):
It seems the PR description does not match the changes. Did you include everything we need?

rmadsen-ks (approved):
Looks good.

Issue #575, opened 2022-05-12
Help message `valid subcommands of` doesnt say what the subcommand is


frederikja163:
**Environment:**

Windows 10
Opentap 9.18

**Steps to reproduce:**
1. Open a command line interface inside OpenTAP's installation folder.
2. Type in an invalid command (e.g. tap package notvalid).
3. Observe the output.

**Actual result**
`Valid subcommands of` is written to the console without specifying which command it is specifying subcommands of.

**Expected result**
`Valid subcommands of [command]` should instead be written to the terminal.

Issue #78, opened 2021-12-09
A Validation Message for Invalid Commands Should Be Printed Before Showing Help


AsgerIversen:
Originally filed June 30 2021 by Vlad Opris on [GitLab](https://gitlab.com/OpenTAP/opentap/-/issues/701)

**Environment:**

Windows 10, 64-bit

OpenTAP 9.15.0

**Steps to reproduce:**

1. Open a command line interface inside OpenTAP's installation folder.
2. Type in an invalid command (e.g. tap notvalid).
3. Observe the output.

**Actual result:**

Help is displayed, but the user has no way of knowing what the command they typed in does.

**Expected result:**

Print a validation message (notvalid is not a recognized command), then print help.

PR #577, opened 2022-05-12
575 and 78 Cleanup help message

frederikja163:
Cleanup the help message a bit to make it more uniform, update it to use log.info() rather than Console.WriteLine(),
this also means that messages now arrive in the console in the correct order.
Also prints an error message for incorrect command usage.

**Closes** #575 and #78

**Example of message before:**
```
PS C:\Users\fredandr\opentap\bin\Debug> .\tap.exe package notvalid
OpenTAP Command Line Interface (9.18.0-alpha.18+cf121119.i215-UninstallingMultiplePacka)
Usage: tap <command> [<subcommand(s)>] [<args>]

Valid subcommands of
Run "tap.exe <command> [<subcommand>] -h" to get additional help for a specific command.

package
   create                 Create a package based on an XML description file.
   download               Download one or more packages.
   install                Install one or more packages.
   list                   List locally installed packages and browse the online package repository.
   show                   Show information about a package.
   test                   Run tests on one or more packages.
   uninstall              Uninstall one or more packages.
   verify                 Verify the integrity of one or all installed packages by checking their fingerprints.
PS C:\Users\fredandr\opentap\bin\Debug>
```

**Example of message after:**
```
PS C:\Users\fredandr\opentap\bin\Debug> .\tap.exe package notvalid
OpenTAP Command Line Interface (9.18.0-alpha.18+cf121119.i215-UninstallingMultiplePacka)
"tap.exe package notvalid" is not a recognized command.
Usage: "tap.exe <command> [<subcommand(s)>] [<args>]"

Valid subcommands of
package
   create                 Create a package based on an XML description file.
   download               Download one or more packages.
   install                Install one or more packages.
   list                   List locally installed packages and browse the online package repository.
   show                   Show information about a package.
   test                   Run tests on one or more packages.
   uninstall              Uninstall one or more packages.
   verify                 Verify the integrity of one or all installed packages by checking their fingerprints.

Run "tap.exe <command> [<subcommand(s)>] -h" to get additional help for a specific command.

PS C:\Users\fredandr\opentap\bin\Debug>
```

### DIFF
--- a/Engine/Cli/CliActionExecutor.cs
+++ b/Engine/Cli/CliActionExecutor.cs
@@ -272,13 +272,9 @@ namespace OpenTap.Cli
                 string tapCommand = OperatingSystem.Current == OperatingSystem.Windows ? "tap.exe" : "tap";
 
                 log.Info("OpenTAP Command Line Interface ({0})", getVersion());
-                if (args.Length == 0)
+                if (args.Length != 0)
                 {
-                    log.Info($"\"{tapCommand}\" is not a recognized command.");
-                }
-                else
-                {
-                    log.Info($"\"{tapCommand} {string.Join(" ", args)}\" is not a recognized command.");
+                    log.Error($"\"{tapCommand} {string.Join(" ", args)}\" is not a recognized command.");
                 }
                 log.Info($"Usage: \"{tapCommand} <command> [<subcommand(s)>] [<args>]\"\n");
 
--- a/Engine/Cli/CliActionExecutor.cs
+++ b/Engine/Cli/CliActionExecutor.cs
@@ -269,13 +269,22 @@ namespace OpenTap.Cli
 
                     return TypeData.FromType(typeof(CliActionExecutor)).Assembly.SemanticVersion.ToString(); // OpenTAP is not installed. lets just return this. 
                 }
+                string tapCommand = OperatingSystem.Current == OperatingSystem.Windows ? "tap.exe" : "tap";
 
-                Console.WriteLine("OpenTAP Command Line Interface ({0})", getVersion());
-                Console.WriteLine("Usage: tap <command> [<subcommand(s)>] [<args>]\n");
+                log.Info("OpenTAP Command Line Interface ({0})", getVersion());
+                if (args.Length == 0)
+                {
+                    log.Info($"\"{tapCommand}\" is not a recognized command.");
+                }
+                else
+                {
+                    log.Info($"\"{tapCommand} {string.Join(" ", args)}\" is not a recognized command.");
+                }
+                log.Info($"Usage: \"{tapCommand} <command> [<subcommand(s)>] [<args>]\"\n");
 
                 if (selectedcmd == null)
                 {
-                    Console.WriteLine("Valid commands are:");
+                    log.Info("Valid commands are:");
                     foreach (var cmd in actionTree.SubCommands)
                     {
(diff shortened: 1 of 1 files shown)


Engine/Cli/CliActionExecutor.cs
```diff
@@ -269,26 +269,34 @@ string getVersion()
...
                     return TypeData.FromType(typeof(CliActionExecutor)).Assembly.SemanticVersion.ToString(); // OpenTAP is not installed. lets just return this. 
                 }
+                string tapCommand = OperatingSystem.Current == OperatingSystem.Windows ? "tap.exe" : "tap";
 
-                Console.WriteLine("OpenTAP Command Line Interface ({0})", getVersion());
-                Console.WriteLine("Usage: tap <command> [<subcommand(s)>] [<args>]\n");
+                log.Info("OpenTAP Command Line Interface ({0})", getVersion());
+                if (args.Length == 0)
+                {
+                    log.Info($"\"{tapCommand}\" is not a recognized command.");
```
rmadsen-ks, line 277: i think we can skip printing this line - if the user just writes "tap.exe" we can assume they just want the help. What do you think?

Engine/Cli/CliActionExecutor.cs
```diff
@@ -269,26 +269,34 @@ string getVersion()
...
+                }
+                else
+                {
+                    log.Info($"\"{tapCommand} {string.Join(" ", args)}\" is not a recognized command.");
```
rmadsen-ks, line 281: Maybe this should be log.Error to highlight the problem?

rmadsen-ks (changes requested):
Looks great, just a couple of suggestions.

Engine/Cli/CliActionExecutor.cs
frederikja163: Yeah, this probably makes sense. Change has been made.

Engine/Cli/CliActionExecutor.cs
frederikja163: This is now a log.Error instead.

Engine/Cli/CliActionExecutor.cs
```diff
@@ -269,26 +269,30 @@ string getVersion()
...
                     return TypeData.FromType(typeof(CliActionExecutor)).Assembly.SemanticVersion.ToString(); // OpenTAP is not installed. lets just return this. 
                 }
+                string tapCommand = OperatingSystem.Current == OperatingSystem.Windows ? "tap.exe" : "tap";
 
-                Console.WriteLine("OpenTAP Command Line Interface ({0})", getVersion());
-                Console.WriteLine("Usage: tap <command> [<subcommand(s)>] [<args>]\n");
+                log.Info("OpenTAP Command Line Interface ({0})", getVersion());
+                if (args.Length != 0)
+                {
+                    log.Error($"\"{tapCommand} {string.Join(" ", args)}\" is not a recognized command.");
```
rmadsen-ks, line 277: I think we should move the error message up before line 274 "OpenTAP Command Line Interface..." This way it is the first the user sees.

rmadsen-ks (changes requested):
one minor thing, otherwise it looks great.

Engine/Cli/CliActionExecutor.cs
frederikja163: I agree, it has been moved up now

rmadsen-ks (approved):
Looks good!

PR #583, opened 2022-05-17
329 SweepLoopError

rmadsen-ks:
Fixed an issue caused by trying to null values with a TapSerializer, which is not possible. This caused an error in the LegacySweepLoop.

The reason why this occurs with DialogStep is that the picture Source is null and not "" when it is created.


alnlarsen (approved):
Nice simple fix, works :)

PR #586, opened 2022-05-17
515 Improve documentation of external parameters

### LINKED ISSUE #515: Improve documentation of external parameters
The current settings for external parameters is a bit lacking: https://doc.opentap.io/User%20Guide/CLI%20Usage/#external-settings

It should be updated to reflect more common usage scenarios. For example, a parameter name, if the default is chosen, will usually be something like `Parameters \ Time Delay`, which of course needs to be quoted due to spaces and backslashes, but because our current examples are extremely simple, we fail to communicate that fact.

In addition, `Time Delay` has a unit attached in the GUI, so a user could be temped to say e.g. `-e "Parameters \ Time Delay=1 s"` or even `"... 1 ms"`, but this will also not work as expected because units are discarded.

In general there is room for improvement here.

alnlarsen:
There has been some confusion about the expected syntax when specifying external parameters on the CLI. Hopefully this documentation improvement will clear things up.

Closes #515

### DIFF
--- a/doc/User Guide/CLI Usage/Readme.md
+++ b/doc/User Guide/CLI Usage/Readme.md
@@ -265,37 +265,83 @@ Never prompt for user input:
 
 ### External Settings
 
-Test step settings can be marked as *External*. This means they can be set from the CLI, or from a file. This makes it
-possible to reuse the same test plan for a variety of tests.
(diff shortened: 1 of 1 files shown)


AsgerIversen:
Just beautiful

PR #588, opened 2022-05-18
Serialize performance improvements

rmadsen-ks:
Added a few performance improvements, primarily affecting deserialize/serialize.

Important changes
- TestStep.Version is now  XmlIgnore. We have Package Dependency versions, so this version should not be necessesary.
- TestStep.Enabled is default true and if it is true, it is not included in the XML
- TestStep.ChildTestSteps is default empty and when its empty, it is not included in the XML
- the TestStep has an acceleration dictionary for dynamic member data values.
- BreakConditions has the default value Inherit and when it has this value, it is not included in the XML.

Generally speaking the XML is a bit more compact now, which helps in XML deserialization performance.

XML
======
Previous:
```xml
<TestPlan type="OpenTap.TestPlan" Locked="false">
  <Steps>
    <TestStep type="OpenTap.Plugins.BasicSteps.DelayStep" Version="9.4.0-Development" Id="c8e24689-4c70-4ec6-9679-aeadf9fc6eae">
      <DelaySecs>0.1</DelaySecs>
      <Enabled>true</Enabled>
      <Name>Delay</Name>
      <ChildTestSteps />
      <BreakConditions>Inherit</BreakConditions>
    </TestStep>
  </Steps>
  <BreakConditions>Inherit</BreakConditions>
  <OpenTap.Description />
  <Package.Dependencies>
    <Package Name="OpenTAP" Version="^9.18.0-alpha.30.2+34848fa8.serializePerformanceImprovemen" />
  </Package.Dependencies>
</TestPlan>
```

Now:
```xml
<TestPlan type="OpenTap.TestPlan">
  <Steps>
    <TestStep type="OpenTap.Plugins.BasicSteps.DelayStep" Id="875987fd-2aaf-4786-80a3-da8aae0ce9fa">
      <DelaySecs>0.1</DelaySecs>
      <Name>Delay</Name>
    </TestStep>
  </Steps>
  <Package.Dependencies>
    <Package Name="OpenTAP" Version="^9.18.0-alpha.30.2+34848fa8.serializePerformanceImprovemen" />
  </Package.Dependencies>
</TestPlan>

```

Testing
===========
Testing using bin/Release/tap.exe profile --serialize-deserialize-long-plan

Before Optimization: (note, one unit test had to be modified for this to work. See the PR).

| Time (ms)|
| ----- |
| 10204 |
|  10436  |
|   10327  |
| *Average* |
|     10322     |

After Optmization:
| Time (ms)|
| ----- |
| 8184 |
|  8031  |
|   8232  |
| *Average* |
|     8149     |


alnlarsen (approved):
This all looks fine to me. It looks like test plan files are a lot smaller now.

Still, I'm not completely convinced it's safe to not serialize default values. If we ever changed the default value of `Delay.DelaySecs` for example, that would affect all test plans that use the value `0.1`. @AsgerIversen , what do you think about this?

alnlarsen:
The concern about default values is not really relevant to this PR though

AsgerIversen:
Very nice @rmadsen-ks. I like this direction. We should be careful to maintain backwards compatibility though.

Yes, changing the default value is a breaking change, so we should be careful to pick a good value, but we can remove the default value again later without breaking I would think. 

We should probably wait with this till after the 9.18 release

PR #590, opened 2022-05-18
481 Re-add the childStepsChanged event when ChildTestSteps instance is changed

### LINKED ISSUE #481: Legacy Sweep Loop Range Can Iterate paramters of deleted steps
**Environment:**

Windows 10, 64-bit

OpenTAP 9.17.3
OpenTAP 9.16.4

**Steps to reproduce:**

1. Import/Open SweepLoopRange.TapPlan and delete Dialog step then run test plan  [SweepLoopRange.zip](https://github.com/opentap/opentap/files/8322462/SweepLoopRange.zip)
2. Create a new test plan with a sweep loop(range) and a Dialog step, add Timeout sweep parameter, then delete Dialog step and run the test plan

**Actual result:**

1. Animation

2. Animation

(Please also attach a log file from %TAP_PATH%\SessionLogs\ that shows the issue when possible/relevant)

**Expected result:**

After child step deletion, Sweep parameters should be cleared and error should be shown when test plan run.

**Other remarks:**

alnlarsen:
During serialization we overwrite the value of TestStepList on teststeps
which caused issues for Sweep Loop Range because it sets up the event in its constructor.

It now listens for the PropertyChanged event and sets up the callback in this case.

Closes #481

### DIFF
--- a/BasicSteps/SweepLoopRange.cs
+++ b/BasicSteps/SweepLoopRange.cs
@@ -50,7 +50,8 @@ namespace OpenTap.Plugins.BasicSteps
         [EnabledIf("SweepBehavior", SweepBehavior.Linear, HideIfDisabled = true)]
         [XmlIgnore] // this is inferred from the other properties and should not be set by the serializer
         [Browsable(true)]
-        public decimal SweepStep {
+        public decimal SweepStep
+        {
             get
             {
                 if (SweepPoints == 0) return 0;
@@ -185,8 +186,15 @@ namespace OpenTap.Plugins.BasicSteps
             Rules.Add(() => ((SweepBehavior != SweepBehavior.Exponential)) || (Math.Sign(SweepStop) == Math.Sign(SweepStart)), "Sweep start and end value must have the same sign.", "SweepEnd", "SweepStart");
 
             ChildTestSteps.ChildStepsChanged += childStepsChanged;
+            PropertyChanged += SweepLoopRange_PropertyChanged;
         }
-        
+
+        void SweepLoopRange_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(ChildTestSteps))
+                ChildTestSteps.ChildStepsChanged += childStepsChanged;
+        }
+
         readonly Dictionary<IMemberData, object> membersCache = new Dictionary<IMemberData, object>();
         void childStepsChanged(TestStepList sender, TestStepList.ChildStepsChangedAction Action, ITestStep Object, int Index)
         {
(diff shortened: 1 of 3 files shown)


Engine/SerializerPlugins/ObjectSerializer.cs
```diff
@@ -238,12 +238,23 @@ public virtual bool TryDeserializeObject(XElement element, ITypeData t, Action<o
...
-                                        if (false == Equals(current, x))
-                                        { // for some value-like type, it may be needed
-                                            // to set the parent object when a property is changed
-                                            // example: complex test plan parameters.
-                                            setter(newobj);
+                                        // Some plugins configure event handlers for ChildTestSteps in their constructor.
+                                        // If the list is already initialized, we should add the content of the newly deserialized list to the current list instead of overwriting it.
+                                        if (x is TestStepList newList && current is TestStepList currentList)
+                                        {
+                                            currentList.AddRange(newList);
```
rmadsen-ks, line 245: If the test step constructor adds a couple of steps, would this cause some issues?

rmadsen-ks (changes requested):
I am not sure that this is the correct solution. It is just important that test step developers knows that if one listen to child steps changed, it is important to also handle changes in the getter/setter.

rmadsen-ks (approved):
Looks good!

BasicSteps/SweepLoopRange.cs
```diff
@@ -185,8 +186,15 @@ public SweepLoopRange()
...
 
             ChildTestSteps.ChildStepsChanged += childStepsChanged;
+            PropertyChanged += SweepLoopRange_PropertyChanged;
         }
-        
+
+        void SweepLoopRange_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(ChildTestSteps))
+                ChildTestSteps.ChildStepsChanged += childStepsChanged;
```
AsgerIversen, line 195: Can we remove the eventhandler from the old childsteps instance here to avoid leaking that?

AsgerIversen (changes requested):

BasicSteps/SweepLoopRange.cs
rmadsen-ks: This is not possible in a PropertyChanged event handler, but that should not be a problem, since 'this' no longer points to that instance of ChildTestSteps that event handler should be collectable by GC.

AsgerIversen: You are right. I just had it turned around in my head.

AsgerIversen (approved):

PR #596, opened 2022-05-20
Workflow fixes for pull requests from forked repos

### LINKED ISSUE #595: The `Build-DevGuide` and `Build-ApiDoc` CI jobs fail for pull requests from forked repositories
Check for instance the `CI` workflow triggered by PR #501 from [g-galante/opentap:495-update-tui-link-on-developer-guide](https://github.com/opentap/opentap/actions/runs/2269892608), or by PR #430 from [alnlarsen/opentap:main](https://github.com/opentap/opentap/actions/runs/2346130737).

The reason is explained in https://github.com/opentap/opentap/pull/430#issuecomment-1131640253.

g-galante:
Resolves #595

The fix is inspired by [Using secrets](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets), [Using encripted secrets in a workflow](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow), and [Context availability](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability).

Commit 838bf92738c5ccce5412a5a04162d8c76f872884 is for my own benefit when running workflows on the `g-galante/opentap` fork with my own local `GITLAB_REGISTRY_TOKEN`.

Not all jobs are protected from failing for missing secrets, but the fix is easily extensible to cover them in case somebody needs to test workflows in his/her own fork with only a subset of the secrets defined.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,13 +52,29 @@ jobs:
 
   CheckSecrets:
     runs-on: ubuntu-latest
+    environment: sign
     outputs:
       gitlab_registry_token_is_set: ${{ steps.check_GITLAB_REGISTRY_TOKEN.outputs.is_set }}
+      sign_is_set: ${{ steps.check_sign.outputs.is_set }}
     steps:
       - id: check_GITLAB_REGISTRY_TOKEN
         name: Check whether GITLAB_REGISTRY_TOKEN is set
         run: |
           if [ -z "${{ secrets.GITLAB_REGISTRY_TOKEN }}" ]; then
+            echo "Not set"
+            echo "::set-output name=is_set::false"
+          else
+            echo "Set"
+            echo "::set-output name=is_set::true"
+          fi
+      - id: check_sign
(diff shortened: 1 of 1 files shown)


rmadsen-ks:
Hi @g-galante, this is some really nice magic!   I must admit I dont fully understand it.

Does this `echo "::set-output name=is_set::false"` actually do something?

g-galante:
It sets the output for the job as shown in: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs. You also use it in the `GetVersion` job defined a few lines above, so I felt it didn't deserve an explanation :)

rmadsen-ks:
Ah ok, I am not really that into the specifics of Github workflows (and I did not write our workflows), but that's pretty neat!

And I guess this build contains most of what we need for now. We can add some additional improvements later.

rmadsen-ks (approved):

PR #597, opened 2022-05-23
Use get-gitversion action instead of GetVersion stage

alnlarsen:
A generic action has been implemented to get GitVersion information, so this no longer needs to be a part of the OpenTAP pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,45 +18,16 @@ env:
 
 jobs:
   GetVersion:
-    runs-on: ubuntu-latest
-    # Map a step output to a job output
+    runs-on: ubuntu-20.04
     outputs:
-      ShortVersion: ${{ steps.asmVer.outputs.ver }}
-      LongVersion: ${{ steps.longVer.outputs.ver }}
-      GitVersion: ${{ steps.gitVer.outputs.ver }}
+      ShortVersion: ${{ steps.gitversion.outputs.ShortVersion }}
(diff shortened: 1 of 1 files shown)


rmadsen-ks (approved):
Looks good!

.github/workflows/ci.yml
```diff
@@ -16,39 +16,18 @@ env:
...
-      - name: tap sdk gitversion --fields 3
-        id: asmVer
-        run: echo ::set-output name=ver::`tap sdk gitversion --fields 3`
-      - name: tap sdk gitversion --fields 4
-        id: longVer
-        run: echo ::set-output name=ver::`tap sdk gitversion --fields 4`
-      - name: tap sdk gitversion
-        id: gitVer
-        run: echo ::set-output name=ver::$(tap sdk gitversion)
+      # The get-gitversion action installs OpenTAP and fetches with fetch-depth: 0
```
AsgerIversen, line 27: Do we even need the Checkout action here?

AsgerIversen (changes requested):
Looks nice. A very nice action (i made a [PR](https://github.com/opentap/get-gitversion/pull/3) in the action with some minor improvement suggestions). 

Can we simplify the job here even further (by removing the checkout action)?

.github/workflows/ci.yml
alnlarsen: It looks like we don't. 🎆

AsgerIversen (approved):

PR #598, opened 2022-05-23
Split package create into multiple steps

alnlarsen:
package-win fails intermittently, and it is hard to tell from the logs what's going wrong.

It should at least help, if not fix the issue, if we split it into multiple steps.


rmadsen-ks (approved):
Seems to work. Lets get it in!

PR #600, opened 2022-05-23
599 remove RedirectAssembly

### LINKED ISSUE #599: tap.exe fails to launch if System.Collections.Immutable 5.0 is in a subdirectory

alnlarsen:
The RedirectAssembly method was added in 2017 to fix a bug that prevented running
tap8 from a tap7 installation, but today it has resurfaced as a bug:
https://forum.opentap.io/t/editor-exe-hangs-when-using-another-system-collections-immutable-dll-version/698/6

The bug seems to be fixed by removing this legacy workaround

Closes #599

### DIFF
--- a/Engine/PluginManager.cs
+++ b/Engine/PluginManager.cs
@@ -287,31 +287,6 @@ namespace OpenTap
             }
         }
 
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Seems to work well. I think mostly this was added for some legacy reasons for a very old version (~2017). I think it is safe to remove.

PR #604, opened 2022-05-24
603 throw OperationCanceledException on UAC cancel

### LINKED ISSUE #603: Wrong error message when canceling elevated installs
Clicking 'no' in the UAC dialog for an elevated install will cause an error log saying the download failed due to a network error. This should result in a 'User cancelled' info log.

alnlarsen:
It looks like the UAC dialog throws a Win32Exception when the user clicks 'no'. This should be interpreted as an OperationCanceledException.

Closes #603


AsgerIversen (approved):
Looks good

PR #609, opened 2022-05-25
118 expected exception

### LINKED ISSUE #118: Introduce "expected" exception
Originally filed November 04 2021 by Petr Jasek on [GitLab](https://gitlab.com/OpenTAP/opentap/-/issues/807)

It would be nice if TAP provided an exception that is recognized as "expected" and is not considered as something wrong with the code. For such exception the stack trace would not be logged (into the debug logs).

This can be useful, for example, in test step PrePlanRun, or Resource opening, when some conditions are not met so the test plan should not run at all. In this case we want to throw exception to stop everything but we do not want to make it look like there is some unexpected crash in the code.

frederikja163:
Adds an expected exception for when a teststep throws expectedly.
Used for canceling a test step wihtout logging a stacktrace to the console.
The expected exception can also set the verdict of the test-step.

Fixes #118

### DIFF
--- a/Engine/ExpectedException.cs
+++ b/Engine/ExpectedException.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OpenTap
+{
+    /// <summary>
+    /// Used to cancel a teststep without printing a stacktrace, will also set the verdice if necessary,
+    /// </summary>
+    public class ExpectedException : Exception
+    {
+        public ExpectedException(string message = "", Verdict verdict = Verdict.Error): base(message)
+        {
+            Verdict = verdict;
+        }
+
+        public Verdict Verdict { get; set; }
+
+        
+    }
+}
--- a/Engine.UnitTests/TestPlanTest.cs
+++ b/Engine.UnitTests/TestPlanTest.cs
@@ -36,6 +36,21 @@ namespace OpenTap.Engine.UnitTests
             }
         }
 
+        public class TestStepExpectedExceptionTest : TestStep
+        {
+            private Verdict _verdict;
+
+            public TestStepExpectedExceptionTest(Verdict verdict)
+            {
+                _verdict = verdict;
+            }
+
+            public override void Run()
+            {
+                throw new ExpectedException("", _verdict);
+            }
+        }
+
         public class TestStepPreExceptionTest : TestStep
         {
             public override void PrePlanRun()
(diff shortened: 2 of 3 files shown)


Engine/ExpectedException.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OpenTap
+{
+    /// <summary>
+    /// Used to cancel a teststep without printing a stacktrace, will also set the verdice if necessary,
```
rmadsen-ks, line 8:
```suggestion
    /// Used to cancel a teststep without printing a stacktrace, will also set the verdict if necessary.
```

rmadsen-ks (changes requested):
From the issue, there is mention of throwing this during PrePlanRun or resource opening. Maybe we should also add this? For example, if an instrument throws an exception during opening.

Also it seems like the intention was that this exception should be able to propagate on to stop the test plan.

rmadsen-ks (changes requested):

Engine/ExpectedException.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// The verdict of this teststep.
+        /// </summary>
+        public virtual Verdict Verdict { get; set; }
+
+        /// <summary>
+        /// Handles this exception by writing the appropriate information to the log.
+        /// </summary>
+        public void Handle(string stepName)
```
rmadsen-ks, line 28:
```suggestion
        internal void Handle(string stepName)
```

Engine/ExpectedException.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public ExpectedException(string message = "", Verdict verdict = Verdict.Error): base(message)
+        {
+            Verdict = verdict;
+        }
+
+        /// <summary>
+        /// The verdict of this teststep.
+        /// </summary>
+        public virtual Verdict Verdict { get; set; }
```
rmadsen-ks, line 23:
```suggestion
        public virtual Verdict Verdict { get; }
```

Engine/ExpectedException.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Creates a new expected exception with an optional message and verdict.
+        /// </summary>
+        public ExpectedException(string message = "", Verdict verdict = Verdict.Error): base(message)
+        {
+            Verdict = verdict;
+        }
+
+        /// <summary>
+        /// The verdict of this teststep.
```
rmadsen-ks, line 21:
```suggestion
        /// The resulting verdict of this exception.
```

Engine/ExpectedException.cs
```diff
@@ -0,0 +1,40 @@
...
+        public void Handle(string stepName)
+        {
+            if (string.IsNullOrWhiteSpace(Message))
+            {
+                TestPlan.Log.Info("Test step {0} stopped.", stepName);
+            }
+            else
+            {
+                TestPlan.Log.Info("{1}", Message);
+            }
```
rmadsen-ks, line 37: if Verdict == Error, then I guess these log messages should be written as error messages.

Engine/ExpectedException.cs
```diff
@@ -0,0 +1,40 @@
...
+        public virtual Verdict Verdict { get; set; }
+
+        /// <summary>
+        /// Handles this exception by writing the appropriate information to the log.
+        /// </summary>
+        public void Handle(string stepName)
+        {
+            if (string.IsNullOrWhiteSpace(Message))
+            {
+                TestPlan.Log.Info("Test step {0} stopped.", stepName);
```
rmadsen-ks, line 32: This is fine when the exception is thrown from TestStep.Run, but if it is PrePlanRun or PostPlanRun, maybe we should have something different.

rmadsen-ks (changes requested):
Had a few comments.

rmadsen-ks:
Lets get this in, but it needs a bit more work. Also maybe we should discuss when this feature should be used.

PR #617, opened 2022-05-30
613 Instrument2Instrument Reference error:

### LINKED ISSUE #613: instrument -> instrument reference bug
To reproduce

- Create an instrument referencing another instrument
- Removed the referenced instrument
- Serialize instruments

Observe: InstrumentSettingsXML contains the referenced instrument directly serialised in the referencing step.
Expected: Referenced instrument is not serialized.

Note: 
Actually when removing the referenced instrument from the InstrumentSettings class it should also be removed from other instruments referenced. However, it might be inserted again later, so we should  not just go and delete the reference immediately after removing it.

Two facts about instrument references:
- the initial referencing was done by selecting it from 'the' list of instruments.
- The referenced instrument used to be in a InstrumentSettings list and now is not.

We should probably trigger the behavior on either of these facts.

Also note, that the current default behavior for GUIs is not to reload all settings after settings are modified. The only time settings are generally loaded is on startup or when reverting settings changes.

This means that we'd have to modify the references after the modification is done.

rmadsen-ks:
- Added detection of using deleted resources and failing the test plan run if that is the case.
- Added annotation so that resources annotations will show an error if a deleted resources is being used.
- Added that XMlSerializer should not serialize deleted resources.
- Added a unit test to verify that it works as intended.

Closes #613

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -2007,12 +2007,26 @@ namespace OpenTap
             }
         }
 
-        class ResourceAnnotation : IAvailableValuesAnnotation, IStringValueAnnotation, ICopyStringValueAnnotation
+        
+        class ResourceAnnotation : IAvailableValuesAnnotation, IStringValueAnnotation, ICopyStringValueAnnotation, IErrorAnnotation
         {
             readonly Type baseType;
             readonly AnnotationCollection a;
-            
-            public IEnumerable AvailableValues => ComponentSettingsList.GetContainer(baseType).Cast<object>().Where(x => x.GetType().DescendsTo(baseType));
+
+            public IEnumerable AvailableValues 
+            {
+                get
+                {
+                    var x = ComponentSettingsList.GetContainer(baseType);
+                    var cv = a.Get<IObjectValueAnnotation>()?.Value as IResource;
+                    var result = x.Cast<object>().Where(y => y.GetType().DescendsTo(baseType)).ToList();
+                    // if the selected value is not in the list show it anyway.
+                    if (cv != null && result.Contains(cv) == false)
+                        result.Add(cv);
+
+                    return result;
+                }
+            }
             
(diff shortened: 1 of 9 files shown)


Engine/ILockable.cs
```diff
@@ -82,11 +82,14 @@ public interface ILockManager : ITapPlugin
...
     {
-        private List<ILockManager> managers;
+        private ILockManager[] managers;
 
         public LockManager()
         {
-            managers = PluginManager.GetPlugins<ILockManager>().OrderByDescending(t => t.GetDisplayAttribute().Order).Select(t => (ILockManager)t.CreateInstance()).ToList();
+            managers = PluginManager.GetPlugins<ILockManager>()
+                .OrderByDescending(t => t.GetDisplayAttribute().Order)
+                .Select(t => (ILockManager)t.CreateInstance())
```
alnlarsen, line 91: Can we use `TrySelect` here? CreateInstance can throw when plugins cannot be loaded for some reason, such as when a plugin depends on an incompatible runtime.

Engine/Annotations/Annotation.cs
```diff
@@ -2007,14 +2007,28 @@ public AnnotationCollection NewElement()
...
+
+                    return result;
+                }
+
+            }
             
-            public IEnumerable AvailableValues => ComponentSettingsList.GetContainer(baseType).Cast<object>().Where(x => x.GetType().DescendsTo(baseType));
-            
-            string IStringReadOnlyValueAnnotation.Value => (a.Get<IObjectValueAnnotation>()?.Value as IResource)?.ToString();
+        string IStringReadOnlyValueAnnotation.Value => (a.Get<IObjectValueAnnotation>()?.Value as IResource)?.ToString();
```
alnlarsen, line 2031: It looks like the indentation ran away from you around here :)

alnlarsen (changes requested):
Minor changes needed. Let's merge this soon! 👍

PR #618, opened 2022-05-30
Add secret flag to disable isolation

alnlarsen:
This adds a secret flag to disable running isolated. Currently, when our pipelines fail due to Sign not being installed correctly, it is difficult to debug because there is no log output from the isolated process. 

We can't even tell if the problem happens inside the isolated process, or if the isolation logic fails for some reason on the runner.

With this change, if the problem does occur again, we at least have a chance to debug the issue from the logs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -454,7 +454,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_SECRET }}
       - name: Install Sign
         working-directory: bin/Release-x64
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks good. Lets hope this fixes the issues on the main branch.

sebastian-pop:
@rmadsen-ks this option is visible in help. Is it supposed to be hidden?

image

alnlarsen:
@sebastian-pop  yes this is supposed to be hidden.

PR #619, opened 2022-05-30
Catch exceptions when construction ICustomPackageAction typedatas

### LINKED ISSUE #620: Package install fails if ICustomTypeData inheritors throw in their constructor
We should catch exceptions occurring during ICustomTypeData constructors.

alnlarsen:
Closes #620


rmadsen-ks (approved):
Looks good

PR #624, opened 2022-05-31
623 Package Name Exists Printed Twice: Fix

### LINKED ISSUE #623: "PackageName exists in cache:..." printed out twice
image

rmadsen-ks:
Fixed the issue by not printing the logmessage inside the CachedLocation method, but instead printed when it did not return null. The problem is that this is used more than once in some situations.

Closes #623


Denrask (approved):
Looks good!

Issue #629, opened 2022-06-01
Annotation with null lists throws an exception


rmadsen-ks:
It should be possible to just create a list instance when it is needed.

In any case, an exception should not be thrown when trying to edit it.

razvancoste:
Verified testing this issue : https://github.com/opentap/opentap/issues/628

Issue #628, opened 2022-06-01
Multi-Selecting list-types is not working


rmadsen-ks:
Currently an exception is thrown if two test steps are selected and they have a list of editable items.

Since out parameter system can support this, we should also be able to support it for multi-select.

razvancoste:
Verified using OpenTAP 9.18.0-beta.47

![image](https://user-images.githubusercontent.com/98153555/171685209-ec0e15af-8af1-4e11-b9a1-c1aef81710d4.png)

PR #630, opened 2022-06-01
628 629 List Annotation Issues: Fix

rmadsen-ks:
Added unit tests for 628 and 629.

Both issues are related to steps that has lists, which is why it makes sense to fix them together.


alnlarsen:
Does this close #629 and #628 ?

rmadsen-ks:
> Does this close #629 and #628 ?

Yes, that is the intention.

PR #635, opened 2022-06-07
Handle exceptions instead of relying on an empty catch

alnlarsen:
This is a nice little usability fix that avoids logging FirstChanceExceptions


alnlarsen:
Rolf approved this verbally

AsgerIversen (approved):

PR #639, opened 2022-06-07
Debian permission fixes

### LINKED ISSUE #637: linux: .lock files created with sudo belong to root
If no `.lock` file exists in an installation, and it is created from an elevated process, the resulting `.lock` file will not be writable by normal users. This is a problem because it will cause all future IsolatedPackageAction CLI actions to require elevation.

We should ensure that the permissions on this lock file are a bit more relaxed.

### LINKED ISSUE #638: Deb package rename: Group and install folder from 'tap' to 'OpenTAP'
The current group name for the OpenTAP installation is 'tap'. It should be renamed to 'OpenTAP'.

The install folder is also '/usr/share/tap' which is a bit of a mismatch from the Windows equivalent, 'C:/Program Files/OpenTAP'. The install folder should be renamed to '/usr/share/OpenTAP'.

### LINKED ISSUE #643: Deb package: permission issues
The deb package has some serious usability issues right now due to permission issues. If a tap process creates files when run with 'sudo', those files belong to root and cannot be deleted by user processes.

All files should be writable by the 'OpenTAP' group to prevent this kind of error.

### LINKED ISSUE #647: deb package: multiuser usability issues
If a user who does not belong to the OpenTAP group, the error message from OpenTAP is not very helpful. It would be good if OpenTAP detected the user does not belong to the owner group, and provided instructions for adding the user.

alnlarsen:
This renames the 'tap' user group to 'OpenTAP',
and updates the install location from '/usr/share/tap' to /usr/share/OpenTAP'.

It also includes a bugfix which prevented isolated package actions from working
if the .lock file was initially created by root.

Closes #637  
Closes #638 
Closes #643 
Closes #647

### DIFF
--- a/LinuxInstall/INSTALL.sh
+++ b/LinuxInstall/INSTALL.sh
@@ -44,7 +44,7 @@ echo "Unzipping tap"
 unzip "$(pwd)/*.TapPackage" -d "$DEST_DIR" # *: match OpenTAPLinux and just TAPLinux.
 
 cd "$DEST_DIR" || exit 1
-chmod -R +w .
+chmod -R +rws "$DEST_DIR"
 
 chmod +x tap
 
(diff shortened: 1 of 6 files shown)


rmadsen-ks (approved):
Seems to work.

AsgerIversen (approved):

Issue #505, opened 2022-04-07
Create dotnet templates from SDK templates


alnlarsen:
We currently have a lot of SDK templates for creating new projects. These are contained in the SDK package, and the source code is in the OpenTAP repository. For example, the CLI command `tap sdk new project MyAwesomeProject` creates a new C# solution and replaces template placeholders with the name `MyAwesomeProject` in a number of files.

This all works very well, but it requires
1. An existing OpenTAP installation
2. The SDK package to be installed

In hindsight, we probably should have implemented these templates as dotnet templates instead. I propose converting all of our `tap sdk` templates to `dotnet new` templates. So instead of typing `tap sdk new project MyAwesomeProject`, we could type `dotnet new -i OpenTap.Project MyAwesomeProject`. As far as I understand, dotnet templates can also be used in Visual Studio and Rider, so we get the immediate benefit of 
1. Being able to integrate the templates in popular C# IDEs
2. Only needing dotnet to be installed

PR #648, opened 2022-06-10
505 templates for dotnet new

frederikja163:
Makes dotnet new templates to implement the feature requested at #505 

The templates feature optional starting ressources like, TestSteps, Duts, CliActions etc.
The templates also allows the user to pick and choose between three different default launch settings, TUI, Editor and Editor CE.


rmadsen-ks:
@frederikja163, awesome stuff! Can you add to this PR how this  can be tested? Do I need to add some nuget package?

rmadsen-ks (dismissed):
This seems like it needs to be rebased. 

Could you also add some information about how to test it?

frederikja163:
To test this you can run the following commands:

**Installing**
```bash
dotnet pack ./templates/;
dotnet new -i ./bin/Release/OpenTap.Templates.1.0.0.nupkg;
```

**Uninstalling**
```bash
dotnet new -u OpenTap.Templates
```

**Listing opentap templates**
```bash
dotnet new -l opentap
```

Once installed visual studio automatically adds the templates to its project creation and solution creation lists.
It can also be installed using `dotnet new` with the same options as in the IDE.

templates/src/project/sln/Directory.Build.Props
```diff
@@ -0,0 +1,14 @@
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup Condition="'$(Configuration)' == ''">
+        <Configuration>Debug</Configuration>
+    </PropertyGroup>
+        
+    <PropertyGroup>
+        <OpenTapVersion>9.19.0</OpenTapVersion>
```
alnlarsen, line 7: Should this be $(GitVersion)

alnlarsen (changes requested):
I find it pretty weird that `Directory.Build.Props` appears in the parent folder of the folder where I create my project. I also find it weird that there's a `sln` folder in my project. It looks like `sln` was supposed to be excluded, but I see it both on Linux and Windows when using the dotnet cli.

When using the template in Visual Studio, everything seems to work correctly if I create a new solution, but if I tick the option to put the soluton and project in the same directory, then I also get `Directory.Build.Props` in the parent folder, which is outside of the project root in this case.

I like the templates, but this seems to be a bit of an issue.

Package/XmlEvaulation/IVariableProvider.cs
```diff
@@ -218,31 +218,41 @@ public GitVersionExpander(string projectDir)
...
         private string version = null;
 
         private static TraceSource log = Log.CreateSource(nameof(GitVersionExpander));
 
         public void Expand(XElement element)
         {
-            if (!element.ToString().Contains("$(GitVersion)")) return;
-            if (version == null && string.IsNullOrWhiteSpace(ProjectDir) == false)
+            ReplaceVersion("GitVersion", 5, ref version);
+            ReplaceVersion("GitLongVersion", 4, ref longVersion);
```
alnlarsen, line 230: I don't really understand why we need GitLongVersion here

frederikja163: Pretty sure its there because nuget packages are named with LongVersion, so in order to install/include the nuget packages in the templates we need a way to write the GitlongVersion inside a file. I thought it would be best to just add it to the system already expanding the GitVersion, instead of doing it from scratch.

This is done in `sdk/sdk.package.xml` for example

templates/src/project/sln/Directory.Build.Props
frederikja163: Possibly, i honestly dont remember if there was a reason to leave it like this. Maybe because we want the lowest possible tap version for templates to allow more backwards compatibility?

frederikja163:
> I find it pretty weird that `Directory.Build.Props` appears in the parent folder of the folder where I create my project. I also find it weird that there's a `sln` folder in my project. It looks like `sln` was supposed to be excluded, but I see it both on Linux and Windows when using the dotnet cli.
> 
> When using the template in Visual Studio, everything seems to work correctly if I create a new solution, but if I tick the option to put the soluton and project in the same directory, then I also get `Directory.Build.Props` in the parent folder, which is outside of the project root in this case.
> 
> I like the templates, but this seems to be a bit of an issue.

This is all due to a few weirdnesses with the whole dotnet new templating engine. It is super inconsistent between the GUI in VS and Rider and the CLI. There isn't really anything we can do about that apart from removing the Directory.Build.Props file yet again. So far i think we have gone back and forth on removing that file like 5 different times. The current reasoning for keeping it as far as i can remember is because this is the way it was showcased to a few people and we didnt want to change behaviour from the demo. However i think by now the demo is so long ago that they dont remember that specific detail anymore.

Anyways, the only real solution to this whole thing is to remove the Directory.Build.Props file, which would mean that projects all generate their own bin directory instead of using a shared one. Thats really the only downside. However this does mean it would function more like a usual C# project with all the projects having their own bin directory.

alnlarsen:
@AsgerIversen I have made a few adjustments to the PR to resolve some issues that appeared after Frederik removed Directory.Build.props, and an issue with the SDK action where SDK would fail to install if the templates were already installed.

alnlarsen (approved):
I think this is better than our old templates now.

PR #650, opened 2022-06-13
649 Dispose Root Session: Fix

rmadsen-ks:
- added the possibility of disposing the root session. This can be added to applications running tap.
- Added a test step for testing it.
- Added a regression testplan for testing it.


rmadsen-ks:
Nvm, this actually does not fix the issue i had. Might be useful for other things though.

PR #652, opened 2022-06-14
InformativePropertyException

### LINKED ISSUE #606: Property Exceptions when loading should be more informative
To reproduce:
1. create a test plan with a delay step. Save as a file.
2. Open the file in a text editor.
3. Change `<DelaySecs>0.1</DelaySecs>` to `<DelaySecs>-1</DelaySecs>`
4. Open the file in OpenTap

Expected result: the problem should be stated in the error message.

Actual result: The problem is hidden in the stack trace.

```
10:10:29.393  Serializer   XML Line 11: Object value was not read correctly.
10:10:29.393  Serializer   TargetInvocationException: Exception has been thrown by the target of an invocation.
10:10:29.393  Serializer       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
10:10:29.393  Serializer       at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
10:10:29.393  Serializer       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
10:10:29.393  Serializer       at System.Reflection.RuntimePropertyInfo.SetValue(Object obj, Object value, Object[] index)
10:10:29.393  Serializer       at OpenTap.MemberData.SetValue(Object owner, Object value)
10:10:29.393  Serializer       at OpenTap.Plugins.ObjectSerializer.<>c__DisplayClass12_3.<TryDeserializeObject>b__10(Object x)
10:10:29.393  Serializer       at OpenTap.Plugins.ObjectSerializer.Deserialize(XElement element, ITypeData t, Action`1 setter)
10:10:29.393  Serializer     Inner ArgumentException: Delay must be a positive value.
10:10:29.393  Serializer       at OpenTap.Plugins.BasicSteps.DelayStep.set_DelaySecs(Double value)
10:10:29.393  Serializer   Exception caught at:
10:10:29.393  Serializer       at System.Object Deserialize(System.Xml.Linq.XDocument, OpenTap.ITypeData, Boolean, System.String)
10:10:29.393  Serializer       at System.Object DeserializeFromFile(System.String, OpenTap.ITypeData, Boolean)
10:10:29.393  Serializer       at OpenTap.TestPlan b(System.String, System.String[], Boolean)
10:10:29.393  Serializer       at Void a(System.String, System.String[], Boolean)

```

image

frederikja163:
Fixed #606 by printing exception message when loading tapplan, instead of "Object value was not read correctly".

Im not sure if it might be worth it to print the exception message as the default, instead of only doing it in specific cases. I could imagine the exception would almost always contain more detailed and better information than any message we can write in the object serializer.

### DIFF
--- a/Engine/SerializerPlugins/ObjectSerializer.cs
+++ b/Engine/SerializerPlugins/ObjectSerializer.cs
@@ -592,7 +592,14 @@ namespace OpenTap.Plugins
             }
             catch(Exception ex)
             {
-                Serializer.PushError(element, $"Object value was not read correctly.", ex);
+                if (ex is TargetInvocationException tarEx && tarEx.InnerException is ArgumentException arEx)
+                {
+                    Serializer.PushError(element, arEx.Message, tarEx);
+                }
+                else
+                {
+                    Serializer.PushError(element, $"Object value was not read correctly.", ex);
+                }
                 return false;
             }
             try
--- a/Engine/SerializerPlugins/ObjectSerializer.cs
+++ b/Engine/SerializerPlugins/ObjectSerializer.cs
@@ -592,9 +592,9 @@ namespace OpenTap.Plugins
             }
             catch(Exception ex)
             {
-                if (ex is TargetInvocationException tarEx && tarEx.InnerException is ArgumentException arEx)
+                if (ex is TargetInvocationException tarEx)
                 {
-                    Serializer.PushError(element, arEx.Message, tarEx);
+                    Serializer.PushError(element, tarEx.InnerException.Message, tarEx);
                 }
                 else
                 {
--- a/Engine/SerializerPlugins/ObjectSerializer.cs
+++ b/Engine/SerializerPlugins/ObjectSerializer.cs
@@ -594,7 +594,7 @@ namespace OpenTap.Plugins
             {
                 if (ex is TargetInvocationException tarEx)
                 {
-                    Serializer.PushError(element, tarEx.InnerException.Message, tarEx);
+                    Serializer.PushError(element, tarEx.InnerException.Message, tarEx.InnerException);
                 }
                 else
                 {


Engine/SerializerPlugins/ObjectSerializer.cs
```diff
@@ -592,7 +592,14 @@ public override bool Deserialize(XElement element, ITypeData t, Action<object> s
             }
             catch(Exception ex)
             {
-                Serializer.PushError(element, $"Object value was not read correctly.", ex);
+                if (ex is TargetInvocationException tarEx && tarEx.InnerException is ArgumentException arEx)
```
rmadsen-ks, line 595: Maybe we should just unwrap all TargetInvocationExceptions not only when the inner exception is ArgumentException?

rmadsen-ks (changes requested):

rmadsen-ks:
Looks good!

Engine/SerializerPlugins/ObjectSerializer.cs
```diff
@@ -592,7 +592,14 @@ public override bool Deserialize(XElement element, ITypeData t, Action<object> s
             }
             catch(Exception ex)
             {
-                Serializer.PushError(element, $"Object value was not read correctly.", ex);
+                if (ex is TargetInvocationException tarEx)
+                {
+                    Serializer.PushError(element, tarEx.InnerException.Message, tarEx);
```
rmadsen-ks, line 597: Maybe this would be better?

```suggestion
                    Serializer.PushError(element, tarEx.InnerException.Message, tarEx.InnerException);
```

rmadsen-ks (changes requested):

PR #660, opened 2022-06-15
659 Add IStringValueAnnotation implementation for DateTime

### LINKED ISSUE #659: `DateTime` fields are not handled properly by OpenTAP GUI
### Description

DateTime fields that are changed by the user in the GUI, do not update the underlying variable, rendering these GUI elements unusable.

### Environment

Windows 10.0.19044 Build 19044, 64-bit
Test Automation 9.17.4+5e508f08 (64-bit) Official Release

### To reproduce

- Open the OpenTap Test Automation with the SDK Examples compiled.
- Go to Settings > Example Settings
- Change the value of the Creation Time field.
- Move to another tab and back to the Example Settings. (Or Press OK to close, and reopen the settings again)
- Observe the Creation Time again shows the initial (default) value, NOT the manually entered value.

### Expected Result
The value entered by the user should be stored and usable by the underlying code.

### Actual Result
The value entered by the user is ignored, and the default value (or a programmatically changed value) is used.

### Other remarks
When using a DateTime field in our own plugins, we observe that the value entered by the user in the GUI does not get stored in the underlying variable. It is possible to change the underlying variable programmatically, and this will be reflected in the value shown by the GUI, but the value changed in the GUI never gets used.

alnlarsen:
This adds a IStringValueAnnotation implementation for DateTime. We already have one for TimeSpan, so I think it makes sense to add one for DateTime as well.

Closes #659

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -528,6 +528,31 @@ namespace OpenTap
         public IEnumerable<string> Errors => currentError == null ? Array.Empty<string>() : new[] { currentError };
     }
 
+    class DateTimeAnnotation : IStringValueAnnotation, ICopyStringValueAnnotation
+    {
+        private AnnotationCollection annotation;
+
+        public string Value
+        {
+            get
+            {
+                if (annotation.Get<IObjectValueAnnotation>(from: this).Value is DateTime dt)
+                    return dt.ToString(CultureInfo.CurrentCulture);
+                return "";
+            }
+            set
+            {
+                if (DateTime.TryParse(value, out var dt))
+                    annotation.Get<IObjectValueAnnotation>(from: this).Value = dt;
+            }
+        }
+
+        public DateTimeAnnotation(AnnotationCollection annotation)
+        {
+            this.annotation = annotation;
+        }
+    }
+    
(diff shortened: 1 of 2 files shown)


Engine/Annotations/Annotation.cs
```diff
@@ -528,6 +528,31 @@ public NumberAnnotation(AnnotationCollection mem)
...
+        {
+            get
+            {
+                if (annotation.Get<IObjectValueAnnotation>(from: this).Value is DateTime dt)
+                    return dt.ToString(CultureInfo.CurrentCulture);
+                return "";
+            }
+            set
+            {
+                if (DateTime.TryParse(value, out var dt))
```
rmadsen-ks, line 545: Here it would be better to do DateTime.Parse, since that would cause a FormatException in case the string is not a valid date time and that turns into a notification for the user.

rmadsen-ks (changes requested):

rmadsen-ks (changes requested):
Could this new class implement IErrorAnnotation, so that we get good error messages when the user types in some invalid date time?

alnlarsen:
@rmadsen-ks that's a terrific idea

rmadsen-ks:
Changed the date time format for more sanity

alnlarsen:
Awesome change!!

rmadsen-ks (approved):

PR #666, opened 2022-06-20
Set AssemblyVersion using the AssemblyInformationalVersion attribute

### LINKED ISSUE #665: 9.18.1: AssemblyData.SemanticVersion not correct
The `AssemblyData.SemanticVersion` for `OpenTap.dll` and `OpenTap.Package.dll` in 9.18.1 return 9.18.0. It looks like the `SemanticVersion.TryParse` call in AssemblyData always fails because `RawVersion` is not actually set to a semantic-version compliant string.

alnlarsen:
Closes #665


rmadsen-ks (approved):
Looks good!

PR #668, opened 2022-06-21
655 TestStep Break Exception

### LINKED ISSUE #655: TestStepBreakException due to break condition from child test steps not catchable in parent
This was introduced by commit 071bd21614acdb3aa1bf0c4f64c89d3b4f34baea (#292)

rmadsen-ks:
This PR fixes a breaking change introduced in 9.17.

If the behavior from 9.18 is wanted new API points can be used for that.

Closes #655

### DIFF
--- a/Engine/TestStep.cs
+++ b/Engine/TestStep.cs
@@ -15,6 +15,8 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Threading;
 using System.Runtime.CompilerServices;
+using System.Runtime.ExceptionServices;
+
 namespace OpenTap
 {
     /// <summary>
@@ -460,6 +462,15 @@ namespace OpenTap
         {
             return RunChildSteps(attachedParameters, CancellationToken.None);
         }
+        
+        /// <summary>
+        /// Runs all enabled <see cref="TestStep.ChildTestSteps"/> of this TestStep. Upgrades parent verdict to the resulting verdict of the childrens run. Throws an exception if the child step does not belong or isn't enabled.
+        /// </summary>
+        /// <param name="throwOnBreak">Whether an exception will be thrown due to break conditions or if they will be caught. Exceptions are still available on child test steps TestStepRun.Exception. </param>
+        protected IEnumerable<TestStepRun> RunChildSteps(bool throwOnBreak)
+        {
+            return RunChildSteps(null, CancellationToken.None, throwOnBreak);
+        }
 
         /// <summary>
         /// Runs all enabled <see cref="TestStep.ChildTestSteps"/> of this TestStep. Upgrades parent verdict to the resulting verdict of the childrens run. Throws an exception if the child step does not belong or isn't enabled.
@@ -468,7 +479,18 @@ namespace OpenTap
         /// <param name="cancellationToken">Provides a way to cancel the execution of child steps before all steps are executed.</param>
         protected IEnumerable<TestStepRun> RunChildSteps(IEnumerable<ResultParameter> attachedParameters, CancellationToken cancellationToken)
         {
-            return this.RunChildSteps(PlanRun, StepRun, attachedParameters, cancellationToken);
+            return RunChildSteps(attachedParameters, cancellationToken, true);
+        }
+
+        /// <summary>
+        /// Runs all enabled <see cref="TestStep.ChildTestSteps"/> of this TestStep. Upgrades parent verdict to the resulting verdict of the childrens run. Throws an exception if the child step does not belong or isn't enabled.
+        /// </summary>
+        /// <param name="attachedParameters">Parameters that will be stored together with the actual parameters of the steps.</param>
+        /// <param name="cancellationToken">Provides a way to cancel the execution of child steps before all steps are executed.</param>
+        /// <param name="throwOnBreak">Whether an exception will be thrown due to break conditions or if they will be caught. Exceptions are still available on child test steps TestStepRun.Exception. </param>
+        protected IEnumerable<TestStepRun> RunChildSteps(IEnumerable<ResultParameter> attachedParameters, CancellationToken cancellationToken, bool throwOnBreak)
+        {
+            return this.RunChildSteps(PlanRun, StepRun, attachedParameters, cancellationToken, throwOnBreak);
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+    public override void Run()
+    {
+        // Insert here things to do before running child steps 
+        RunChildSteps();
+        // Insert here things to do after running child steps.
+    }
+```
+
+`RunChildSteps` will take care of setting the verdict of the calling test step based on the verdict of the child test steps. The default algorithm here is to take the most severe verdict and use that for the parent test step. 
+For example, if the child test steps has verdict Pass, Inconclusive and Fail, the parent test step will get the Fail verdict.
```
alnlarsen, line 357:
```suggestion
For example, if the child test steps have verdicts Pass, Inconclusive and Fail, the parent test step will get the Fail verdict.   
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+        // Insert here things to do before running child steps 
+        RunChildSteps();
+        // Insert here things to do after running child steps.
+    }
+```
+
+`RunChildSteps` will take care of setting the verdict of the calling test step based on the verdict of the child test steps. The default algorithm here is to take the most severe verdict and use that for the parent test step. 
+For example, if the child test steps has verdict Pass, Inconclusive and Fail, the parent test step will get the Fail verdict.   
+
+`RunChildSteps` may throw an exception, this happens if the break conditions for the parent test step are satisfied.
```
alnlarsen, line 359:
```suggestion
`RunChildSteps` may throw an exception. This happens if the break conditions for the parent test step are satisfied. 
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+        RunChildSteps();
+        // Insert here things to do after running child steps.
+    }
+```
+
+`RunChildSteps` will take care of setting the verdict of the calling test step based on the verdict of the child test steps. The default algorithm here is to take the most severe verdict and use that for the parent test step. 
+For example, if the child test steps has verdict Pass, Inconclusive and Fail, the parent test step will get the Fail verdict.   
+
+`RunChildSteps` may throw an exception, this happens if the break conditions for the parent test step are satisfied. 
+If the break conditions are satisfied due to an exeception being thrown, resulting in an Error verdict, the same exception will be thrown.
```
alnlarsen, line 360:
```suggestion
If the break conditions are satisfied due to an exception being thrown, resulting in an Error verdict, the same exception will be thrown.
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+To avoid this behavior the overload `RunChildSteps(throwOnBreak: false)` can be used, but the exceptions are still available from the returned list of test step runs.
```
alnlarsen, line 361:
```suggestion
To avoid this behavior, the overload `RunChildSteps(throwOnBreak: false)` can be used, but the exceptions are still available from the returned list of test step runs.
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+
+To override a verdict set by `RunChildSteps` the exception need to be caught or `throwOnBreak` needs to be set. After this the `Verdict` property can be set directly.
```
alnlarsen, line 363:
```suggestion
To override a verdict set by `RunChildSteps`, the exception must be caught, or `throwOnBreak` must be set. After this, the `Verdict` property can be set directly.
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
+```cs
+    public override void Run()
+    {
+        RunChildSteps(throwOnBreak: false);
+        Verdict = Verdict.Pass; // force the verdict to be 'Pass'
+    }
+```
+
+Child test steps can be run in a separate thread. This should be done with the TapThread.Start method. Before returning from the parent step the child step thread should be waited for or alternatively this should be done in a Defer operation. A great example of using parallelism can be found in the [Parallel Step](https://github.com/opentap/opentap/blob/main/BasicSteps/ParallelStep.cs).
```
alnlarsen, line 372:
```suggestion
Child test steps can be run in a separate thread. This should be done with the TapThread.Start method. Before returning from the parent step, the child step thread should be waited for, or alternatively this should be done in a Defer operation. A great example of using parallelism can be found in the [Parallel Step](https://github.com/opentap/opentap/blob/main/BasicSteps/ParallelStep.cs).
```

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -334,7 +335,42 @@ For different approaches to publishing results, see the examples in:
...
 -	`TAP_PATH\Packages\SDK\Examples\PluginDevelopment\TestSteps\PublishResults`
 
-	
+## Child Test Steps
+
+Test step can have any number of child test steps. Exactly which can be controlled by using `AllowAnyChildAttribute`, and `AllowChildOfType`. 
+
+To execute child test steps within a test step run, the RunChildStep method can be used to run a single child step or RunChildSteps can be used to run all of them.
+
+Depending on the verdict of the child steps, the break condition setting of the parent step and the way RunChildSteps is called there are a number of different ways that the control flow will be affected.
```
alnlarsen, line 344:
```suggestion
Depending on the verdict of the child steps, the break condition setting of the parent step, and the way RunChildSteps is called, there are a number of different ways that the control flow will be affected.
```

Engine/TestStep.cs
```diff
@@ -707,12 +747,17 @@ public static IEnumerable<TestStepRun> RunChildSteps(this ITestStep Step, TestPl
...
                             }
                         }
-                        // if skip to next step, dont add it to the wait queue.
+                        // if skip to next step, don't add it to the wait queue.
                     }
                     if (run.BreakConditionsSatisfied())
                     {
                         run.LogBreakCondition();
-                        break;
+                        if (throwOnBreak)
```
alnlarsen, line 755: Do we need to break here if `throwOnBreak == false` ?

alnlarsen (changes requested):

alnlarsen:
I think the implementation looks good. I just have one question and a few corrections.

Engine/TestStepRun.cs
```diff
@@ -59,7 +59,10 @@ public Verdict Verdict
             }
             protected internal set => Parameters.SetIndexed((nameof(Verdict), GROUP), ref verdictIndex, value);
         }
-
+        
+        /// <summary> Exception causing the Verdict to be 'Error' in any. </summary>
```
rmadsen-ks, line 63:
```suggestion
        /// <summary> Exception causing the Verdict to be 'Error'. </summary>
```

PR #670, opened 2022-06-21
Update OpenTapApiReference copyright

razvancoste:
updated copyright to 2022


alnlarsen (approved):
Looks good. It looks like this is the only location where the copyright year is stated in the API reference, so I guess this is fine for now, but we also have copyright years in the headers of a ton of .cs source files.

PR #672, opened 2022-06-22
671 ChildStep Visibility Default To Collapsed:

rmadsen-ks:
- Changed the behavior so the visibility of child items defaults to collapsed.
- Modified the test to verify this behavior.


alnlarsen (approved):
This is fine by me. Let's do it.

PR #677, opened 2022-06-27
Fix case of Debian package artifact

### LINKED ISSUE #678: Debian artifact package not uploaded
It looks like our artifact path is wrong. Let's fix it!

alnlarsen:
I'm not sure when we broke this, but let's fix it

Closes #678


rmadsen-ks (approved):
Looks good!

PR #679, opened 2022-06-27
fix incorrect assembly versions for dependencies folders

### LINKED ISSUE #676: Dependencies DLL versions not detected correctly
In 9.18.2 when checking the Dependencies folder, `DotNet.Glob` is placed in a folder named `DotNet.Glob.3.0.0.1` which is strange when you consider that the actual version we ship is `3.0.1.0`. 

This is probably a regression from #666

alnlarsen:
Try to parse AssemblyVersion as Version in case AssemblyInformationalVersion cannot be parsed as an AssemblyVersion.

Closes #676

### DIFF
--- a/.gitversion
+++ b/.gitversion
@@ -3,7 +3,7 @@
 
 # This is the version number that will be used. Prerelease numbers are calculated by 
 # counting git commits since the last change in this value.
-version = 9.18.2
+version = 9.18.3
 
 # A version is determined to be a "beta" prerelease if it originates from the default branch
 # The default branch is the first branch that matches the following regular expession.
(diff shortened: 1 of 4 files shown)


rmadsen-ks (approved):
Looks great! I verified that the package now contains the same dependencies from 9.18.1.

image

rmadsen-ks (changes requested):
Lets change it so that the AssemblyData.Version is just a `{get; internal set;}`. Lets remove the part about Semantic version in that context.

rmadsen-ks (approved):
Looks good.

PR #685, opened 2022-06-30
Added Host config in Authentication settings and method to retrieve p…

Denrask:
…reconfigured HttpClient.

I've tested using the following code in a test (however, I do not want to commit a test in CI that depends on a repository or internet connectivity)

```csharp
        [Test]
        public void RelativeUrl()
        {
            string host = REPOSITORY;
            // REST-API
            AuthenticationSettings.Current.BaseAddress = host;


            // Plugin in REST-API Process
            HttpClient client = AuthenticationSettings.Current.GetClient();
            Assert.AreEqual(host, client.BaseAddress.ToString());

            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Get, "/packages/3.0/GetPackageNames");
            requestMessage.Headers.Accept.Clear();
            requestMessage.Headers.Accept.Add(new System.Net.Http.Headers.MediaTypeWithQualityHeaderValue("application/json"));

            var response = client.SendAsync(requestMessage).GetAwaiter().GetResult();
            List<string> packageNames = JsonConvert.DeserializeObject<List<string>>(response.Content.ReadAsStringAsync().GetAwaiter().GetResult());   
            Assert.IsTrue(packageNames.Any());
        }
```

### DIFF
--- a/Engine.UnitTests/AuthenticationTests.cs
+++ b/Engine.UnitTests/AuthenticationTests.cs
@@ -57,10 +57,9 @@ namespace OpenTap.UnitTests
         [Test]
         public void RelativeUrl()
         {
-            
             string host = "https://thisdoesnotexist.io/";
             // REST-API
-            AuthenticationSettings.Current.Host = host;
+            AuthenticationSettings.Current.BaseAddress = host;
 
 
             // Plugin in REST-API Process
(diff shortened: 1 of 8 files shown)


AsgerIversen (approved):
Looks good. Small reduction in new APIs that we never should have added in 9.18.0

PR #690, opened 2022-07-05
689 stop shipping netstandard

### LINKED ISSUE #689: Stop shipping netstandard.dll on Linux
It looks like we have been shipping some .net framework version of 'netstandard.dll' which consists entirely of `TypeForwardedTo(...)` implementations pointing to the .NET Framework 462 implementation. This is causing issues for some plugins, so let's just stop shipping it. 

I investigated why it is being shipped in the first place, and it looks like it is being copied to the output directory when we build the MSBuild SDK project.

alnlarsen:
Maybe this hardcoded solution isn't the best.

Closes #689

### DIFF
--- a/Package/CreatePackage/Package.cs
+++ b/Package/CreatePackage/Package.cs
@@ -491,6 +491,13 @@ namespace OpenTap.Package
                         // add them as payload in this package in the Dependencies subfolder
                         foreach (var unknown in dependentAssemblyNames)
                         {
+                            if (unknown.Name == "netstandard")
+                            {
+                                // Packaging netstandard.dll can cause issues for some plugins, so let's always skip it.
+                                // It should always be packaged with the runtime anyway.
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks good!

AsgerIversen:
@alnlarsen is this problem specific to netstandard.dll? I would imagine that if MSBuild put netstandard.dll in the output folder, it also put some other dlls there (some of the ones netstandard.dll forwards types to).

alnlarsen:
@AsgerIversen I don't know. Removing netstandard.dll here is solving a concrete problem, but I think this is a pretty suspect fix, but in this case we know that netstandard.dll is **always** provided by the runtime.

It would be a good idea to launch an investigation and find a more general fix when we have the time.

alnlarsen:
I prefer #791

PR #693, opened 2022-07-05
Use correct base64-url encoding when decoding the payload of a JWT token

AsgerIversen:


Denrask (dismissed):
This looks good to me!

Denrask (approved):

PR #697, opened 2022-07-06
TokenInfo now holds an access_token and possible also a refresh_token…

AsgerIversen:
… in the same object making it much easier to find the refresh token to use if the access token has expired.


Engine/Authentication/TokenInfo.cs
```diff
@@ -67,15 +87,29 @@ public TokenInfo()
...
+        public TokenInfo(string access_token, string refresh_token, string domain)
         {
-            TokenData = jwtString;
-            Type = type;
+            AccessToken = access_token;
+            RefreshToken = refresh_token;
             Domain = domain;
         }
 
+        /// <summary> Parses tokens from OAuth response string (json format) and adds them to current Tokens list. </summary>
```
Denrask, line 100: They are no longer added to the current tokens list

Engine.UnitTests/AuthenticationTests.cs
```diff
@@ -39,8 +39,8 @@ public void ParseTokens()
     ""session_state"": ""05d7a6f2-70dd-48e3-806b-dc9ccb5e7f3e"",
     ""scope"": ""openid profile email""
 }";
-            AuthenticationSettings.Current.AddTokensFromResponse(response, "http://packages.opentap.io");
-            Assert.AreEqual(2, AuthenticationSettings.Current.Tokens.Count);
+            ti = TokenInfo.FromResponse(response, "http://packages.opentap.io");
+            Assert.IsNotNull(ti.RefreshToken);
             AuthenticationSettings.Current.Tokens.Clear();
```
Denrask, line 44: We dont need to clear anymore, as they are never added

Denrask (changes requested):
Minor comments

PR #709, opened 2022-07-13
Modified PrepareRequest logic to more reliably add header tokens base…

Denrask:
…d on own BaseAddress


AsgerIversen:
After discussing with @Denrask we agree that it is pretty much the way we want it already. Tokens are not distinguished by the entire BaseAddress (a full URI) but only the host part.

We'll try to enforce that more, to avoid this confusion: https://github.com/opentap/opentap/pull/712

PR #715, opened 2022-07-15
460 dependency resolver with graph ql

### LINKED ISSUE #460: Transitive dependency resolution bug
# Problem

We have some serious issues related to dependency resolution particularly surrounding packages that use the 'any' version specifier. This has recently caused projects that use OpenTAP 9.17 and "Developer's System" to not build at all due to the SDK package resolving to an RC version which depends on an incompatible version of OpenTAP, which is problematic because this is a very common scenario. The same issue can easily arise in many different scenarios, and it is critical that we get this resolved. It is easy to construct the exact scenario where the issue arises:

Imagine an image which asks for a specific version of OpenTAP, and "Visual Studio SDK" in any version.

When we expand the dependency tree, we get something like this:

```
├── OpenTAP 9.17.2
│   └── OpenTAP 9.17.2
└── Visual Studio SDK Any
    └── Visual Studio SDK 1.2.4
        └── SDK Any
            └── SDK 9.17.3 RC
                └── OpenTAP ^9.17.3 RC
                    └── OpenTAP 9.17.3 RC
```

We ask for OpenTAP in exactly version 9.17.2, and we get 9.17.2. So far, so good.

Visual Studio SDK "Any" resolves to version 1.2.4. Ok. 

This adds a dependency on "SDK" in version 'Any', which (at the time of writing) resolves to "9.17.3 RC". Ok.

This adds a dependency on "OpenTAP ^9.17.3 RC", which is incompatible with the dependency on exactly 9.17.2 from the root! Not ok!

If our dependency resolver was clever, it would detect this issue and select "SDK 9.17.2" which is able to satisfy our dependency on OpenTAP 9.17.2, but it doesn't.

# Solution

The dependency resolver should be able to resume dependency resolution from an earlier stage when it gets stuck. 

This can add a significant overhead especially in cases where there is no way to satisfy all dependencies. Every new attempt to resolve dependencies will require new requests, potentially to several http repositories, so we need to be clever about how we resolve versions.

As Dennis suggested, it would be helpful if the http repository implemented a mechanism for getting all packages that depend on a given package, but regardless, OpenTAP must be able to resolve the issue without explicit repository support because we still need to be able to support older versions of the repository.

Consider the following dependency tree:

```
├── OpenTAP 9.17.2
│   └── OpenTAP 9.17.2
├── Visual Studio SDK CE Any
│   └── Visual Studio SDK CE 1.2.4
│       └── SDK Any
│           └── SDK 9.17.3 RC
│               └── OpenTAP ^9.17.3 RC
│                   └── OpenTAP 9.17.3 RC
└── Developer's System CE Any
    ├── Developer's System CE 9.17.1
    │   └── Editor CE Any
    │       └── Editor CE 9.17.1
    │           └── OpenTAP ^9.17.2
    │               └── OpenTAP 9.17.2
    ├── OpenTAP Any
    │   └── OpenTAP 9.17.2
    └── Visual Studio SDK CE Any
        └── Visual Studio SDK CE 1.2.4
            └── SDK Any
                └── SDK 9.17.3 RC
                    └── OpenTAP ^9.17.3 RC
                        └── OpenTAP 9.17.3 RC

```

In this case, there are several identical subtrees that occur multiple times in the dependency tree. Visual Studio SDK in version 'any' is a root dependency, but it is also a dependency of 'Developer's System'

During both resolutions of different versions of 'Visual Studio SDK', we are attempting to resolve the same version of 'SDK' which turns out to be incompatible. 

A naive solution to resuming dependency resolution from an earlier stage can easily add exponential complexity to an already slow operation, so we should take great care to detect when a subtree is already either resolved, or known to be unresolvable.

rmadsen-ks:
Closes #460, #720

## New Features
- Being able to solve any possible dependency configuration (Often quite fast, possibly really slow)
- Added a new format to easily deploy an image: tap image install e.g: `OpenTAP,REST-API,TUI:beta,CSV`
- It is how harder to break an installation.. tap package uninstall now asks you to either cancel or also remove dependencies. --force can be used to override this behavior.
- install/uninstall with --force can now only be used a precision tool, when it is used, fancy resolving is removed only exactly what is asked for is done.
- Automatic repair of the package installation. Due to the fact that  package resolution can only find a globally valid solution, it will also repair a broken installation (if possible) by installing missing packages during "tap package install"

## Testing
Try running:
```sh
tap image install -t TESTFOLDER "OpenTAP,REST-API,TUI:beta,CSV"

tap image install -t TESTFOLDER2 "OpenTAP:^9.17,REST-API,TUI:beta,CSV"
```


rmadsen-ks:
After fighting with this for a while it now works perfectly!

Package.UnitTests/Image/Resolve.cs
```diff
@@ -296,7 +292,7 @@ public void ResolvePackages(string package1, string package1version, string pack
             catch (AggregateException ex)
             {
                 if (resultingVersion1 is null || resultingVersion1 == "error")
-                    Assert.AreEqual(1, ex.InnerExceptions.Count());
+                    ; //Assert.AreEqual(1, ex.InnerExceptions.Count());
```
Denrask, line 295: What's this? 🦩

Package.UnitTests/Image/Resolve.cs
```diff
@@ -324,20 +320,7 @@ public void ResolveDependencyVersionConflicts()
         [Test]
         public void TreeWithMissing()
```
Denrask, line 321: Remove the test

Package.UnitTests/Image/Resolve.cs
```diff
@@ -355,54 +338,13 @@ public void TrimOperatingSystemEntries()
         [Test]
         public void TestUnknownDependencies()
         {
```
Denrask, line 340: Remove

Package/Image/ImageResolver.cs
```diff
@@ -0,0 +1,280 @@
...
+        CancellationToken cancelToken;
+        public ImageResolver(CancellationToken cancelToken)
+        {
+            this.cancelToken = cancelToken;
+        }
+           
+
+        public long Iterations;
+
+        public ImageResolution ResolveImage(ImageSpecifier image, PackageDependencyGraph graph)
```
Denrask, line 24: Consider splitting this method up in multiple methods. 220 lines in one method is a bit long 😄

Package/PackageActions/ImageInstall.cs
```diff
@@ -26,12 +23,24 @@ internal class ImageInstallAction : IsolatedPackageAction
...
         public bool Merge { get; set; }
-
+        
         /// <summary>
         /// Never prompt for user input.
         /// </summary>
         [CommandLineArgument("non-interactive", Description = "Never prompt for user input.")]
         public bool NonInteractive { get; set; } = false;
+        
+        [CommandLineArgument("OS", Description = "Never prompt for user input.")]
```
Denrask, line 33: Fix description

Package/PackageActions/ImageInstall.cs
```diff
@@ -26,12 +23,24 @@ internal class ImageInstallAction : IsolatedPackageAction
...
+        public string Os { get; set; }
+
+        [CommandLineArgument("Architecture", Description = "Never prompt for user input.")]
```
Denrask, line 36: Fix description

Denrask (changes requested):
Included my initial review comments
GitHub also has some comments about xml comments and the obsoleted IgnoreDependencies
I'll begin testing.

Denrask:
It takes ~3 seconds to ask our internal repository for packages (estimate 30-40k packages)

alnlarsen:
Try this: 

`./tap image install -t ~/TESTFOLDER "OpenTAP:^9.17,PathWave License Manager"`

Then try to cancel downloading PWLM. The download can't be cancelled

I saw something similar when trying to cancel extracting files.

Package/Image/ImageSpecifier.cs
```diff
@@ -21,12 +23,77 @@ public class ImageSpecifier
         /// </summary>
         public List<PackageSpecifier> Packages { get; set; } = new List<PackageSpecifier>();
 
+        /// <summary> Installed specifiers. </summary>
+        internal PackageSpecifier[] FixedPackages { get; set; } = Array.Empty<PackageSpecifier>();
+        
+        /// <summary> The installed packages. </summary>
+        public ImmutableArray<PackageDef> InstalledPackages { get; set; } = ImmutableArray<PackageDef>.Empty;
```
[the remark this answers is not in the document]
rmadsen-ks:
```suggestion
        internal ImmutableArray<PackageDef> InstalledPackages { get; set; } = ImmutableArray<PackageDef>.Empty;
```

Package/Image/ImageSpecifier.cs
```diff
@@ -21,12 +23,77 @@ public class ImageSpecifier
...
+        {
+            var toInstall = new List<PackageSpecifier>();
+            var installed = installation.GetPackages().ToList();
+            foreach (var package in newPackages)
+            {
+                var ext = installed.FirstOrDefault(x => x.Name == package.Name);
+                if (ext != null)
+                {
+                    installed.Remove(ext);
+                    // warn about downgrade.
```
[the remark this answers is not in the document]
rmadsen-ks:
```suggestion
```

Package/Image/ImageResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+                    // any version -> take the newest first.
+                    versions.Reverse();
+                }else if(pkg.Version.MatchBehavior.HasFlag(VersionMatchBehavior.Exact))
+                {
+                    //exact may be more than one version, even though the match behavior is 'exact'.
+                    // for example OpenTAP '9.17' is exact, but many versions matches that.
+                    // We are interested in the newest in this case, so order newest -> oldest within the span.
+                    versions.Sort();
+                    versions.Reverse();
+                }
```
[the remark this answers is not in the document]
rmadsen-ks: I agree, but I think it is good to have a place to put the comments.

alnlarsen:
Check the package diff output:

```
 Public Methods:
+Void OpenTap.Package.ImageSpecifier::.ctor(Collections.Generic.List`1<OpenTap.Package.PackageSpecifier>, String)

Public Properties:
+Collections.Immutable.ImmutableArray`1<OpenTap.Package.PackageDef> OpenTap.Package.ImageSpecifier::InstalledPackages { public get; public set; }
+String OpenTap.Package.ImageSpecifier::OS { public get; public set; }
+OpenTap.Package.CpuArchitecture OpenTap.Package.ImageSpecifier::Architecture { public get; public set; }
+String OpenTap.Package.PackageSpecifier::ToString()

Public Fields:
+OpenTap.Package.VersionSpecifier OpenTap.Package.VersionSpecifier::AnyRelease
+OpenTap.Package.PackageDef[] OpenTap.Package.FilePackageRepository::GetAllPackages(Threading.CancellationToken)
+String OpenTap.Package.FilePackageRepository::ToString()
+String OpenTap.Package.HttpPackageRepository::ToString()
```

Do all of these need to be public?

Package/Image/ImageSpecifier.cs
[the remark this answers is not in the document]
rmadsen-ks: It is actually not necessary here.

Package/Image/ImageResolver.cs
```diff
@@ -0,0 +1,275 @@
...
+                        {
+                            var pkg2 = packages[j];
+                            if (pkg2.Name == dep.Name)
+                            {
+                                // this dependency can be satisfied?
+                                if (!pkg2.Version.IsSatisfiedBy(dep.Version) &&
+                                    !dep.Version.IsSatisfiedBy(pkg2.Version))
+                                {
+                                    // this dependency is unresolvable
+                                    return new FailedImageResolution(Array.Empty<PackageSpecifier>(), Iterations);
```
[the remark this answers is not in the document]
rmadsen-ks: Good idea.

Package/Image/ImageResolver.cs
rmadsen-ks: I actually don't think it would help much. The order of all these things are quite important.

alnlarsen:
This looks good!

alnlarsen (dismissed):
I think this looks good.

Package/Image/ImageSpecifier.cs
```diff
@@ -21,12 +23,74 @@ public class ImageSpecifier
...
+
+        
         /// <summary>
         /// OpenTAP repositories to fetch the desired packages from
         /// These should be well formed URIs and will be interpreted relative to the BaseAddress set in AuthenticationSettings.
+        /// May be null if unspecified.
         /// </summary>
-        public List<string> Repositories { get; set; } =
-            new List<string>() { new Uri(PackageCacheHelper.PackageCacheDirectory).AbsoluteUri };
+        public List<string> Repositories { get; set; }
```
Denrask, line 85: You are still not happy with initializing this, @rmadsen-ks ? :)

AsgerIversen:
@alnlarsen mentioned an issue with cancelling downloads. @rmadsen-ks will you look at that before we merge?

alnlarsen:
Thanks, @rmadsen-ks, cancelling image installs works now!

AsgerIversen (changes requested):
Some initial small things from testing a bit:

Package.UnitTests/Image/DependencyGraphTest.cs
```diff
@@ -0,0 +1,181 @@
...
+        [TestCase("OpenTAP", "9.10.0", "ExactDependency", "1.0.0", "error", "error")]
+        [TestCase("OpenTAP", "9.13.1", "ExactDependency", "1.0.0", "9.13.1", "1.0.0")]
+        [TestCase("OpenTAP", "9.13", "ExactDependency", "1.0.0", "9.13.1", "1.0.0")]
+        [TestCase("Cyclic", "1.0.0", "Cyclic2", "1.0.0", "1.0.0", "1.0.0")]
+        [TestCase("OpenTAP", "9.13", "ExactDependency", "^1.0.0", "9.13.1", "1.0.0")]
+        [TestCase("OpenTAP", "^9.10.0", "Demonstration", "^9.0.6", "9.12.0", "9.1.0")]
+        public void ResolveDependencies(string package1, string package1version, string package2, string package2version, string resultingVersion1, string resultingVersion2)
+        {
+        }
+    }
```
AsgerIversen, line 23: This doesn't seem to do much. Should we delete or add some content to these test cases?

Package.UnitTests/Tap.Package.UnitTests.csproj
```diff
@@ -24,6 +24,7 @@
   </ItemGroup>
   
   <ItemGroup>
+    <EmbeddedResource Include="Image\opentap-packages.json.gz" />
```
AsgerIversen, line 27: Is this file used anywhere?

Package/Image/ImageSpecifier.cs
```diff
@@ -36,16 +99,34 @@ public class ImageSpecifier
...
+            if (image.Success == false)
+            {
+                var unsatisfiedDependencies = InstalledPackages.Where(x => false == x.Dependencies.All(dep =>
+                    InstalledPackages.Any(x2 =>
+                        x2.Name == dep.Name && dep.Version.IsSatisfiedBy(x2.Version.AsExactSpecifier())))).ToArray();
+                if (unsatisfiedDependencies.Any())
+                {
+                    throw new ImageResolveException(image,
+                        string.Format("This is probably due to the current following package dependencies being broken: {0}",
+                        string.Join(", ", unsatisfiedDependencies.Select(x => x.Name))));
```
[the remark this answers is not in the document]
rmadsen-ks: I agree the error message here is not the best.

Your suggestion for a very specific error message e.g" X depends on Y so this cannot work" in a consistent manner is very hard to implement. Mostly because the reality is not that clear cut - there is always at least two ways to state a dependency problem. See for example this:
```
./tap image install  -t ../../deply "OpenTAP:9.16,TUI:0.1.0-beta.161```
Unable to resolve packages: TUI: 0.1.0-beta.161
```
In this case we are unable to resolve TUI 0.1.0-beta.161. This is because that version depends on OpenTAP 9.18, but we explicitly specify a 9.16 version.

But depending on the order, you could also say that it is `OpenTAP:9.16` that we could not resolve.

What we are printing now is the missing dependencies in the last failed attempt to resolve the problem, but there could be many other places to fix the issue (e.g Updating OpenTAP instead of downgrading TUI).


Anyway, I have changed it so that it says "Unable to resolve packages" is that better?

Package.UnitTests/Tap.Package.UnitTests.csproj
rmadsen-ks: It is used in tests.

Denrask:
`ImageSpecifier.MergeAndResolve` throws an exception:

image

I'll continue to investigate why.

Denrask:
Can we create a better error description then this?

"Unable to resolve packages: Runner: beta" -> There are plenty of Runner betas, but none is compatible with the specified OpenTAP version.

image

AsgerIversen (approved):

PR #725, opened 2022-07-26
Fixed small bug in GetClient API. Added unittests to show intended behavior

### LINKED ISSUE #724: Small logic error in the AuthenticationSettings.Current.GetClient API.
If `AuthenticationSettings.Current.BaseAddress` is set to "http://localhost" and I request a client with baseaddress "api/results", I expect the resulting address to be "http://localhost/api/results" as this is the intended design. However, I'm now getting an exception. We got a minor if check wrong.

Denrask:
Closes #724


rmadsen-ks (approved):
Looks good!

PR #735, opened 2022-08-04
734 remove --quiet shortname

### LINKED ISSUE #734: CLI option '--quiet' works, but '-q' does not
image
image

alnlarsen:
This removes the '-q' shortname option for 'quiet'. I prefer this to adding the '-q' shortname since it is likely to lead to collisions that are hard to avoid since '-q' is specially handled.

Closes #734

### DIFF
--- a/Cli/TapEntry.cs
+++ b/Cli/TapEntry.cs
@@ -173,7 +173,7 @@ namespace OpenTap.Cli
             {
                 var args = Environment.GetCommandLineArgs();
                 bool isVerbose = args.Contains("--verbose") || args.Contains("-v");
-                bool isQuiet = args.Contains("--quiet") || args.Contains("-q");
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks good!

AsgerIversen:
@alnlarsen can you elaborate on how `-q` could lead to collisions?

AsgerIversen:
How about https://github.com/opentap/opentap/pull/750

alnlarsen:
@AsgerIversen I was thinking if someone else's CLI action used the `-q` shortname somewhere, using that option would suddenly turn the output quiet if we hardcode the `-q` flag like you suggested. But I don't think that's actually possible because the ArgumentParser has always known about our `-q` shortname, and it just wasn't being honored.

Personally I don't really care which of the two solutions we go with. Your PR looks fine too.

PR #738, opened 2022-08-05
737 Tap Run NonInteractive Load Error Handling

### LINKED ISSUE #737: tap run --non-interactive runs a test plan even if it fails to load it
If something goes wrong loading a test plan, it might change the flow of the plan execution but we still run the test plan without any error. We should fail instead.

One place were I want to use this is running a test plan related to the python plugin in order to test that things works. If none of my steps can load, i might still be able to run a broken plan, resulting in a pass, but it should have been an error.

```
bruger@Brugers-Air Debug % ./tap run test2.TapPlan --non-interactive
OpenTAP Command Line Interface 9.4.0-Development
XML line 4 column 6: Unable to locate type 'OpenTap.Plugins.BasicSteps.LogStep2'. Are you missing a plugin?
XML line 4 column 6: Unable to deserialize step.
XML Line 4: Unable to locate type 'OpenTap.Plugins.BasicSteps.LogStep2'. Are you missing a plugin?
Loaded test plan from test2.TapPlan [16.3 ms]
Test Plan: test2
-----------------------------------------------------------------
Starting TestPlan 'test2' on 08/05/2022 16:45:13, 0 of 0 TestSteps enabled.
---- Summary of test plan started 08/05/2022 16:45:13 ----
----------------------------------------------------------
------ Test plan completed successfully in 17.0 ms -------
```

rmadsen-ks:
Closes #737 

This might be slightly breaking in cases somebody were relying on the behavior to ignore load errors, but in general I think it is better to get proper error messages and errors than to get possibly very broken behavior.

### DIFF
--- a/Engine/Cli/RunCliAction.cs
+++ b/Engine/Cli/RunCliAction.cs
@@ -188,7 +188,9 @@ namespace OpenTap.Cli
             }
             catch (TestPlan.PlanLoadException ex)
             {
-                log.Error(ex.Message);
+                // at this point the log messages are already written out saying what went wrong.
+                log.Error("Unable to load test plan.");
+                log.Debug(ex);
                 return (int)ExitStatus.LoadError;
             }
             catch (OperationCanceledException) when (TapThread.Current.AbortToken.IsCancellationRequested)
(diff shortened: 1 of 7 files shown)


alnlarsen:
This seems to work as intended. Let's do it!

alnlarsen (approved):

PR #740, opened 2022-08-09
739 don't include Package.Dependencies element when serializing `PackageVersion[]`

### LINKED ISSUE #739: <Package.Dependencies> element is included when serializing `PackageVersion[]`
`<Package.Dependencies>` is a concept which only makes sense for test plans, but we include it in other instances, such as when serializing `PackageVersion[]`. This is annoying because it will add a dependency on the currently used version of OpenTAP in order to deserialize a data object, which will cause serializer errors when trying to deserialize that same string in an older version of OpenTAP.

Let's fix this :)

alnlarsen:
Update the PackageVersion serializer to disable writing `Package.Dependencies` when serializing either a single `PackageVersion`, or a single collection of `PackageVersion`s.

Closes #739

### DIFF
--- a/Package/PackageDefinitionSerializerPlugin.cs
+++ b/Package/PackageDefinitionSerializerPlugin.cs
@@ -97,7 +97,9 @@ namespace OpenTap.Package
         }
 
         /// <summary>
-        /// Called as part for the serialization chain. Returns false if it cannot serialize the XML element.  
+        /// Called as part for the serialization chain. Returns false if it cannot serialize the XML element.
+        /// Disables the 'WritePackageDependencies' feature of the <see cref="TestPlanPackageDependencySerializer"/>
+        /// further up the chain when it runs.
         /// </summary>
         public override bool Serialize(XElement node, object obj, ITypeData expectedType)
         {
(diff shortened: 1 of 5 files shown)


rmadsen-ks (changes requested):
``var writePackageDependencies = Serializer.GetUsedTypes().Any(t => t.DescendsTo(typeof(TestPlan)));```

We don't want to only do this when the type is TestPlan.

If you dont want the behavior in some specific case, why not just add the option to disable the package dependency check?

Package/PackageDefinitionSerializerPlugin.cs
```diff
@@ -101,7 +101,9 @@ public override bool Deserialize(XElement node, ITypeData t, Action<object> sett
         /// </summary>
         public override bool Serialize(XElement node, object obj, ITypeData expectedType)
         {
-            if (expectedType.IsA(typeof(PackageDef)) == false && expectedType.IsA(typeof(PackageIdentifier)) == false)
+            if (expectedType.IsA(typeof(PackageDef)) == false
```
rmadsen-ks, line 104: can you modify the xml comment of this method to be more specific?

rmadsen-ks (changes requested):
OK, I understand this now. Looks good, but I had a couple of suggestions.

alnlarsen:
I have added some comments and improved the license element bug fix. Please take another look :)

rmadsen-ks (approved):
Looks good!

PR #744, opened 2022-08-09
Always set the PackageSource when resolving a package

Denrask:
We know where we just resolved the package from. The repo does not necessarily know it's correct location, if e.g. behind a proxy.


AsgerIversen (approved):
Looks good to me.

PR #747, opened 2022-08-10
746 improve package extraction error handling

### LINKED ISSUE #746: Error messages are unhelpful when installer fails to extract files
Consider this scenario where I don't have enough disk space to install a package:

This is the verbose log:
image

In many cases, a retry is completely useless, and in the case of missing space, it would be easy to add a prompt to free up space. If we are running non-interactively in this scenario, retrying is a complete waste of time.

In any case, we should log the exception instead of silently ignoring it.

alnlarsen:
Add retry dialog when out of space.
Also improve general error messages during package extraction

Closes #746

### DIFF
--- a/Package/Installer.cs
+++ b/Package/Installer.cs
@@ -274,7 +274,7 @@ namespace OpenTap.Package
 
                 while (isPackageFilesInUse(tapDir, packagePaths, exclude))
                 {
-                    var req = new AbortOrRetryRequest("Package files are in use", inUseString) {Response = AbortOrRetryResponse.Abort};
+                    var req = new AbortOrRetryRequest("Package Files Are In Use", inUseString) {Response = AbortOrRetryResponse.Abort};
                     UserInput.Request(req, waitForFilesTimeout, true);
 
                     if (req.Response == AbortOrRetryResponse.Abort)
--- a/Package/Installer.cs
+++ b/Package/Installer.cs
@@ -405,7 +405,6 @@ namespace OpenTap.Package
     }
 
     [Obfuscation(Exclude = true)]
-    [Display("Package files are in use")]
     class AbortOrRetryRequest
     {
         public AbortOrRetryRequest(string message)
--- a/Package/Installer.cs
+++ b/Package/Installer.cs
@@ -274,7 +274,7 @@ namespace OpenTap.Package
 
                 while (isPackageFilesInUse(tapDir, packagePaths, exclude))
                 {
-                    var req = new AbortOrRetryRequest(inUseString) {Response = AbortOrRetryResponse.Abort};
+                    var req = new AbortOrRetryRequest("Package files are in use", inUseString) {Response = AbortOrRetryResponse.Abort};
                     UserInput.Request(req, waitForFilesTimeout, true);
 
                     if (req.Response == AbortOrRetryResponse.Abort)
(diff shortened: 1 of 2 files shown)


Package/Installer.cs
```diff
@@ -405,7 +405,6 @@ enum AbortOrRetryResponse
     }
 
     [Obfuscation(Exclude = true)]
-    [Display("Package files are in use")]
     class AbortOrRetryRequest
```
rmadsen-ks, line 408: Now you removed this, but does that not affect the original use of the code?

rmadsen-ks (changes requested):
Really nice improvement. I had one comment though.

Package/Installer.cs
alnlarsen: You're right, I forgot to add a generic title :)

Package/Installer.cs
```diff
@@ -274,7 +274,7 @@ private void WaitForPackageFilesFree(string tapDir, List<string> packagePaths)
 
                 while (isPackageFilesInUse(tapDir, packagePaths, exclude))
                 {
-                    var req = new AbortOrRetryRequest(inUseString) {Response = AbortOrRetryResponse.Abort};
+                    var req = new AbortOrRetryRequest("Package files are in use", inUseString) {Response = AbortOrRetryResponse.Abort};
```
rmadsen-ks, line 277:
```suggestion
                    var req = new AbortOrRetryRequest("Package files are in use.", inUseString) {Response = AbortOrRetryResponse.Abort};
```

I guess we should use either title case or proper punctuation here.

Package/PluginInstaller.cs
```diff
@@ -455,7 +455,16 @@ internal static List<string> UnpackPackage(string packagePath, string destinatio
                             {
                                 throw;
                             }
-                            catch
+                            catch (IOException ex) when (ex.Message.Contains("There is not enough space on the disk"))
+                            {
+                                log.Error(ex.Message);
+                                var req = new AbortOrRetryRequest("Not enough disk space", $"File '{part.FullName}' requires {Utils.BytesToReadable(part.Length)} of free space. " +
```
rmadsen-ks, line 461:
```suggestion
                                var req = new AbortOrRetryRequest("Not Enough Disk Space", $"File '{part.FullName}' requires {Utils.BytesToReadable(part.Length)} of free space. " +
```

Here we can use title case.

rmadsen-ks (changes requested):
What do you think about these minor changes?

alnlarsen:
Let's use title casing in both dialogs then

rmadsen-ks (dismissed):
Looks great!

AsgerIversen:
Just thinking about how this will behave when the REST-API session manager is installing stuff for a new session... 

If on one is subscribed to the UserInput SignalR thing, I think the response will now just be "User aborted installation." when the real reason is "There is not enough space on the disk"

@Denrask maybe you can comment?

alnlarsen:
That's a good point. How does REST handle the dialog when package files are in use? @Denrask @AsgerIversen 

In any case, I think the exception should mention that the installation was aborted due to missing disk space.

Denrask:
I've been sylting this for a while. I'm not too concerned with this. I'll just handle it with default behavior.

Denrask (approved):

PR #748, opened 2022-08-10
Changed the FilePackageRepository constructor to allow uri, relative …

Denrask:
…and absolute paths.


AsgerIversen (approved):
Looks good

PR #750, opened 2022-08-10
Also ask '--quiet' if  '-q' is specified (as the --help says we do)

AsgerIversen:


alnlarsen (approved):
Looks great. I leave up to you which of these two candidates we merge.

PR #756, opened 2022-08-12
Check whether a property is browsable when determining if it should be visible

### LINKED ISSUE #755: CLI options with [Browsable(false)] are visible

 - OS: Windows 10
 - OpenTAP Version 9.18.4

**To Reproduce**

1. Run `tap package list -h` - > the option `--no-isolation` is visible

2. Run `tap package check-updates -h` -> the option '--startup' is visible

**Actual behavior**

These options have [Browsable(false)] attribute and are visible

**Expected behavior**

CLI options with [Browsable(false)] attribute should be hidden

image

image

### LINKED ISSUE #764: Obsolete CommandLineArgumentAttribute.Visible in favor of BrowsableAttribute
Related: https://github.com/opentap/opentap/issues/755

It is a bit confusing that we want two different ways of supporting setting the visibility of a command line argument.

Since both ways are equally good, we should support the one which is more in line with all of our other APIs.

alnlarsen:
Check whether a property is browsable when determining if it should be visible
Cleanup formatting in unrelated unittest.

Also obsoletes `CommandLineArgumentAttribute.Visible`

Closes #755
Closes #764

### DIFF
--- a/Shared/ICliActionExtensions.cs
+++ b/Shared/ICliActionExtensions.cs
@@ -6,6 +6,7 @@
 using OpenTap.Package;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


Shared/ICliActionExtensions.cs
```diff
@@ -92,7 +93,7 @@ public static int PerformExecute(this ICliAction action, string[] parameters)
                 }
 
                 var arg = ap.AllOptions.Add(attr.Name, attr.ShortName?.FirstOrDefault() ?? '\0', needsArg, description);
-                arg.IsVisible = attr.Visible;
+                arg.IsVisible = attr.Visible && (prop.GetAttribute<BrowsableAttribute>()?.Browsable ?? true);
```
rmadsen-ks, line 96: Maybe we should obsolete the Visible property?

rmadsen-ks (dismissed):
Looks good

rmadsen-ks (approved):
Looks great!

PR #757, opened 2022-08-12
ignore system-wide packages for dependency issues

### LINKED ISSUE #754: Ignore dependencies from system-wide packages
I tried creating an OpenTAP 9.10.0 installation from the CLI: 
`~/tap/tap package install OpenTAP --version 9.10.0 -t .\testdir\`

I expected this to work amazingly, but then reality struck:

```
Package 'PathWave License Manager' depends on 'OpenTAP' version '^9.14.0', but '9.10.0+53e49232' is requested.
Package 'VCRedist' depends on 'OpenTAP' version '^9.13.0+a5998b9e', but '9.10.0+53e49232' is requested.
To install packages despite the conflicts, use the --no-dependencies option.
```

The packages in my system-wide package directory would make any installation of OpenTAP older than 9.13.0 'broken'.

I think we should never consider dependencies from system-wide packages. Otherwise it is trivially to get 'broken' installations that are not really broken.

alnlarsen:
This solves a problem where system-wide packages can make any installation suddenly broken.

Closes #754


Package/PackageActions/Install.cs
```diff
@@ -190,7 +190,7 @@ private int DoExecute(CancellationToken cancellationToken)
 
                 RaiseProgressUpdate(10, "Gathering dependencies.");
                 bool checkDependencies = (!IgnoreDependencies && !Force) || CheckOnly;
-                var issue = DependencyChecker.CheckDependencies(installationPackages, packagesToInstall,
+                var issue = DependencyChecker.CheckDependencies(installationPackages.Where(p => p.IsSystemWide() == false), packagesToInstall,
```
rmadsen-ks, line 193: What happens if an installed package requires a system-wide package to be installed? Does those dependencies then fail?

Also, it'd be good with a comment about why we are removing system wide packages from the list of dependencies.

rmadsen-ks (changes requested):

Package/PackageActions/Install.cs
alnlarsen: I understand your point. Maybe we need to discuss the optimal behavior here.

AsgerIversen: After testing i can confirm that you still get a system-wide package (PathWave License Manager) installed as a dependency e.g. with `tap package install "Switch Manager"`

AsgerIversen (approved):

rmadsen-ks (approved):
Looks good!

PR #759, opened 2022-08-12
Filter installed packages by OS if an OS was specified.

### LINKED ISSUE #732: "tap package list --os" should show an error with an invalid argument
OS: macOS x64 with version 12.5
OpenTAP version 9.18.4+7dec4717

**To Reproduce**

1. Access the Terminal
2. Go to .tap folder
3. Use command ./tap package list --os test
4. You should get a list with installed apps on the local machine
5. Use command ./tap package list --os 443fdfs
6. You should get a list with installed apps on the local machine
7. Use command ./tap package list -i
8. You should get the same results as above
9. A clear and concise description of what the bug is.

**Actual behavior**

You get the same results from a wrong command when using **--os** in front and any random numbers or letters after

MicrosoftTeams-image (1)

**Expected behavior**

Since the argument for **--os** is invalid, running this command should show an error

alnlarsen:
Closes #732

### DIFF
--- a/Package/PackageActions/List.cs
+++ b/Package/PackageActions/List.cs
@@ -45,6 +45,7 @@ namespace OpenTap.Package
 
         protected override int LockedExecute(CancellationToken cancellationToken)
         {
+            // OS was explicitly specified. This is interpreted as: Show only packages compatible with that OS. 
             bool checkOs = OS != null;
             if (OS == null)
             {
--- a/Package/PackageActions/List.cs
+++ b/Package/PackageActions/List.cs
@@ -45,6 +45,7 @@ namespace OpenTap.Package
 
         protected override int LockedExecute(CancellationToken cancellationToken)
         {
+            bool checkOs = OS != null;
             if (OS == null)
             {
                 switch (Environment.OSVersion.Platform)
@@ -73,6 +74,8 @@ namespace OpenTap.Package
                 Target = FileSystemHelper.GetCurrentInstallationDirectory();
 
             HashSet<PackageDef> installed = new Installation(Target).GetPackages().ToHashSet();
+            if (checkOs)
+                installed = installed.Where(p => p.OS.Split(',').Any(o => o.Equals(OS, StringComparison.OrdinalIgnoreCase))).ToHashSet();
 
 
             VersionSpecifier versionSpec = VersionSpecifier.Parse("^");
--- a/Package/PackageActions/List.cs
+++ b/Package/PackageActions/List.cs
@@ -72,11 +72,10 @@ namespace OpenTap.Package
 
             if (Target == null)
                 Target = FileSystemHelper.GetCurrentInstallationDirectory();
-
+            
             HashSet<PackageDef> installed = new Installation(Target).GetPackages().ToHashSet();
             if (checkOs)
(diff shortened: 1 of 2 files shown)


Package/PackageActions/List.cs
```diff
@@ -73,6 +74,8 @@ protected override int LockedExecute(CancellationToken cancellationToken)
                 Target = FileSystemHelper.GetCurrentInstallationDirectory();
 
             HashSet<PackageDef> installed = new Installation(Target).GetPackages().ToHashSet();
+            if (checkOs)
+                installed = installed.Where(p => p.OS.Split(',').Any(o => o.Equals(OS, StringComparison.OrdinalIgnoreCase))).ToHashSet();
```
rmadsen-ks, line 78: This looks good, but I think we should try to make the code a bit less implicit

by doing something like
```cs
static bool SupportsOS(this PackageDef pkg, string os) => pkg.OS.Split(',').Any(o => o.Equals(os, StringComparison.OrdinalIgnoreCase));

installed = installed.Where(pkg =>pkg.SupportsOS(OS)).ToHashSet();
```
```

**Review by rmadsen-ks — CHANGES_REQUESTED** [review]

File: Package/PackageActions/List.cs (hunk 1 above)
Comment (alnlarsen) [reply Package/PackageActions/List.cs:78]: I agree.

**Comment by alnlarsen** [comment]
@rmadsen-ks we already had an `IsOsCompatible` method elsewhere that we can reuse. Please take another look.

File: Package/PackageActions/List.cs [hunk 2]
```diff
@@ -45,6 +45,7 @@ public PackageListAction()
 
         protected override int LockedExecute(CancellationToken cancellationToken)
         {
+            bool checkOs = OS != null;
```
rmadsen-ks, line 48:
```suggestion
            // OS was explicitly specified. This is interpreted as: Show only packages compatible with that OS. 
            bool checkOs = OS != null;
```

Package/PackageActions/List.cs
```diff
@@ -71,9 +72,10 @@ protected override int LockedExecute(CancellationToken cancellationToken)
 
             if (Target == null)
                 Target = FileSystemHelper.GetCurrentInstallationDirectory();
-
+            
             HashSet<PackageDef> installed = new Installation(Target).GetPackages().ToHashSet();
-
+            if (checkOs)
+                installed = installed.Where(pkg => pkg.IsOsCompatible(OS)).ToHashSet();
```
rmadsen-ks, line 78: Why does the checkOs only apply to installed packages and not the packages found in repositories?

rmadsen-ks (changes requested):
Added a comment, I am not sure how this works.

Package/PackageActions/List.cs
alnlarsen: Because OS was always applied to packages found in repositories:
`packages.AddRange(PackageRepositoryHelpers.GetPackageNameAndVersionFromAllRepos(repositories, new PackageSpecifier("", versionSpec, Architecture, OS)));`

This fix just applies it to installed packages

rmadsen-ks (approved):
Looks good!

Issue #765, opened 2022-08-23
All test step threads are blocked while trying to publish results


anthrax-0:
**Describe the bug**
It seems that in version 9.17.0 new [optimization](https://github.com/opentap/opentap/commit/048c20a19f0717adb9abe87e1bd8f7f22dc58344#diff-0f52992e9cbff4905adc7e2dbfef0b1bc85a8b9fe173f046d834316e4d23a65aR588) was introduced in the ResultProxy class. This code is calling [Dequeue](https://github.com/opentap/opentap/commit/048c20a19f0717adb9abe87e1bd8f7f22dc58344#diff-c9264f59ffd3b7c1a04ec62456b0842135077223e479088a04db767e26e359a3R193) method of the WorkQueue introduced in the same commit. But this code does not interact with the addSemaphore semaphore which means that now the queue is sometimes getting cleared by the worker [thread](https://github.com/opentap/opentap/commit/048c20a19f0717adb9abe87e1bd8f7f22dc58344#diff-c9264f59ffd3b7c1a04ec62456b0842135077223e479088a04db767e26e359a3L116)  which calls Wait on semaphore and sometimes it is cleared by Dequeue method which does not. As the worker thread waits for the queue to be not empty, it can not decrease semaphore count on its own. So semaphore count is increasing and after some time the thread trying to publish results is stuck in the [loop](https://github.com/opentap/opentap/commit/048c20a19f0717adb9abe87e1bd8f7f22dc58344#diff-c9264f59ffd3b7c1a04ec62456b0842135077223e479088a04db767e26e359a3L176)  without possibility to exit unless there is some time guard step controlling it.
When changing in the debugger the currentCount of the semaphore to 0, the test continues, and logs are produced once more.

As it is the bug is a full **blocker** preventing any use of OpenTAP with versions higher than 9.16.4

**Environment (please complete the following information):**

 - OS: Windows 10, Linux
 - OpenTAP Version: 9.17.0+ 

**To Reproduce**

1. Create a result listener

2. Create a test plan that will publish some results and run for some time to detect the problem.

3. To not wait for hours to see the deadlock, one can connect with the debugger and check the state of the semaphore `addSemaphore` in WorkQueue class. As this semaphore is supposed to prevent the overloading of the queue it should not have more threads available than queue count, if it has more and the number is increasing - at some point threads will  get stuck.

**Actual behavior**

[...]

PR #767, opened 2022-08-25
765 ManyResultsThreadBlock (Hotfix/patch)

AsgerIversen:
Cherry pick fix for #765 to a 9.18.5 patch release.


rmadsen-ks (approved):
Looks good, but we need to bump .gitversion as well.

AsgerIversen:
> Looks good, but we need to bump .gitversion as well.

We already did that on the release9x branch

PR #768, opened 2022-08-25
Tried to fix the race condition related to type data searchers, by us…

rmadsen-ks:
…ing an immutable list instead of a list accessible from multiple threads.

### DIFF
--- a/Engine/TypeData.cs
+++ b/Engine/TypeData.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


db-ks:
Thank you @rmadsen-ks for the fix; I have tested on one of our setups for 2 days in a track. I have not experienced the issue since I updated to the alpha version you provided.

rmadsen-ks:
Thanks for verifying @db-ks.

@Denrask, do you approve this PR?

Denrask (approved):

PR #770, opened 2022-08-26
Fixed a bug occuring because the WorkQueue did not decrement (Hotfix)

AsgerIversen:
…e and pop work items in the right order when doing Dequeue.


rmadsen-ks (approved):
Lets do it!

PR #775, opened 2022-09-01
774 Available Values Overwritten during parameterization: Fix + test

### LINKED ISSUE #774: AvailableValues are overridden when parameterizing parameterizing properties
Consider the following example steps:

```csharp
public class S1 : TestStep
{
    [Browsable(false)]
    public string[] Available { get; set; } = new[] { "On", "Off" };
    [AvailableValues(nameof(Available))]
    public string MyValue { get; set; }
    public override void Run()
    {
        throw new NotImplementedException();
    }
}
public class S2 : TestStep
{
    [Browsable(false)]
    public string[] Available { get; set; } = new[] { "Blue", "Green", "Red" };
    [AvailableValues(nameof(Available))]
    public string MyValue { get; set; }
    public override void Run()
    {
        throw new NotImplementedException();
    }
}
```

If `MyValue` from both of these steps is parameterized on a sweep step, and a value is selected in the drop down, then the `Available` property on the second instance is assigned to the `Available` property from the first instance. E.g. the `S2` instance will get an `Available` property with the values `{ "On", "Off" }`.

I could reproduce this issue with the 9.19 beta, but not with OpenTAP 9.17.4.

rmadsen-ks:
Close #774

Fixed the issue by changing the way we control list equality.

The annotation system thinks the lists has been changed because the returned value from merged value annotation was not null initially and the value becomes aligned unintentionally afterwards.

This is really a corner case for Browsable(False) properties which are being used for AvailableValues, but for now this should fix the issue.

A unit test has been added to ensure the issue no appears in the future.

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -1978,7 +1978,6 @@ namespace OpenTap
                                         
                                     lst2 = Array.CreateInstance(typedata.ElementType.Type, nElements);
                                     lst = lst2;
-                                    objValue.Value = lst2;
 
                                 }
                                 else
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -737,8 +737,21 @@ namespace OpenTap
                     {
                         var x = merged[i];
                         var thisVal = x.Get<IObjectValueAnnotation>().Value;
-                        if (selectedValue is IEnumerable && !(selectedValue is string))
-                            return selectedValue;
+                        if (thisVal == selectedValue) return selectedValue;
+                        if (selectedValue is IEnumerable ie1 && !(selectedValue is string))
+                        {
+                            // if the two lists has the same content it is fine to just return one of them.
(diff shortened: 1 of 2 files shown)


Engine/Annotations/Annotation.cs
```diff
@@ -1962,8 +1977,9 @@ public void Write(object source)
                                         throw new Exception($"Cannot add elements to collection because it is not writable.");
                                         
                                     lst2 = Array.CreateInstance(typedata.ElementType.Type, nElements);
+                                    lst = lst2;
                                     objValue.Value = lst2;
```
alnlarsen, line 1981: This assignment seems redundant now

alnlarsen, line 1981:
```suggestion
```

alnlarsen (approved):
Awesome!

khang-yong:
This fix works. It fixes Editor X issue [#461](http://gitlab.it.keysight.com/tap/editors/editor-x/issues/461)

Issue #607, opened 2022-05-25
Optimization Opportunity: PluginSearcher should not read already loaded assemblies


rmadsen-ks:
PluginSearcher is quite fast when analyzing plugins, but for DLLs which are already loaded, it should probably use plain-old reflection.

By default, these assemblies are loaded:
```
00:00:00.070 : Resolver      : Debug       : Loaded assembly mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 from C:\Windows\Microsoft.NET\Framework64\v4.0.30319\mscorlib.dll
00:00:00.075 : Resolver      : Debug       : Loaded assembly tap, Version=8.0.0.0, Culture=neutral, PublicKeyToken=null from C:\Keysight\Development\opentap\bin\Debug\tap.exe
00:00:00.075 : Resolver      : Debug       : Loaded assembly OpenTap.Cli, Version=9.4.0.0, Culture=neutral, PublicKeyToken=null from
00:00:00.075 : Resolver      : Debug       : Loaded assembly netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 from C:\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\netstandard\v4.0_2.0.0.0__cc7b13ffcd2ddd51\netstandard.dll
00:00:00.076 : Resolver      : Debug       : Loaded assembly System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 from C:\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System.Core\v4.0_4.0.0.0__b77a5c561934e089\System.Core.dll
00:00:00.076 : Resolver      : Debug       : Loaded assembly System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 from C:\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System\v4.0_4.0.0.0__b77a5c561934e089\System.dll
00:00:00.076 : Resolver      : Debug       : Loaded assembly OpenTap, Version=9.4.0.0, Culture=neutral, PublicKeyToken=null from C:\Keysight\Development\opentap\bin\Debug\OpenTap.dll
[...]

PR #779, opened 2022-09-02
Make valuetypes constructable

### LINKED ISSUE #610: TypeData.CanCreateInstance has invalid value for ValueTypes
Every ValueType can be instanced using the default constructor, but our plugin searcher does not correctly determine if an instance can be created.

This can give issues if the user wishes to create a `List<KeyValuePair<string,string>>` because KeyValuePair is a value type.

This mostly occurs because of the order in which we discover type references. In some cases the opentap dll is loaded before the mscorlib dll and in this case the ValueType base class cannot be determined before later. 

The only way to determine if something is a ValueType is by looking at the baseclass.

alnlarsen:
Mark valuetypes as constructable in the plugin searcher
Make typedata treat valuetypes as constructable
Add unittest to verify the fix
CLoses #610

### DIFF
--- a/Engine/PluginSearcher.cs
+++ b/Engine/PluginSearcher.cs
@@ -507,35 +507,63 @@ namespace OpenTap
                 return tp;
             return null; // This is not a type that we care about (not defined in any of the files the searcher is given)
         }
-        
-        private bool IsValueType(TypeDefinition typeDef)
-        {
-            try
-            {
-                var name = CurrentReader.GetString(typeDef.Name);
-                if (name == nameof(ValueType)) return true;
 
-                var baseType = typeDef.BaseType;
-                switch (baseType.Kind)
+        private static readonly Dictionary<string, bool> ValueTypeMap = new Dictionary<string, bool>(); 
+        private bool IsValueType(TypeDefinition typeDef, string typeName = null)
+        {
+            bool helper(string s)
+            {
+                try
                 {
-                    case HandleKind.TypeReference:
-                        var tr = (TypeReferenceHandle)baseType;
-                        var r = CurrentReader.GetTypeReference(tr);
-                        return CurrentReader.GetString(r.Name) == nameof(ValueType);
-                    case HandleKind.TypeDefinition:
-                        var td = (TypeDefinitionHandle)baseType;
-                        var d = CurrentReader.GetTypeDefinition(td);
-                        return IsValueType(d);
-                    default:
-                        return false;
+                    if (s == typeof(ValueType).FullName) return true;
+
+                    var baseType = typeDef.BaseType;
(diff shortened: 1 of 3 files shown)


Engine/PluginSearcher.cs
```diff
@@ -507,6 +507,29 @@ private TypeData PluginFromTypeRef(TypeReferenceHandle handle)
...
+            if (name == nameof(ValueType)) return true;
+
+            var baseType = typeDef.BaseType;
+            switch (baseType.Kind)
+            {
+                case HandleKind.TypeReference:
+                    var tr = (TypeReferenceHandle)baseType;
+                    var r = CurrentReader.GetTypeReference(tr);
+                    return CurrentReader.GetString(r.Name) == nameof(ValueType);
+                    break;
```
rmadsen-ks, line 523:
```suggestion
```

Engine/PluginSearcher.cs
```diff
@@ -507,6 +507,29 @@ private TypeData PluginFromTypeRef(TypeReferenceHandle handle)
...
+                case HandleKind.TypeDefinition:
+                    var td = (TypeDefinitionHandle)baseType;
+                    var d = CurrentReader.GetTypeDefinition(td);
+                    return isValueType(d);
+                    break;
```
rmadsen-ks, line 528:
```suggestion
```

rmadsen-ks (changes requested):
It seem it doesnt build, but otherwise it makes sense.

alnlarsen:
It looks like the valuetype check threw in some cases. I never saw that locally, so I'm glad it happened in the pipelines!

rmadsen-ks (changes requested):
Testing this with "tap benchmark startup --iterations 100" gives this:

this branch:
694.24 ms

main branch:
685.25 ms

So it does seem like this is somewhat slower than what was before, although I had expected it to actually be faster. Maybe we could add some caching of which types are value types to avoid iterating too much over common base types?

alnlarsen:
Thanks a lot for the benchmarks. I will try to squeeze a bit more performance out of this.

alnlarsen:
I ran the benchmarks again:

this branch:
718.74 ms

main branch: 
782.49 ms

Can you please verify these results?

rmadsen-ks (approved):
Looks good

AsgerIversen:
@rmadsen-ks according to a comment from Dennis in the issue, this could also be fixed by https://github.com/opentap/opentap/issues/607. Is that correct, and if so, are we not doing that?

alnlarsen:
@AsgerIversen I don't see the relation between those two issues. There are constructable value types that aren't always preloaded which this solution handles.

PR #780, opened 2022-09-04
Now that we support multiple .gitversion files, we should not rewind …

Denrask:
…to root directory when we start to calculate gitversion.

### DIFF
--- a/Package/CreatePackage/GitVersionAction.cs
+++ b/Package/CreatePackage/GitVersionAction.cs
@@ -77,7 +77,6 @@ namespace OpenTap.Package
                     return (int)ExitCodes.ArgumentError;
                 }
             }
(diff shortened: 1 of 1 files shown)


Package/CreatePackage/GitVersionAction.cs
```diff
@@ -77,7 +77,6 @@ public int Execute(CancellationToken cancellationToken)
                     return (int)ExitCodes.ArgumentError;
                 }
             }
-            RepoPath = repositoryDir;
```
AsgerIversen, line 80: We can remove all of the repositoryDir stuff now, right?

AsgerIversen (changes requested):

AsgerIversen (approved):

PR #782, opened 2022-09-05
Discourage forward slashes

### LINKED ISSUE #550: Packages cannot contain '/' characters. This should be enforced and documented
Package names with path-like symbols are not supported, but we claim they are.

alnlarsen:
This modifies the documentation which states that forward slashes are allowed to instead discourage it
It also adds a warning during package creation to discourage forward slashes.

Closes #550


Package/PackageActions/Create.cs
```diff
@@ -111,7 +115,26 @@ private int Process(string[] OutputPaths, CancellationToken cancellationToken)
...
+                                    log.Warning($"Use of the character '/' in a package name is discouraged. " +
+                                                $"This will never be supported by the package repository, " +
+                                                $"and might lead to problems in the future.");
+                                    break;
+                                default:
+                                    log.Warning($"Use of the character '{ch}' in a package name is discouraged.");
+                                    break;
+                            }
+                        }
+                    }
```
rmadsen-ks, line 136: Can this be simplified to this?
```suggestion
                    foreach (var ch in pkg.Name.Distinct())
                    {
                            switch (ch)
                            {
                                case '/':
                                    log.Warning($"Use of the character '/' in a package name is discouraged. " +
                                                $"This will never be supported by the package repository, " +
                                                $"and might lead to problems in the future.");
                                    break;
                                default:
                                    break;
                            }
                    }
```

alnlarsen:
Thanks for the suggestions. I have committed them!

rmadsen-ks (approved):
Looks great!

PR #783, opened 2022-09-05
781 Scpi Instrument Not Supported on ubuntu 2204: Fix

### LINKED ISSUE #781: ScpiInstrument not supproted on Ubuntu 22.04
we are using libdl.so, but it has been renamed in a way that prevents .net from loading it.

the new name is libdl.so.2, we need to add code to check for that one if the first is unable to load.

rmadsen-ks:
- Added support for loading libdl.so.2 when using Visa.

This was manually tested to work on SCPI enabled instruments on Ubuntu 2204.

Close #781

### DIFF
--- a/Engine/LibDl.cs
+++ b/Engine/LibDl.cs
@@ -1,0 +1,114 @@
+using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+
+static class LibDl
+{
+    static IntPtr load(string name) => libdl.dlopen(name, rtld_now);
+    static void close(IntPtr ptr) => libdl.dlclose(ptr);
+
+    public static string GetError()
+    {
+        var error = libdl.dlerror();
+        if (IntPtr.Zero == error) return null;
+        return Marshal.PtrToStringAuto(error);
+
+    }
+    
+    static void checkError()
+    {
+        var error = libdl.dlerror();
+        if (error != IntPtr.Zero)
+            throw new Exception("Unable to load python: " + error.ToString());
+    }
+
+    static void clearError() => libdl.dlerror();
+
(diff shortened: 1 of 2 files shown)


Engine/LibDl.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    public static IntPtr Sym(IntPtr lib, string name) => libdl.dlsym(lib, name);
+
+    public static IntPtr Load(string name)
+    {
+        clearError();
+        IntPtr p = load(name);
+        if (p == IntPtr.Zero)
```
alnlarsen, line 103: This check looks pointless. Did you mean to return null?

Engine/LibDl.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    public static string GetError()
+    {
+        var error = libdl.dlerror();
+        if (IntPtr.Zero == error) return null;
+        return Marshal.PtrToStringAuto(error);
+
+    }
+    
+    static void checkError()
```
alnlarsen, line 18: It doens't look like this method is used?

Engine/LibDl.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        var error = libdl.dlerror();
+        if (error != IntPtr.Zero)
+            throw new Exception("Unable to load python: " + error.ToString());
```
alnlarsen, line 22: What is this about python?

alnlarsen (changes requested):
It looks like we prefer libdl2 over libdl1 if they are both available. Can this cause issues?

alnlarsen (approved):
Looks good now

PR #784, opened 2022-09-05
751 dont open resources when runmonitors throw

### LINKED ISSUE #751: Resources are Opened after LicenseException
It looks as if OpenTAP opens resources used by a test plan after a LicenseException has been thrown. The exception will cause that the test plan will not be executed so no reason to open the resources.

alnlarsen:
This removes a try {} finally {} clause that would cause resources to be opened when TestPlanRunMonitors throw (which causes the plan to fail to start anyway)

There was a comment warning about why this was required which may or may not be relevant today. The workaround was added in the "first source code drop" commit from 9.0. I tracked the commit back to 2015 where I'm sure it fixed some relevant issue, but a lot has changed since then.

After reviewing the code with @AsgerIversen, we agreed that it is probably not necessary anymore, and we have a unittest that verifies the behavior with all OpenTAP resource managers at least.

Closes #751

### DIFF
--- a/Engine.UnitTests/PlanRunMonitorTests.cs
+++ b/Engine.UnitTests/PlanRunMonitorTests.cs
@@ -68,20 +68,41 @@ namespace OpenTap.Engine.UnitTests
                 base.Open();
             }
         }
+
+        public class UnopenedResultListener : ResultListener
+        {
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks great!

PR #791, opened 2022-09-07
Align targetFramework for Tap.Upgrader to avoid shipping netstandard.dll

### LINKED ISSUE #689: Stop shipping netstandard.dll on Linux

AsgerIversen:
Alternative to #690 

Close #689

### DIFF
--- a/Tap.Upgrader/Tap.Upgrader.csproj
+++ b/Tap.Upgrader/Tap.Upgrader.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net462</TargetFramework>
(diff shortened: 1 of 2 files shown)


alnlarsen:
Let's do this instead. My only concern here is that OpenTAP could add a dependency to `netstandard.dll` from third party plugins now. But let's just tackle that if it actually does become a problem.

alnlarsen (approved):

PR #796, opened 2022-09-08
795 Resource Deserialize Fallback: Improvement

### LINKED ISSUE #795: Fall back to matching name rather than declared type when deserializing resources
A test plan may contain the name and type of serialized resources:
```xml
    <TestStep type="OpenTap.Engine.UnitTests.SerializerTests+StepA">

      <Instr type="OpenTap.Engine.UnitTests.SerializerTests+InstrA" Source="OpenTap.InstrumentSettings">AABBCC</Instr>

      <Name Metadata="Step Name">StepA</Name>
    </TestStep>
```
In this case, the Instr property is actually just an interface e.g IPsu.

In this case it might actually be more useful to grap another IPsu with the same name rather than the first and best 'InstrA'.

Another thing to consider is than we serialize object where the type is matching / 'redundant' we actually dont serialize the type:

```xml
<Instrument Source="OpenTap.InstrumentSettings">SCPI</Instrument>
```

This means that in this case we actually more _preference_ to the name.

This means that we actually dont check for an exact type match, but rather anything that inherits from the type given in the XML, so we might as well just use the property type for resources.

rmadsen-ks:
- Now if a resource matching the XML-declared type and name is not found, we look for one matching the property of the property and name. If everything else fails it takes the one matching the property type.
- Added improvement and unit test that verifies it works.
- Added supprot for collection serializer to specify which element type is currently being deserialized.

 The following are the priorities for resource resolution.
 1. matching name and exact type.
 2. matching name and compatible type.
 3. matching type.
 4. defaulting to (errors emitted) compatible type.

Close #795


alnlarsen (approved):
Looks good.

PR #802, opened 2022-09-13
Make process isolation Linux compatible

### LINKED ISSUE #801: Can't install packages on Linux
This is a regression from a couple of days ago:

https://github.com/opentap/opentap/pull/794

We added process isolation on Linux and assumed it would just work.

```
> ./tap package install Demonstration
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.PlatformNotSupportedException: Message transmission mode is not supported on this platform.
   at OpenTap.ExecutorSubProcess.getStream(String& pipeName)
   at OpenTap.ExecutorSubProcess.Start()
   at OpenTap.Cli.TapEntry.goIsolated()
   at OpenTap.Cli.TapEntry.Go()
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at tap.Program.Main(String[] args)
[1]    151145 IOT instruction (core dumped)  ./tap package install Demonstration
```

It looks like we are using a NamedPipe flag which is not supported on Linux: `PipeTransmissionMode.Message`. From some brief testing it looks like everything still works if we omit this constructor flag.

It looks like we don't have any pipelines testing package installation on Linux. We need a test for this as well.

alnlarsen:
This adds a regression test for verifying that packages can actually be installed on all platforms, and it makes the process isolation logic use cross-platform flags when creating named pipes.

Closes #801

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,6 +175,8 @@ jobs:
           dotnet-version: '6.0.x'
       - name: Build
         run: dotnet publish -c Release OpenTAP.sln
+      - name: Copy Debug Package 
+        run: cp ./bin/Release/OpenTap.Debug.TapPackage ./bin/Release/publish
       - name: Upload binaries
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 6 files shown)


rmadsen-ks (approved):
Looks good and seems to work!

PR #819, opened 2022-09-15
Fix bug causing small packages to have incorrect hashes

alnlarsen:
Because we didn't flush our writer, we never calculated correct hashes. This became apparent for smaller packages which got identical hashes because the memory stream would be empty, but for larger packages we likely also calculated the hash from incomplete data.


Denrask (approved):
This looks good

Denrask (approved):

PR #823, opened 2022-09-15
Redefined the meaning of VersionSpecifier.Any and

### LINKED ISSUE #824: tap package install Any Release / Any does not resolve to the expected packages
Currently the meaning of Any Release, means a compatible version specifier. This means that it actually looks for the oldest version of the package.

So "tap package install OpenTAP" installs the oldest OpenTAP compatible with what you otherwise have in the installation.

```
./tap image install "REST-API:2.6,OpenTAP" --dry-run
Resolved packages:
   Keysight Floating Licensing:    1.0.44+8197912e
   OpenTAP:    9.16.0+6cab4b01
   REST-API:    2.6.3+4b18b59f
```
Where as we expected the following:
```
Resolved packages:
   Keysight Floating Licensing:    1.0.44+8197912e
   OpenTAP:    9.18.5+0328e9b4   (Newest Release)
   REST-API:    2.6.3+4b18b59f
```

### LINKED ISSUE #825: tap package install cannot properly resolve "any" versions
When beta versions exists, `tap package install --version any` cannot resolve the package.

This is because of an optimization that limits the number of requests to the package repository api by looking at the PackageDependencySpecifier version pre-release, but not whether the AnyPreRelease flag has been set.

rmadsen-ks:
VersionSpecifier.AnyRelease

Now the version match behavior is Exact, which means it will map to any version, but always the newest

Closes #824
Closes #825

### DIFF
--- a/Package/Image/PackageDependencyGraph.cs
+++ b/Package/Image/PackageDependencyGraph.cs
@@ -163,9 +163,18 @@ namespace OpenTap.Package
         readonly Dictionary<string, string> currentPreReleases = new Dictionary<string, string>();
         public IEnumerable<SemanticVersion> PackagesSatisfying(PackageSpecifier packageSpecifier)
         {
-            if (!string.IsNullOrWhiteSpace(packageSpecifier.Version.PreRelease))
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
This fixes the issue I observed. I think `version=any` still behaves a bit unexpectedly, but I think `any` has always been kind of undefined.

PR #833, opened 2022-09-16
818 Specify Architecture Or OS without force: Fix

### LINKED ISSUE #818: Running the command to install with a specific --architecture will not return a warning for incompatibility
**Describe the bug**

Using this command on macOS: **./tap package install "Example Plugin" --arhitecture x86** on a x64 device will not return any warning of incompatibility and install directly the x64 version

 - OS: macOS 12.8
 - OpenTAP Version 9.19.0-beta.54

**To Reproduce**

1. Run this command in Terminal **./tap package install "Example Plugin" --arhitecture x86** on a x64 machine

2.  Check if there is a warning message

**Actual behavior**

Using this command on macOS: **./tap package install "Example Plugin" --arhitecture x86** on a x64 device will not return any warning of incompatibility and install directly the x64 version

**Expected behavior**

Trying to install a x86 package on a x64 machine needs to have a warning message and after install de compatible version

rmadsen-ks:
Added warning message.

Close #818


alnlarsen (approved):
Looks good

alnlarsen:
@rmadsen-ks it seems like this doesn't work

alnlarsen:
Never mind, this works as intended

PR #834, opened 2022-09-16
808 Force Install Version Behavior: Fix

### LINKED ISSUE #808: Behavior change with old plugins 9.19 vs 9.18
**Describe the bug**

9.18 would install a plugin missing a dependency with `-f` but 9.19 will not

I am reporting this as a behavior change, not sure if this is a bug or a feature

 - OS: Windows 10
 - OpenTAP Version 9.19.0-rc1

**To Reproduce**

1. `tap package install RemoteSigntool -r "http://packages.opentap.keysight.com" -f`

**Actual behavior**

on **9.19.0-rc.1**, the plugin is not installed. Log [SessionLog 2022-09-14 00-27-07.txt](https://github.com/opentap/opentap/files/9563842/SessionLog.2022-09-14.00-27-07.txt). Note: It will be installed if `--version any` is specified

image

on **9.18.5**, the plugin is installed. Log [SessionLog 2022-09-14 00-27-22.txt](https://github.com/opentap/opentap/files/9563853/SessionLog.2022-09-14.00-27-22.txt)

image

rmadsen-ks:
- Reverted so that AnyRelease.ToString() => "". Previously an empty version meant AnyRelease and some code might still use this convension.
- Also Made a more explicit check for 'AnyRelease in HttpPackageRepository'

Close #808

### DIFF
--- a/Package/PackageSpecifier.cs
+++ b/Package/PackageSpecifier.cs
@@ -17,8 +17,12 @@ namespace OpenTap.Package
     public class PackageSpecifier
     {
         /// <summary> Gets a readable string for this package specifier. </summary>
-        public override string ToString() => $"[{Name} ({Version})]";
-        
+        public override string ToString()
+        {
+            var versionString = Version == VersionSpecifier.AnyRelease ? "Any Release" : Version.ToString();
+            return $"[{Name} ({versionString})]";
+        }
+
         /// <summary>
         /// Search for parameters that specifies a range of packages in the OpenTAP package system. Unset parameters will be treated as 'any'.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Package/PackageSpecifier.cs
```diff
@@ -194,7 +198,7 @@ public override string ToString()
             if (this == VersionSpecifier.Any)
                 return "Any";
             if (this == VersionSpecifier.AnyRelease)
-                return "Any Release";
+                return "";
```
alnlarsen, line 201: Can we add "AnyRelease" / "AnyRelease" / "" to the VersionSpecifier.Parse method?

alnlarsen (changes requested):

alnlarsen:
This fixes the issue observed with `RemoteSigntool`

Package/PackageSpecifier.cs
rmadsen-ks: Lets do this at a later point

alnlarsen:
Okay, I have opened a separate issue for parsing the `AnyRelease` specifier: https://github.com/opentap/opentap/issues/836

PR #842, opened 2022-09-22
835 Compatible Sort Bug: fix

### LINKED ISSUE #835: Installing a package in another folder: Installs old OpenTAP and extra log messages
**Describe the bug**

 - OS: [ Windows 10]
 - OpenTAP Version 9.19.0-beta.58

**To Reproduce**

1. Create a new folder `c:\temp`

2. `tap package install Demonstration -t "c:\temp" -y`

**Actual behavior**

Demonstration 9.0.5 is installed in c:\temp - this is ok
Demonstration has a dependency on OpenTAP ^9.5.0
OpenTAP 9.11.0-beta.4 is installed in c:\temp

Extra `Multiple assemblies` messages in log. [SessionLog 2022-09-16 04-30-30.txt](https://github.com/opentap/opentap/files/9583610/SessionLog.2022-09-16.04-30-30.txt)

image

**Expected behavior**

Which version of OpenTAP should be installed? In this scenario, we are installing a very old version. Should we consider the latest compatible `Release`. Or the installed -beta version if compatible

Should we display the `Multiple assemblies` messages?

rmadsen-ks:
Fixed the problem by changing the sort order for compatible but not-prerelease version specifiers.

Close #835

### DIFF
--- a/Package/Image/ImageResolver.cs
+++ b/Package/Image/ImageResolver.cs
@@ -193,6 +193,8 @@ namespace OpenTap.Package
             for (int i = 0; i < allVersions.Count; i++)
             {
                 var pkg = packages[i];
+                // skip the following if there is less than two versions.
+                if (allVersions[i].Length <= 1) continue;
                 var versions = allVersions[i].ToList();
                 
                 if (pkg.Version == VersionSpecifier.Any)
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
This looks good to me. With this fix, I get OpenTAP 9.13.0 due to my system-wide packages, which is a bit strange, but nonetheless the expected behavior now.

AsgerIversen:
I kind of like #843 better, we need more time to dig into that. We'll do this for now as a quick and safe solution for the 9.19.1 patch.

PR #857, opened 2022-10-03
856 Disallow Parameterizing Swept Property: Fix

rmadsen-ks:
- Fixed the issue by adding an (IEnabledAnnotation) annotation when the correct pattern is detected.
- Added a unit test to verify the behavior.

image


alnlarsen:
The fix works. But if a parameter is already parameterized on the test plan, the sweep loop still allows sweeping it. Is this a simple thing to fix?

alnlarsen:
Everything seems to work as intended!

alnlarsen (approved):

PR #859, opened 2022-10-03
858 improve error message for invalid version specs

### LINKED ISSUE #858: null reference during 'tap package create'
If the version in a package.xml file has four fields (e.g. 1.2.3.4), we encounter a null reference exception. We should handle this exception better.

alnlarsen:
When a package has an invalid version specifier, we now fail immediately after parsing during package create.

Closes #858


rmadsen-ks:
Looks good to me!

image

Works for semantic versions.

If Version="" is specified the following message (info) is emitted:
```
Package version is 0.0.0 due to blank or missing 'Version' XML attribute in 'Package' element
```

rmadsen-ks (approved):

PR #865, opened 2022-10-05
864 mimic pre-9.19 URI behavior when BaseAddress is not set

### LINKED ISSUE #864: Relative URL problems
The relative URL changes in 9.19 are wreaking havoc. Consider the following example, provided by us:

https://doc.opentap.io/Developer%20Guide/Getting%20Started%20in%20Visual%20Studio/#reference-other-opentap-packages

This is no longer going to work because `packages.opentap.io` is now interpreted as a file path. Why? Because it is impossible for us to distinguish a relative URL from a file path. We therefore now require URI schemes. As a result, projects that used to build with 9.18 may not build with 9.19.

alnlarsen:
This fixes some behavior changes introduced by 9.19, such as not being able to list packages using `tap package list -r packages.opentap.io` due to the way relative URIs are resolved.

Closes #864


rmadsen-ks (dismissed):
This looks good to me. Let's remember to verify that it needs to work at:
- tap package list
- tap package install
- PackageManager.exe (in UI)
- And in csproj OpenTapPackageReference

rmadsen-ks (approved):
Looks good and seems to work well.

sebastian-pop:
Fixed. Verified on 9.19.2-rc.1

image

PR #873, opened 2022-10-13
Improve CLI Error message

rmadsen-ks:
Improved the CLI error message by removing the part of the message which would be confusing to non-programmers or people who are not familiar with the ICliAction API.


AsgerIversen (approved):

sebastian-pop:
Fix confirmed on 9.19.2-rc.1

On 9.19.1

image

On 9.19.2-rc.1

image

PR #885, opened 2022-10-19
884 Allow Setting Package Cache: Fix

rmadsen-ks:
Added support for setting the environment OPENTAP_PACKAGE_CACHE, which affects the package cache location.

Tested it by setting it and calling "tap package install" and everything seems to work.

### DIFF
--- a/Package/PackageCacheHelper.cs
+++ b/Package/PackageCacheHelper.cs
@@ -18,7 +18,7 @@ namespace OpenTap.Package
         // - Linux: /home/<USER>/.local/share/OpenTAP/PackageCache
         // - Windows: C:\Users\<USER>\AppData\Local\OpenTAP\PackageCache
         //
-        // Unless OPENTAP_PACKAGE_CACHE is set, then that path will be used instead.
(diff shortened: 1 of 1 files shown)


Package/PackageCacheHelper.cs
```diff
@@ -17,7 +17,21 @@ static class PackageCacheHelper
...
+            get
+            {
+                var alt = Environment.GetEnvironmentVariable(PackageCacheOverrideEnvironmentVariable);
+                if (string.IsNullOrWhiteSpace(alt) == false)
+                    return alt;
+                return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData, Environment.SpecialFolderOption.Create), "OpenTap", "PackageCache");
+            }
+        } 
+
+        const string PackageCacheOverrideEnvironmentVariable = "OPENTAP_PACKAGE_CACHE";
```
AsgerIversen, line 33: Should we prepend "DIR" to the name?

AsgerIversen (changes requested):

AsgerIversen (approved):

PR #895, opened 2022-10-26
872 Scpi Instrument Virtual Methods: Fix

### LINKED ISSUE #872: Ability to override *CLS from ScpiInstrument Open
ScpiInstrument Open will always send clear status command to instrument. However not all scpi instruments supports that command.
Could it be ok to create separate method for clear state and then call that in Open? That way in derived class one could override clear state or change the command something suitable for the particular instrument.

Method be like:
```csharp
public void ClearStatus()
{
    ScpiCommand("*CLS");
}
```

Then this :   
`ScpiCommand("*CLS"); // Empty error log`
Into this:  
`ClearStatus(); // Empty error log`

rmadsen-ks:
Added a way to override the common scpi commands used in ScpiInstrument.

- IDN
- OPC
- RST
- CLS
- SYST:ERR

Close #872

### DIFF
--- a/Engine/ScpiInstrument.cs
+++ b/Engine/ScpiInstrument.cs
@@ -377,11 +377,11 @@ namespace OpenTap
                         base.Open();
                         SetTerminationCharacter(scpiIO.ID);
 
-                        IdnString = ScpiQuery("*IDN?");
+                        IdnString = QueryIdn();
                         IdnString = IdnString.Trim();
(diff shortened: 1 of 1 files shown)


AsgerIversen (approved):
I like it. Any reason this is still a "draft" @rmadsen-ks ?

rmadsen-ks:
@AsgerIversen, this is a suggestion for an implementation. If you think this is the right direction, lets merge it.

PR #899, opened 2022-10-27
898 stepnameannotationvalue on teststep throws an exception

### LINKED ISSUE #898: StepNameAnnotation.Value on TestStep[] throws an exception
**Describe the bug**
StepNameAnnotation.Value on TestStep[] throws an exception

 - OS: Windows 10
 - OpenTAP Version 9.19

**Expected behavior**

The annotation should not throw an exception, instead the value should be null if all teststeps dont have the same name, and otherwise it should be the shared name of all the teststeps in the array.

frederikja163:
Closes #898

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -1667,17 +1667,23 @@ namespace OpenTap
             {
                 get
                 {
-                    ITestStep value;
+                    object value;
                     if (member)
                     {
-                        value = (ITestStep) annotation.ParentAnnotation.Get<IObjectValueAnnotation>().Value;
+                        value = annotation.ParentAnnotation.Get<IObjectValueAnnotation>().Value;
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


Engine/Annotations/Annotation.cs
```diff
@@ -1667,17 +1667,23 @@ public string Value
...
-                        value = (ITestStep)annotation.Get<IObjectValueAnnotation>().Value;
+                        value = annotation.Get<IObjectValueAnnotation>().Value;
                     }
 
-                    return value.GetFormattedName();
+                    if (value is ITestStep step)
+                    {
+                        return step.GetFormattedName();
+                    }
+                    IEnumerable<ITestStep> steps = (IEnumerable<ITestStep>)value;
```
alnlarsen, line 1684: Can we type match here instead of casting? And do something a bit more defensive than `steps.First()`

alnlarsen (changes requested):
This looks good. Can you make one minor change?

alnlarsen (approved):
Looks good!

PR #907, opened 2022-11-03
786 add sweep value from file

### LINKED ISSUE #786: add sweep values from file
Right now, Sweep  values can only be manually loaded from a file with the Edit window, but they can't be combined with other Test Plan Parameters. Ideally this can be exposed some how (either parameterizing the Loaded file itself or the lower level table)   
image

ivandiep-ks:
Closes #786

### DIFF
--- a/BasicSteps/SweepParameterStep.cs
+++ b/BasicSteps/SweepParameterStep.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
@@ -31,6 +32,11 @@ namespace OpenTap.Plugins.BasicSteps
             }
         }
 
+        public AnnotationCollection annotations;
+
+        [FilePath(fileExtension: "csv")]
+        [Display("Import Sweep Values", "Import sweep values from a .csv file", "Import", 2, Collapsed: true)]
+        public string csvSweepValues { get; set; }
 
         /// <summary>
         /// This property declares to the Resource Manager which resources are declared by this test step. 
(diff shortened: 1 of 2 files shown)


bdirenzo-keysight:
@rmadsen-ks / @AsgerIversen are one of you able to review this? I'd like to get it into the next release as the customer is waiting on it. At a minimum we'd like a build that is in the repo that can be used (even if only alpha/beta version)

BasicSteps/SweepParameterStep.cs
```diff
@@ -224,11 +230,41 @@ public override void PrePlanRun()
             base.PrePlanRun();
             iteration = 0;
 
+            if (csvSweepValues != null && csvSweepValues != "")
+            {
+                // Populates sweep tables with csv from import path
+                var timer = System.Diagnostics.Stopwatch.StartNew();
+                Type plugin = PluginManager.GetPlugins<ITableImport>().First();
```
alnlarsen, line 237: OpenTAP has no implementations of `ITableImport`. I assume this works for you because you have the `CSV` plugin installed, but we can't make that assumption.

BasicSteps/SweepParameterStep.cs
```diff
@@ -224,11 +230,41 @@ public override void PrePlanRun()
...
+                var fileHandler = (ITableImport)Activator.CreateInstance(plugin);
```
alnlarsen, line 238: You probably want something like `TypeData.GetDerivedTypes<ITableImport>().FirstOrDefault(td => td.CanCreateInstance)?.CreateInstance();` instead. We don't really use `PluginManager.GetPlugins` anymore.

BasicSteps/SweepParameterStep.cs
```diff
@@ -224,11 +230,41 @@ public override void PrePlanRun()
             base.PrePlanRun();
             iteration = 0;
 
+            if (csvSweepValues != null && csvSweepValues != "")
```
alnlarsen, line 233: Use `string.IsNullOrWhiteSpace(csvSweepValues)`

BasicSteps/TableView.cs
```diff
@@ -0,0 +1,282 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace OpenTap.Plugins.BasicSteps
+{
+    class TableView
```
alnlarsen, line 8: Is this copy-pasted from the CSV plugin?

alnlarsen (changes requested):
I don't really like that we're adding a property to an OpenTAP step that depends on another plugin. 

Have you considered creating your own SweepParameterStep in your own plugin instead?

alnlarsen:
If we add this feature to OpenTAP, at the very least we shouldn't depend on the CSV plugin. 

Another option could be to implement this as an attached property in the CSV plugin instead:
https://github.com/opentap/opentap/blob/main/sdk/Examples/PluginDevelopment/Advanced%20Examples/AttachedPropertyExample.cs

BasicSteps/SweepParameterStep.cs
```diff
@@ -31,6 +32,11 @@ public SweepRowCollection SweepValues
             }
         }
 
+        public AnnotationCollection annotations;
+
+        [FilePath(fileExtension: "csv")]
+        [Display("Import Sweep Values", "Import sweep values from a .csv file", "Import", 2, Collapsed: true)]
+        public string csvSweepValues { get; set; }
```
alnlarsen, line 39: We need to add `[Unsweepable]` to this property.

PR #908, opened 2022-11-03
906 Test Plan License Exception improvements

rmadsen-ks:
Improved the way exceptions are printed when they occur during deserialization by adding a general HandleError method.

Added printing the exception directly in the error and not some generic error message.

Added ToString overloads for TypeData specilizations.


alnlarsen (approved):
This looks good :)

PR #927, opened 2022-11-22
Build and publish installer for windows and linux

alnlarsen:
With this we will be able to always download the latest installer from:

Windows: `https://packages.opentap.io/4.0/Objects/www/OpenTAP.exe(?version=specifier)`
Linux: `https://packages.opentap.io/4.0/Objects/www/OpenTAP(?version=specifier)`


StefanHolst (approved):
Looks good

PR #932, opened 2022-12-01
Feature make visa pluggable

alan-copeland-keysight:
Add the infrastructure to allow alternate VISA implementations to be used by OpenTAP. For example,   https://github.com/alan-copeland-keysight/PythonVisa is a Python plugin implementation that uses pyvisa/pyvisa-py to provide VISA connections on platforms not supported by a major I/O vendor (tested on Raspberry Pi 4 and Mac x64).

By default this change retains the existing OpenTAP behavior of using the System VISA. If no System VISA is detected, this mechanism can be used to automatically fallback on an alternate I/O plugin that is installed (e.g. PythonVisa mentioned above as a concrete example).

This PR also enables the use of NI-VISA on MacOS using OpenTAP's built-in system VISA lookup=


alan-copeland-keysight:
@rmadsen-ks

alan-copeland-keysight:
Have a pointer to the format / content you'd like for the xml comments?

rmadsen-ks:
@alan-copeland-keysight, not particularly. I would say just add some one-liner comments.

e.g:
```<summary>Activates the thing on the device.</summary>```

**Comment by alan-copeland-keysight** [comment]
Ok, I've added method summaries based on the content in the Keysight VISA help file

File: Engine/Visa.cs [hunk 1]
```diff
@@ -4,746 +4,1043 @@
...
+        #endregion
+
+        #endregion
     }
 
-    static Visa()
+    /// <summary>
+    /// A mechanism for retrieving IVisa implementations.
+    /// </summary>
+    public interface IVisaProvider : ITapPlugin {
```
rmadsen-ks, line 735: Can we refactor these new classes into separate files?

Engine/Visa.cs
```diff
@@ -4,746 +4,1043 @@
...
+        /// <summary>Event handler prototype</summary>
+        delegate int viEventHandler(int vi, int eventType, int context, int userHandle);
+    }
 
-        Func<IntPtr> LoadLib;
-        Func<IntPtr, string, int, IntPtr> LoadSym;
+    /// <summary>
+    /// Attempt to load VISA libraries from the system
+    /// </summary>
+    public class VisaLibraryLoader : IVisa, IVisaProvider
```
rmadsen-ks, line 807: I think this can be internal. OpenTAP sets PluginAssembly attribute so internal types are also shown as plugins.

Engine/Visa.cs
```diff
@@ -4,746 +4,1043 @@
...
+            var instances = new SortedDictionary<double, (string, IVisa)>();
+            foreach (var typeDataIVisa in visaProviders) 
+            {
+                if (typeDataIVisa != null)
+                {
+                    try 
+                    {
+                        staticLog.Debug("Attempting to load IVisaProvider: {0}", typeDataIVisa.Name);
+                        IVisaProvider providerInstance = (IVisaProvider)typeDataIVisa.CreateInstance();
+                        if (providerInstance.Visa != null)
```
rmadsen-ks, line 94: maybe it is better to first sort the instances of ITypeDataProviders and then take the first non-null visa instance?

This way we don't need to call Visa on the low-priority  providers, just in case this is a slow action.

Engine/Visa.cs
```diff
@@ -4,746 +4,1043 @@
...
+            object instance = null;
+            var visaProviders = TypeData.GetDerivedTypes<IVisaProvider>();
+            var instances = new SortedDictionary<double, (string, IVisa)>();
+            foreach (var typeDataIVisa in visaProviders) 
+            {
+                if (typeDataIVisa != null)
+                {
+                    try 
+                    {
+                        staticLog.Debug("Attempting to load IVisaProvider: {0}", typeDataIVisa.Name);
```
rmadsen-ks, line 92: Maybe we should remove this log message and add one in the catch clause. 

This works best if we dont try to call Visa on low-priority providers.

Engine/Visa.cs
```diff
@@ -4,746 +4,1043 @@
...
+                    catch {}
+                }            
+            }
+            if (instances.Count == 0)
+            {
+                throw new Exception("IVisaProvider unable to load VISA");
+            }
+            
+            instance = instances.Values.First().Item2;
+            staticLog.Debug("Using IVisaProvider: {0}", instances.Values.First().Item1);
```
rmadsen-ks, line 109: maybe we dont need this if we order a bit earlier. In the default use case it should just say "Loaded IVisaprovider  X" and nothing else.

rmadsen-ks (changes requested):

Engine/Visa.cs
alan-copeland-keysight: The order is currently a property on the instance. Is there some other mechanism to get the order @rmadsen-ks ?

alan-copeland-keysight: In the default case, there will only be the default/system IVisaProvder, so maybe we just optimize that to be the fast path?

Engine/Visa.cs
alan-copeland-keysight: See responses on comments above. I don't see a path to reorder differently.

alnlarsen:
@rmadsen-ks is this something we want in 9.20? If so, is this something you can approve?

alan-copeland-keysight:
@alnlarsen - I think @rmadsen-ks had made some changes and created a newer PR - https://github.com/opentap/opentap/pull/934

alan-copeland-keysight:
I will close this PR, and we can focus on #934

PR #937, opened 2022-12-09
935 breakcondition inherit toggle reset value

### LINKED ISSUE #935: Break condition value is reset when the enabled checkmark is toggled
If `Break Conditions` is manually set to `None`, and the setting is disabled and then re-enabled, the value of the setting is reset to the default value. It should retain the previous value.

frederikja163:
closes #935 
Save old value of BreakConditionDynamicMember
Use old value in case inherit flag is removed


rmadsen-ks:
Unfortunately we decided to not fix this issue.

PR #938, opened 2022-12-09
846 lattest session log

### LINKED ISSUE #846: SessionLogs/Latest
Currently tap creates a sessionlog every time it starts up, but the newest sessionlog always has a new name, making it harder than necessary to find it. I suggest we make either a symlink og a hardlink to the lattest sessionlog and save it in a SessionLog/Latest.txt file, so the latest session log is always easy to find.

frederikja163:
Create a hardlink to SessionLogs/Latest.txt for the latest sessionlog Should be useful when wanting to open the latest sessionlog without knowing what the name of the file is.

Closes #846


rmadsen-ks (changes requested):
I think it looks awesome, but we could improve it a bit.

frederikja163:
This should now work on Linux, Mac and Windows, however i havent been able to test on either Linux or Mac.

alnlarsen:
This works on Linux. I will test on Mac when I get home.

alnlarsen:
It works on Mac

alnlarsen (approved):
This looks good to me.

AsgerIversen (approved):

PR #939, opened 2022-12-13
armPackageSupport

### LINKED ISSUE #1158: Support more OSs
We should have native packages for arm architectures for something like: macOS with M1 and Raspberry PI.

Without native packages there seems to be issues when installing packages and resolving native dependencies.

StefanHolst:
Creating new additional opentap packages for arm

Closes #1158


alnlarsen (changes requested):
This looks good, but can you create an issue explaining why we are making this change?

PR #942, opened 2022-12-16
Register executables from packages as CLI actions

### LINKED ISSUE #941: Register shipped executables as CLI actions
Other tools such as `git` and `docker` will register executables from some predefined locations as cli actions if the executable is prefixed with `git-` or `docker-`, e.g. `docker` will call `docker-compose` when the action `docker compose` is run. 

We could do something similar for OpenTAP. I propose registering executables from packages within an installation to be detected as CLI actions, so if my package ships `tap-smartinstaller-inject`, then OpenTAP should detect the executable and recognize the CLI action `tap smartinstaller inject`. Then OpenTAP should forward all arguments after the `inject` to the subprocess.

alnlarsen:
This adds a typedata searcher which adds a CLI action for any file from
a package which matches `^tap-.+`

We might need to add a few restrictions, such as a stricter regex, and requiring the name to end with `.exe` on windows, and for the file to have the Executable flag on Linux and MacOS, though I was a bit hesitant to add that restriction because I thought we might want to allow e.g. Python files to be registered as CLI actions. (so my file, `Packages/MyPlugin/tap-calibrate_dut.py` could be registered as a CLI action.

As an example I have included two OpenTAP CLI actions for starting OpenTAP with `.net6` and `.net7`. I have added the file `Packages/OpenTAP/net7/tap-runtime-net7`, which adds a CLI action equivalent to:

```csharp
[DisplayAttribute("net7", "Packages/OpenTAP/net7/tap-runtime-net7 from OpenTAP", new string[]{ "runtime" })]
public class RuntimeNet7 : ICliAction 
{
  public int Run(CancellationToken token) 
  {
    // run executable with the arguments that followed `tap-runtime-net7` in a subprocess
  }
}
```

Closes #941


Package/ExternalCliAction.cs
```diff
@@ -0,0 +1,136 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using OpenTap.Cli;
+
+namespace OpenTap.Package
```
rmadsen-ks, line 10: Split this file into one for each class.

No need to add the internal specifier for each class, that is implied.

Package/PackageActions/IsolatedPackageAction.cs
```diff
@@ -127,7 +127,7 @@ PackageDef findPackageWithFile(string file)
                     {
                         foreach (var pkgfile in package.Files)
                         {
-                            var filePath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), pkgfile.FileName);
+                            var filePath = Path.Combine(ExecutorClient.ExeDir, pkgfile.FileName);
```
rmadsen-ks, line 130: This seems a bit risky. What difference does this make in practice?

Package/PackageFile.cs
```diff
@@ -207,6 +207,43 @@ public override bool Equals(object obj)
         public override int GetHashCode() =>  (Name ?? "").GetHashCode() * 7489019 + (RawVersion ?? "").GetHashCode() * 41077013;
     }
 
+    /// <summary>
+    /// Represents an external command that can be executed as a CLI action
+    /// </summary>
+    public class ExternalCliAction
```
rmadsen-ks, line 213: Consider moving this to a separate file.

tap/Program.cs
```diff
@@ -23,7 +23,27 @@ static void Main(string[] args)
 
                 Assembly load(string file)
                 {
-                    file = Path.Combine(appDir, file);
+                    var originalFile = file;
+                    string GetParent(string path, int level)
```
rmadsen-ks, line 27: Can you add a comment here specifying what you are trying to do? Or encapsulate this functionality in a separate method?

rmadsen-ks (changes requested):
I am not sure how this works or what exactly the purpose is. I had some comments for the code, but maybe you could add some more information to the PR, showing how this is supposed to be used?

For example, how does it look if I want to run a test plan in a .net 7 process?

PR #943, opened 2022-12-19
902 Add iteration display to SweepParameterRangeStep

### LINKED ISSUE #902: Add iteration functionality to SweepParameterRangeStep
SweepParameterStep contains a parameter IterationInfo with the current iteration number, as well as the total number of iterations the step will run. I can't find a parameter with the same functionality in SweepParameterRangeStep, and it would be nice to have

frederikja163:
closes #902 

Add an iteration property to SweepParameterRangeStep

### DIFF
--- a/BasicSteps/SweepParameterRangeStep.cs
+++ b/BasicSteps/SweepParameterRangeStep.cs
@@ -44,7 +44,7 @@ namespace OpenTap.Plugins.BasicSteps
         public uint SweepPoints { get; set; }
 
         private int iteration = 0;
-        [Output]
+        [Output(OutputAvailability.BeforeRun)]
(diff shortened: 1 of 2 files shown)


rmadsen-ks (dismissed):
Works perfectly!

AsgerIversen (changes requested):
Nice addition. 

I think there is a good point with the `OutputAvailability.BeforeRun` that @cpetku uses in [the patch](https://github.com/opentap/opentap/issues/902#issuecomment-1322617603) in the issue. 

@alnlarsen, @rmadsen-ks what do you say we add that to the output here and in SweepParameterStep?

Without `OutputAvailability.BeforeRun`:

image

With:

image

frederikja163:
@AsgerIversen I have tagged both outputs as available BeforeRun

AsgerIversen (approved):

PR #948, opened 2022-12-19
946 Run Isolated Test Plans: Fix

rmadsen-ks:
- Added Nested Test Plan, which inherits from Test Plan Reference and support running with Session.
- Added Unit test that verifies the behavior.
- Fixed ExpandedMemberData so that it supports sub classes of TestPlanReference.


BasicSteps/NestedTestPlan.cs
```diff
@@ -0,0 +1,136 @@
...
+                
+                // override the plans break conditions with the break conditions of this step.
+                // otherwise the nested plan will get the break conditions from engine settings, which would be
+                // inconsistent with how it otherwise works.
+                BreakConditionProperty.SetBreakCondition(plan2, BreakConditionProperty.GetBreakCondition(this));
+                
+                // this is what actually runs the test plan.
+                // since it is running inside a session, it can be aborted.
+                // it has very limited ways to affect the outer test plan.
+                var subRun = plan2.Execute((ForwardResults ? new IResultListener[] {new ResultForwarder(Results)} as IEnumerable<IResultListener> : ResultSettings.Current));
```
alnlarsen, line 129: What happens if `plan2` requires resources held by the parent test plan?

alnlarsen (commented):
I think this looks pretty promising. I just have some questions

BasicSteps/NestedTestPlan.cs
rmadsen-ks: That is why it gets serialized and deserialized. If it is e.g a SCPI instrument, it should work fine. If it is a COM port, it wont work, since the resource becomes locked by first use. 

It depends a bit on how the resources are used. For most, I think it will work.

alnlarsen (approved):
Looks good!

PR #953, opened 2023-01-05
Fixes to hardened Docker image

MikaelElkiaer:
~~It seems that OpenTAP requires `$HOME` to be set and for that dir to be writeable for the running user.
I've chosen `/var/lib/tap` for this, as it seems to be the convention for system processes.~~

~~Note: It is not enough to set `OPENTAP_PACKAGE_CACHE_DIR`, as the home dir is also used for logs and ID.~~

Fixes include:

- Proper set up of installation and cache dirs
- Set user to UID instead of user name (user name could hide root user)

### DIFF
--- a/docker/Linux/Dockerfile
+++ b/docker/Linux/Dockerfile
@@ -43,19 +43,25 @@ RUN apt-get update && apt-get install -y --no-install-recommends libunwind8 curl
 ENV PATH="/opt/tap:${PATH}"
 ENV TAP_PATH="/opt/tap"
 
+### Hardened security - will run as non-root user by default 
 FROM production AS production-hardened
 
-# Create non-root system user and group
-ARG USER=opentap UID=101 GROUP=opentap GID=101 DUMMYSHELL=/bin/false
-ARG USERHOME="/home/${USER}"
-RUN groupadd --system --gid=$GID $GROUP && \
-  useradd --system --uid $UID --gid $GID --shell $DUMMYSHELL $USER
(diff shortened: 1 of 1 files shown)


docker/Linux/Dockerfile
```diff
@@ -43,19 +43,29 @@ RUN apt-get update && apt-get install -y --no-install-recommends libunwind8 curl
...
+# Create opentap user and group
+# - set home dir, since OpenTAP expects home dir to maintain cache
+# - no default shell, as user should only run tap
+ARG USER=opentap UID=1001 GROUP=opentap GID=1001 DUMMYSHELL=/bin/false
+ARG HOMEDIR=/home/${USER}
+ARG OPENTAP_CACHE=${HOMEDIR}/.local/share/OpenTap
+RUN groupadd --gid=$GID $GROUP && \
+  useradd --uid $UID --gid $GID --shell $DUMMYSHELL --home-dir $HOMEDIR $USER && \
+  install --directory --owner $USER --group $GROUP --mode u=rwx,g=rwx,o= $HOMEDIR && \
+  install --directory --owner $USER --group $GROUP --mode u=rwx,g=rwx,o= $OPENTAP_CACHE
```
alnlarsen, line 58: Should we create the directories with `rw` instead of `rwx`?

alnlarsen: Nevermind

alnlarsen (approved):
Looks good!

PR #958, opened 2023-01-12
Add PrepareUninstall package action verb

### LINKED ISSUE #957: Feature: PrepareUninstall package action verb
We support the `uninstall` verb in package action extensions in cases where a plugin needs to clean up after itself. Before this action is run, however, OpenTAP checks if ANY file from any package being uninstalled is currently in use. 

This causes issues if part of the "clean up" of a package is to ensure that its files are no longer in use.

I propose adding the `PrepareUninstall` verb, which runs immediately before a package is uninstalled, even if package files are already in use.

alnlarsen:
adds the PrepareUninstall ActionName to PackageActionExtension. A PrepareUninstall action runs right before uninstalling a package, before OpenTAP checks whether files are in use. This can be used to ensure that files in the installation are released if this is the case.

Closes #957


AsgerIversen (approved):
Looks good.

PR #959, opened 2023-01-12
doc: add prepare uninstall documentation

### LINKED ISSUE #957: Feature: PrepareUninstall package action verb

alnlarsen:
This adds documentation for the new PrepareUninstall verb

Closes #957

### DIFF
--- a/doc/Developer Guide/Plugin Packaging and Versioning/Readme.md
+++ b/doc/Developer Guide/Plugin Packaging and Versioning/Readme.md
@@ -262,7 +262,13 @@ A package can contain any number of **ActionStep** elements, but they must be co
 </Package>
 ```
 
-The above example is an excerpt of a plugin.xml file that defines a plugin that includes a test plan to verify that the steps it provides are working correctly, and a binary executable for generating debug waveform data. It also contains CLI actions that allow further configuration when it is installed, and a CLI action that it needs to run before it is uninstalled, and a CLI action that it needs to run when it is uninstalled. All this is done using the **ActionStep** elements.
+The above example plugin definition makes use of the following features:
+
++ A CLI action to be run when it is installed
++ A CLI action to be run when OpenTAP is preparing to uninstall it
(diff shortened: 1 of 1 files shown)


doc/Developer Guide/Plugin Packaging and Versioning/Readme.md
```diff
@@ -257,7 +262,8 @@ A package can contain any number of **ActionStep** elements, but they must be co
 </Package>
 ```
 
-The above example is an excerpt of a plugin.xml file that defines a plugin that includes a test plan to verify that the steps it provides are working correctly, and a binary executable for generating debug waveform data. It also contains CLI actions that allow further configuration when it is installed, and a CLI action that it needs to run when it is uninstalled. All this is done using the **ActionStep** elements.
+The above example is an excerpt of a plugin.xml file that defines a plugin that includes a test plan to verify that the steps it provides are working correctly, and a binary executable for generating debug waveform data. It also contains CLI actions that allow further configuration when it is installed, and a CLI action that it needs to run before it is uninstalled, and a CLI action that it needs to run when it is uninstalled. All this is done using the **ActionStep** elements.
```
AsgerIversen, line 265: Looks like a "typo" that `and a CLI action that it needs to run when it is uninstalled` is now here twice

AsgerIversen (changes requested):
Looks good. One small typo

alnlarsen:
I ended up changing the whole sentence to an itemized list because it was getting out of hand.

AsgerIversen (approved):

PR #963, opened 2023-01-17
Remove TOC for release notes

alnlarsen:
We never link to this page. We forget to update it. Let's remove it.

This also removes a package.json development command which uses syntax which is not compatible with Windows


doc/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "vuepress build",
-    "dev": "NODE_ENV=development vuepress dev"
+    "dev": "vuepress dev"
```
StefanHolst, line 8: Do we not need this anymore? If not, let's remove the use in the `.vuepress/config.js`

StefanHolst (changes requested):
Minor comment

doc/package.json
alnlarsen: It doesn't work on Windows, and I got tired of removing it

alnlarsen: It only affects the external link to the API reference. I will remove it from the config

StefanHolst (approved):
Looks good

PR #965, opened 2023-01-17
Just fixed some unit tests and updated nuget packages.

nerdlark:
see commit message for more information

### DIFF
--- a/BasicSteps/Tap.Plugins.BasicSteps.csproj
+++ b/BasicSteps/Tap.Plugins.BasicSteps.csproj
@@ -11,11 +11,11 @@
   </PropertyGroup>
   -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.10.0" />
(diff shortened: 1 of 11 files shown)


rmadsen-ks:
Hi @nerdlark.

Thanks for submitting a PR. 

In general, changing DLL package references (especially to breaking versions) is a bit tricky and might break existing dependencies. So we normally don't change them unless there is a very good reason. 

If it is because you have a conflict with a dll version you want to use, you can use a .OpenTapIgnore file to override a specific DLL dependency. You can read more about that at doc.opentap.io. 

Could you clarify why you need these specific versions compared to the previous versions? 

As for your changes with regards to the unit tests, I would recommend fixing the current culture, which is what we generally do, but that is normally something you manually need to specify during app start. If we want to fix this I think we should fix it for all cultures and not just the ',' using ones. What do you think about that? 

As a side note, to merge this we need you to sign a CLA in order for us to accept the changes. You can read more about it here https://github.com/opentap/opentap/blob/main/CONTRIBUTING.md

PR #969, opened 2023-01-19
Fix issue where package.xml file might get corrupted 

AsgerIversen:
`tap package create --install` would open any existing file and start writing the new content to the beginning. If the file already existed, and had more content than the new content, partial stuff would be left at the end of the file.

### DIFF
--- a/Package/PackageActions/Create.cs
+++ b/Package/PackageActions/Create.cs
@@ -207,7 +207,7 @@ namespace OpenTap.Package
                 {
                     var path = PackageDef.GetDefaultPackageMetadataPath(pkg, Directory.GetCurrentDirectory());
                     Directory.CreateDirectory(Path.GetDirectoryName(path));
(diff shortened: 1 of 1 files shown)


alnlarsen (dismissed):
This looks good to me. 

Were you able to reproduce the issue? It looks very likely that this was the problem.

I am working on fixing the pipeline issue in a separate branch

AsgerIversen:
@alnlarsen yes, I could reproduce. As far as I saw, it will only happen once after you make some change that affects the package.xml (e.g. add/remove a dependency). The second time some other code would move the now "broken" package xml file out of the way before we get to this point.

alnlarsen (approved):
Looks good

AsgerIversen:
@rmadsen-ks The file should just get overwritten, if it already exists. 

https://learn.microsoft.com/en-us/dotnet/api/system.io.filemode?view=net-7.0#system-io-filemode-create

PR #977, opened 2023-01-23
976 CLS and RST to use ScpiCommand

### LINKED ISSUE #976: *CLS / *RST erroneously setting viRead after command (incorrectly using ScpiQuery instead of ScpiCommand)
The issue was introduced in this PR (Scpi Instrument Virtual Methods):

*CLS and *RST should use ScpiCommand, not ScpiQuery

https://github.com/opentap/opentap/pull/895

rmadsen-ks:
Fixed the issue by changing to using ScpiCommand instead of query for RST and CLS.

This fixes a regression introduced.

Close #976


alan-copeland-keysight:
PR looks good!

image

AsgerIversen (approved):

PR #978, opened 2023-01-23
970 search async memory leak fix

### LINKED ISSUE #970: Memory Leak : if PluginManager.SearchAsync()  called repeatedly
**Describe the bug**

We found memory leak happening if the PluginManager.SearchAsync() being called repeated. 

 - OS: [e.g. Windows 10]
 - OpenTAP Version :  9.19.4 , 9.15.4 and it is abit lower in 9.10.4 

**To Reproduce**

1. Create a loop 

2.  inside the loop call the PluginManager.SearchAsync() , Load the Testplan and execute
 
3 . If the PluginManager.SearchAsync() is located outside of the loop , the same scenario is not happening

**Actual behavior**

This cause the application to crash

**Expected behavior**

The object should be release properly before going to the next iteration.

**Additional context**

Experiment Project https://github.com/darwinsubramaniam/SampleOpenTapExecutor 

It contain the memory profiling file captured via Visual Studio 2022 for respective OpenTAP version.

2 Type of Scenario capture: 

1. PluginManager.SearchAsnyc() inside the loop **( Report name Contains "Repeat" )**
2. PluginManager.SearchAsync() outside the loop **( Report name Contains "NonRepeat" )**

rmadsen-ks:
The formally fixes #970. 

The issue was caused by a concurrent dictionary cache holding on to references to previous ITypeData instances.

Although this should generally not be an issue, this can save a few MBs of memory from time to time.

Another solution could have been to move to conditional weak tables, but they use reference equality instead of object equality which make it a not so great general solution.

Close #970

### DIFF
--- a/Shared/Cache.cs
+++ b/Shared/Cache.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Immutable;
+namespace OpenTap
+{
+    /// <summary> Cache class which clears when a certain criteria is met. </summary>
+    internal class Cache<K, V>
+    {
+        // using immutable dictionary to avoid issues with race conditions.
+        // note that for caches data races are not that important.
+        ImmutableDictionary<K, V> dict = ImmutableDictionary<K, V>.Empty;
+        
+        // this function returns an object which is used for checking if the cache should be cleared.
+        readonly Func<object> clearTrigger;
+        
+        // last object returned by clearTrigger.
+        object currentObj;
+
+        ///<summary> The current number of cached elements. </summary> 
+        public int Count
+        {
+            get
+            {
+                CheckClear();
+                return dict.Count;
+            }
+        }
+
+        public Cache(Func<object> clearTrigger)
+        {
+            this.clearTrigger = clearTrigger;
+            currentObj = clearTrigger();
+        }
+        
+        void CheckClear()
+        {
+            var clearCheck = clearTrigger();
+            if (Equals(clearCheck, currentObj) == false)
+            {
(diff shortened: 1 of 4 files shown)


Shared/Cache.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary> tries getting a value for a key. </summary>
+        public bool TryGetValue(K key, out V value)
+        {
+            CheckClear();
+            return dict.TryGetValue(key, out value);
+        }
+
+        public V AddValue(K key, V value)
+        {
+            dict = dict.Add(key, value);
```
alnlarsen, line 53: Call ClearCheck here?

alnlarsen (changes requested):
Looks good. I didn't test it yet. Can we add a ClearCheck to the Add method?

alnlarsen (approved):
I reproduced the issue and verified the fix. Looks good.

AsgerIversen (changes requested):
@rmadsen-ks / @alnlarsen would it be possible to use our Memorizer class and avoid introducing another Caching utility class?

rmadsen-ks:
> @rmadsen-ks / @alnlarsen would it be possible to use our Memorizer class and avoid introducing another Caching utility class?

I think we could, but in that case we would have to introduce extra functionality to the memorizer class. We need something that can automatically invalidate everything in the memorizer and then clear that. This is a bit different from what it currently does.

I would say that introducing this new Cache class is less risky than trying to add the extra functionality to a class which is already quite complicated.

AsgerIversen (approved):

PR #979, opened 2023-01-24
967 Image install with empty specifier

### LINKED ISSUE #967: tap image install without an image specified causes it to install a 'null' image, deleting everything.
Run "tap image install" causes everything to be deleted.

This is an issue for me because the 'enter' button is right next to 'quotes' button which I use to write the image specifier.

**Actual behavior**

Everything, including OpenTAP gets deleted.

**Expected behavior**

An error specifying that an image without any packages cannot be installed.

Note that "image install" is non-browsable currently, so it is not likely to be used on accident.

rmadsen-ks:
- Added ways to catch if the image specifier is not present or if the image is resolved to no packages (this does not happen).

Close #967

### DIFF
--- a/Package/Image/ImageHelper.cs
+++ b/Package/Image/ImageHelper.cs
@@ -1,11 +1,6 @@
-﻿using OpenTap.Package;
-using System;
+﻿using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Runtime.CompilerServices;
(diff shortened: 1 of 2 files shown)


alnlarsen (dismissed):
Looks good. The failing pipeline issue has been solved and is included in the main branch

alnlarsen (approved):
Looks good!

Issue #931, opened 2022-11-30
Set MacroString.Context on deserialization


rmadsen-ks:
it seems that while MacroString.Context is set during object construction, it is not set when the object gets deserialized, forcing people to do something like this:

```cs
    MacroString filepath = new MacroString(){Text = "123"};
    public MacroString Filepath{
          get => filepath;
          set {
                  filepath = value;
                  filepath.Context = this;

          }
   }
```
instead, MacroString.Context should be set to the target object current MacroString.Context value.

alnlarsen:
I am unable to reproduce this, but I have created a unittest that verifies the correct behavior: #986 

Rolf, can you provide some more information about how to trigger this issue?

PR #986, opened 2023-01-25
Add unittest to verify MacroString context is correctly deserialized

alnlarsen:
I was trying to reproduce #931, but couldn't. @rmadsen-ks are you certain this is an issue?


rmadsen-ks (approved):
Looks good

PR #994, opened 2023-01-30
991 add suggested corrections to package actions

### LINKED ISSUE #920: Installing a package that does not exist as a release version is not straight forward
If the user correctly specify a package name that should be installed but does not exist in a release version then the process exit like if you defined a package that is misspelled.

Can we let the user know that the package does not exist in a release version and installing it require you to use the --version any (to get the latest version). 

Alternatively just install the latest version and let the user know that it happened as a release did not exist. 

(Runner package exists but not as a release):
MicrosoftTeams-image (5)

### LINKED ISSUE #991: Improve usability of the `tap package` group of CLI actions
We can improve the usability a lot in the `tap package` CLI action group when it comes to understanding user input. There are a few key problems:

1. Package names are case sensitive, but to my knowledge, there are no existing cases in the repository where ignoring casing would cause ambiguities
2. We have not enforced any standards for word separators in package names. Here are a few examples off the top of my head: `OpenTap.UnitTest`, `OpenTAP`, `REST-API`, `Package Diff`. Notice that in these four examples, the casing scheme, and the word separators are all different. And these are packages published by members of the OpenTAP team! I often find myself going to `https://packages.opentap.io` just to check what the correct casing is for some package.

If someone asks for `OpenTAP UnitTest`, for example, the user gets an error saying that this package doesn't exist and exits. We are already in contact with the repository at this point, and we could easily ask the user if they meant `OpenTap.UnitTest`.

I propose adding a fuzzy matcher to the package resolver used in our common CLI actions. If the user's input doesn't have an exact match, we should use a fuzzy matcher to match the input against all available packages, and provide the user with a list of suggested corrections which have a fuzzy score exceeding some threshold. If the action is running with `--non-interactive`, we should fall back to the current behavior.

I think this would be relatively simple to implement, and it would make the CLI much simpler to use.

alnlarsen:
This adds a UserInput with valid corrections if a user specifies a package name that we don't know about. This logic is skipped if the `--non-interactive` flag is used, so we won't see automatic corrections in e.g. pipelines and scripts

Examples:

```bash
$ tap package show opentap`
Package 'opentap' not found. Did you mean 'OpenTAP'?
0: 'Yes' (default)
1: 'No'
Please enter a number or name (Yes):
```

If there are multiple likely options:
```bash
$ tap package list 'editors'
Package 'editors' not found. Did you mean:
0: 'Editor' (default)
1: 'Editor X'
2: 'Editor CE'
3: 'Cancel'
Please enter a number or name (Editor):
```

If no version was specified (AnyRelease), and a single package was requested,
we now suggest retrying the install with a beta version instead:
```bash
$ tap package install Runner
Package 'Runner' has no compatible release version. Try a beta version instead?
0: 'Yes'
1: 'No' (default)
Please enter a number or name (No):
```

The fuzzy matcher is not *that* smart, but can be improved in a future release if there are glaring problems.

Unfortunately, this feature adds a noticeable delay, especially if a slow repository is configured. This is because the feature adds a round-trip to all repositories in all cases, except the case where the user correctly types the name of a package which is available locally.

Closes #991
Closes #920


Package/PackageActions/Install.cs
```diff
@@ -151,12 +152,39 @@ private int DoExecute(CancellationToken cancellationToken)
...
+                    // If the problem is exactly that we failed to resolve a compatible release version of a package,
+                    // ask the user to retry the resolution with a pre-release instead.
+                    // We could also ask for an RC / Any, but if there are no releases, then it is likely also the case that there are no RCs.
+                    // 'any' would be the most likely specifier to succeed, but it is probably a bit too extreme to suggest something so bleeding edge.
+                    // A beta version strikes a good middle ground. It is pretty likely to resolve, and probably won't be too unstable.
+                    if (NonInteractiveUserInputInterface.IsSet()) throw;
+                    if (false == (ex.Result is FailedImageResolution fir)) throw;
+                    if (fir.resolveProblems.Count != 1) throw;
+
+                    var problem = fir.resolveProblems[0];
```
StefanHolst, line 174: Should we make sure the problem is with the package the user is trying to install? We could get problems with other packages, right?

StefanHolst (changes requested):
Some comments. Let me know what you think

Package/PackageActions/List.cs
```diff
@@ -70,6 +70,10 @@ protected override int LockedExecute(CancellationToken cancellationToken)
             {
                 repositories = PackageManagerSettings.Current.GetEnabledRepositories(Repository);
             }
+            
+            var arr = new[] { Name };
+            AutoCorrectPackageNames.Correct(arr, repositories);
+            Name = arr[0];
```
StefanHolst, line 76: Seems a bit redundant?

Package/FuzzyMatcher.cs
```diff
@@ -0,0 +1,99 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace OpenTap.Package
+{
+    internal class FuzzyMatcher
```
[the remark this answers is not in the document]
alnlarsen: I wanted more fine grained control over our matching instead of using some stock matcher. I would also prefer to avoid adding dependencies since the algorithm is pretty simple anyway

StefanHolst (approved):
Looks fine

AsgerIversen (approved):
Very nice improvement. Thanks @alnlarsen

AsgerIversen (approved):

PR #1001, opened 2023-02-07
997 Fix issue preventing multi-editing of IInput types

### LINKED ISSUE #997: Multiple step edit: `If Verdict` step: Step setting `If` is not updated
**Describe the bug**

When editing multiple steps at the same time, the step setting `If` is not updated for all selected steps

 - OS: Windows 10
 - OpenTAP Version  9.19.4

**To Reproduce**

1. Add two If Verdict steps and a Dialog step
2. Select both If Verdict steps
3. Set `If` to the verdict from Dialog

**Actual behavior**

Only one If Verdict step has been updated 

**Expected behavior**

Both If Verdict steps should be updated

**Additional context**

Reproduced with both KS8400 (https://gitlab.it.keysight.com/tap/tap/issues/6627) and Editor X (https://gitlab.it.keysight.com/tap/editors/editor-x/issues/563)

Also reproduced for Repeat step: the step setting `Verdict Of`

alnlarsen:
* Provide object context to ObjectCloner in MergedValueAnnotation implementation
* Improve InputStringConvertProvider implementation to accept IInput instances as context in addition to teststeps
* Add unittest to verify the fix

Closes #997


Engine/IStringConvertProvider.cs
```diff
@@ -832,7 +832,7 @@ public object FromString(string stringdata, ITypeData type, object contextObject
                 if (!type.DescendsTo(typeof(IInput))) return null;
                 
                 var s = stringdata.Split(':');
-                var step = contextObject as ITestStep;
+                var step = contextObject as ITestStep ?? (contextObject as IInput)?.Step;
```
rmadsen-ks, line 835:
```suggestion
                // under normal circumstances the contextObject is a test step
                // but during cloning it may also be another Input<T> base instance.
                var step = contextObject as ITestStep ?? (contextObject as IInput)?.Step;
```

rmadsen-ks: Read my other suggestion - maybe we don't need this?

rmadsen-ks (changes requested):
Had an idea for a more generic adjustment for when ObjectCloner is used.

Engine/Annotations/Annotation.cs
```diff
@@ -760,22 +760,23 @@ public object Value
...
+
                 // same as for parameters.
                 var first = merged[0];
                 first.Get<IObjectValueAnnotation>().Value = value;
                 var cloner = new ObjectCloner(value);
                 foreach (var m in merged.Skip(1))
                 {
                     var memberType = m.Get<IMemberAnnotation>()?.ReflectionInfo ?? TypeData.GetTypeData(value);
-                    if(cloner.TryClone(null, memberType, false, out var val ))
+                    if (cloner.TryClone(value, memberType, false, out var val))
```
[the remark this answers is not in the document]
alnlarsen: I don't really understand how we can get that from `m`. In this case, `m` only has an uninitialized  MemberValueAnnotation.

Did you want to try to ad-hoc find the teststep from `value`?

Engine/IStringConvertProvider.cs
alnlarsen: I'm a bit on the fence with this. 

It would of course be more sane if the context was always a teststep, but it would also make sense if `InputStringConvertProvider` understood an IInput as context.

alnlarsen:
I see, thanks for your feedback, Rolf! I think this more generic way is better.

rmadsen-ks (changes requested):
I tested this with KS8400 and it does not seem to work. I just installed the OpenTAP package, but multi-edit for Input still does not seem to work.

Did you test this way?

alnlarsen:
No, I didn't test using the editor. I am looking into it!

Engine/Annotations/Annotation.cs
```diff
@@ -746,7 +747,7 @@ public object Value
                             if (thisVal is IEnumerable ie2)
                             {
                                 if (ie2.Cast<object>().SequenceEqual(ie1.Cast<object>()))
-                                    return selectedValue;
+                                    continue;
```
alnlarsen, line 750: Same as above

Engine/Annotations/Annotation.cs
```diff
@@ -737,7 +737,8 @@ public object Value
                     {
                         var x = merged[i];
                         var thisVal = x.Get<IObjectValueAnnotation>().Value;
-                        if (thisVal == selectedValue) return selectedValue;
+                        if (thisVal == selectedValue)
+                            continue;
```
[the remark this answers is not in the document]
rmadsen-ks: I think this is probably only triggered during parameterization, in which case the value is generally always the same, so we are just assuming that if the second value is the same as the first then they will all be the same.

If there is a bug with that I am not able to reproduce it.

Is there any reason why you changed this?

rmadsen-ks (changes requested):
Had a few comments.

alnlarsen:
I have changed the suspected loop bug back, and added some comments. Please take another look, Rolf!

rmadsen-ks (approved):
This seems to work. Let's get it in :)

rmadsen-ks:
I had hoped we did not need to deal with MergedValueAnnotation explicitly, but maybe there is no way around it.

PR #1002, opened 2023-02-08
Add errors out parameter in `ComponentSettings.SetCurrent` overload

### LINKED ISSUE #1000: ComponentSettings.SetCurrent does not return errors
**Describe the bug**

OpenTAP no longer throws exceptions in ComponentSettings.SetCurrent. We realized this when the runners didn't return any serialization errors in the response of `UploadComponentSettings()`. 

These errors typically happen if the packages on which the TapSettings/TapPlan file depends are not installed in that session/runner and are sent back as a response from methods such as `LoadTestPlanFromRepository()`.

alnlarsen:
This adds a new overload for `ComponentSettings.SetCurrent` which has an out parameter containing XML errors

Closes #1000


Engine/ComponentSettingsContext.cs
```diff
@@ -164,8 +164,7 @@ public ComponentSettings GetCurrent(Type settingsType)
             return objectCache.Invoke(settingsType);
         }
 
-
-        public void SetCurrent(Stream xmlFileStream)
+        public XmlError[] SetCurrent(Stream xmlFileStream)
```
Denrask, line 167: This is not considered breaking, because of the wrapper?

Denrask (approved):
Looks good. Just want to verify that it is not breaking on an API level

Engine/ComponentSettingsContext.cs
alnlarsen: Yes, `ComponentSettingsContext` is an internal class

AsgerIversen (changes requested):
How about we just make the new ComponentSettings.LoadErrors property public? That would also make load errors available to callers of Load, Reload and Invalidate.

I'm not sure I fully agree that the existing naming is bad. This is a kind of singleton that you can access through the static ComponentSettings.Current property, or the GetCurrent function. You can then change its value with the SetCurrent function, I think that makes sense.

Engine/ComponentSettings.cs
```diff
@@ -627,17 +627,19 @@ public virtual void Initialize()
...
         /// <summary>
         /// Sets current settings for a component setting based on a stream of the file contents of a ComponentSettings XML file.
         /// </summary>
-        /// <param name="xmlFileStream"></param>
+        /// <exception cref="InvalidDataException">If the input stream is not valid XML</exception>
+        /// <param name="xmlFileStream">The component settings stream to be set</param>
+        /// <param name="errors">Any XML errors that occurred during deserialization</param>
         /// <returns></returns>
-        public static XmlError[] SetFromXmlStream(Stream xmlFileStream) => context.SetCurrent(xmlFileStream);
+        public static void SetCurrent(Stream xmlFileStream, out XmlError[] errors) => errors = context.SetCurrent(xmlFileStream);
```
AsgerIversen, line 642: Should we switch to IEnumerable<XmlError> instead?

Also, how complicated would it be to have this overload never throw?

AsgerIversen (changes requested):
As we just discussed, maybe there is a few improvements we could make to this new API

Engine/ComponentSettings.cs
alnlarsen: It is easy to return an XmlError instead of throwing `InvalidDataException`, but it seems weird to wrap a generic exception in an XmlError

AsgerIversen (approved):

PR #1004, opened 2023-02-08
Fix unittest pipeline stage

### LINKED ISSUE #1003: Engine UnitTests diabled in builds
At some point the unit tests has stop being run, although the job still succeeds.

This is a critical issue that should be fixed ASAP.

alnlarsen:
This is supposed to fix three problems:

1. When tests fail to run, the pipeline succeeds instead of failing
2. Our unittests fail to run
3. One of our unittests started failing because an online resource was moved

Closes #1003


StefanHolst (approved):
Looks fine to me

PR #1005, opened 2023-02-08
980 add Required option to CommandLineArgumentAttribute

### LINKED ISSUE #980: Improve CLI `CommandLineArgument` validation
We have a lot of instances of CLI actions with some code like:

```csharp
if (arg == null) {
  throw new ArgumentException("'arg' not specified.", nameof(arg));
}
```

We already have a `Required` property for `UnnamedCommandLineArgument`. I think it makes sense to add it for `CommandLineArgumentAttribute`.

It would also be nice if we had a pattern similar to teststep validation errors for CLI actions. We could make a `CLIAction` implementation which implements `IValidatingObject` so OpenTAP can validate the input before starting the CLI action.

alnlarsen:
This adds the Required option to CommandLineArgumentAttribute

This allows authors of CLI actions to indicate that a named argument is required. If the required argument is not provided, the CLI action executor will give an appropriate error message, and fail with the help text for the CLI action.

This should save CLI action authors to avoid a lot of boiler plate that we ourselves duplicate in all of our plugins:

Code like this should no longer be necessary:
```csharp
if (string.IsNullOrWhiteSpace(SomeProp))
  throw new Exception($"{nameof(SomeProp)} is required");

```

Closes #980

### DIFF
--- a/Engine/Cli/CommandLineArgumentAttribute.cs
+++ b/Engine/Cli/CommandLineArgumentAttribute.cs
@@ -36,6 +36,11 @@ namespace OpenTap.Cli
         /// </summary>
         [Obsolete("Please use the 'Browsable' attribute.")]
         public bool Visible { get; set; } = true;
+        
+        /// <summary>
+        /// Indicates whether this argument is required or optional.
+        /// </summary>
+        public bool Required { get; set; }
 
         /// <summary>
         /// Primary constructor.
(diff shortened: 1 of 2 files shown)


alnlarsen:
This is pretty low value. Null checks is the smallest part of input validation, and forcing validation into a validation error system doesn't really help anyone.

PR #1007, opened 2023-02-09
1006 Dynamic Serializer Dependency: Fix + Test

### LINKED ISSUE #1006: Add a way for a serializer plugin to dynamically specify if it is needed in deserialization
This is needed in order to be able to have a serializer which only dynamically adds information to the test plan XML and hence is not always required in order to deserialize the XML.

This is need to avoid some package dependencies from spreading into test plans which don't need them.

Otherwise, it is not possible to create some types of serializer plugins without the package dependencies spreading like a virus - if you install a package with the dependency you will get the serializer installed, which means all your test plans will now also have the dependency added.

rmadsen-ks:
Added a way for a serializer plugin to specify it is needed for deserialization and hence become added as a package dependency if a 3rd plugin (aside from OpenTAP) implements it.

Added a unit test to verify the behavior by checking TapSerializer.GetUsedTypes.

Close #1006

### DIFF
--- a/Engine.UnitTests/SerializerTests.cs
+++ b/Engine.UnitTests/SerializerTests.cs
@@ -288,10 +288,15 @@ namespace OpenTap.Engine.UnitTests
 
             public override double Order => 1000;
             HashSet<XElement> visitedNodes = new HashSet<XElement>();
+            static XName testPlanName = "TestPlan";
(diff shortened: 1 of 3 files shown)


alnlarsen:
It looks like some tests are failing, @rmadsen-ks

alnlarsen (approved):
Looks fine to me.

PR #1013, opened 2023-02-10
Fix newtonsoft not shipped on Linux

alnlarsen:


AsgerIversen (approved):
Let's get it in. Then we'll improve the check when files gets removed from the package (so it fails the pipeline) in a separate PR.

PR #1019, opened 2023-02-15
1018 corrupted input fix

### LINKED ISSUE #1018: Moving steps may cause Inputs to become corrupted
1. Create a test plan with two If Verdict
2. The last step: Select the first step as input.
3. Swap the steps
4. The (now) first step, select None in the drop down.
5. swap the steps again
6. The seconds step: Note that the input is blank (not none)

This is caused because the input has two properties: property and step. 
When a step is deleted, any inputs referring to it are set to None.
When moving a step, it looses the context and the step is set to null, but then when it is inserted at the new position we use the step id to find to right step reference.
However, when explicitly setting the input, the property and step are set to null, but the Step ID is still set. When we then move the step the Step is set to something, while the property is still null, causing a corrupted Input configuration.

image

rmadsen-ks:
See the commit message for details.

Close #1018

### DIFF
--- a/Engine/Input.cs
+++ b/Engine/Input.cs
@@ -126,33 +126,37 @@ namespace OpenTap
         Guid stepId;
         void ChildTestSteps_CollectionChanged(object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
         {
-            if(Step != null && e.OldItems != null && e.Action == System.Collections.Specialized.NotifyCollectionChangedAction.Remove)
+            if (Step != null && e.OldItems != null && e.Action == System.Collections.Specialized.NotifyCollectionChangedAction.Remove)
             {
                 if (e.OldItems.Contains(step) || Step.GetParents().Any(e.OldItems.Contains))
                 {
                     stepId = Step.Id;
                     Step = null;
                 }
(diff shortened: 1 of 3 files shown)


Engine/Input.cs
```diff
@@ -142,7 +144,15 @@ void ChildTestSteps_CollectionChanged(object sender, System.Collections.Speciali
...
-                    if(step.Id == stepId)
+                    // Update the Step value when the step has moved inside the test plan.
+                    // We do this because we want to set the Input to null if the test plan changes in a way
+                    //     where the target step disappears.
+                    
+                    // Avoid setting the Step if the property name is null - this usually means that 
+                    //     the property has been set to null, but the stepId still has a remaining value.
+                    //     this should be ok to ignore.
+                    
+                    if(step.Id == stepId && Property != null)
```
alnlarsen, line 155: If Property is null, why did we enter this loop in the first place?

alnlarsen (changes requested):
The fix works, but I think we can move the property null check out of the loop

alnlarsen (approved):
Applied the small change from my review as well. Let's get this into 9.20.0

PR #1030, opened 2023-02-22
910 StringValueAnnotation for Enabled<T>

### LINKED ISSUE #910: Cannot import external parameter of type Enabled<T> through csv file
**Describe the bug**

Test Step has a setting of type Enabled<int> which needs to be parameterized.
When we try to parameterize on Test Plan and import the parameters through csv file, it fails.

 - OS: [e.g. Windows 10]
 - OpenTAP Version [e.g. 9.18.4]

**To Reproduce**

1. Create a TestStep with an Enabled<int> type setting.
2. Parameterize the setting on the TestPlan
3. Try to import values/export values of this parameter
4. Warning shows up saying "Unable to export TestPlan / Setting since it cannot be converted to/from a string value

**Actual behavior**
1. Warning shows up saying "Unable to import TestPlan / Setting since it cannot be converted to/from a string value and parameter is not imported

**Expected behavior**
1. Allow importing/exporting parameterizing the Enabled<T> values through a csv file.

image

rmadsen-ks:
Fixed the issue by adding a string annotation for Enabled<T>.

Added a unit test toverify the fix

Close #910


Engine/Annotations/Annotation.cs
```diff
@@ -2253,6 +2253,20 @@ public override string ToString()
...
+            readonly AnnotationCollection a;
+            public EnabledStringAnnotation(AnnotationCollection a)
+            {
+                this.a = a;
+            }
+
+            public string Value
+            {
+                get => StringConvertProvider.GetString(a.Get<IObjectValueAnnotation>().Value);
+                set => a.Get<IObjectValueAnnotation>().Value = StringConvertProvider.FromString(value, a.Get<IReflectionAnnotation>()?.ReflectionInfo, a?.ParentAnnotation?.Source ?? a.Source);
```
alnlarsen, line 2267: Could we just use EnabledStringConvertProvider directly here instead of consulting the stack of providers? Or do we think it's possible that a plugin StringConvertProvider is needed?

alnlarsen (commented):
I think this looks good, but I have a question

kaushik-keysight:
I don't think this is working. I tried exporting a setting with the type Enabled<String> and tried to change the value in the csv file and reimport it. 
It does not update the value. It still shows the old value. I am not sure if I am missing something.

rmadsen-ks:
> I don't think this is working. I tried exporting a setting with the type Enabled and tried to change the value in the csv file and reimport it. It does not update the value. It still shows the old value. I am not sure if I am missing something.

@kaushik-keysight, for me it seems to work perfectly. I tested by parameterizing the Regular Expression of Run Program step. Can you try this?

kaushik-keysight:
A8373BE3-D127-4282-900F-B77121191076

Hi @rmadsen-ks , 
I tried with the run program step on this branch 910-StringValueAnnotationForEnabled
It still does not import the values. The message shows as if it imported the values successfully, but the values remain unchanged. 
Am I doing something wrong in testing out the branch? I checked out this branch and built the solution, went to the bin/Debug folder and installed the Editor there (using tap.exe package install "Developer's System" ), and then tried it out. 
Is this the right way to try it out?

Also will the enabled<T> import work only for Strings and not other data types?

PR #1038, opened 2023-02-24
1037 Type Menu Annotations: Fix

### LINKED ISSUE #1037: Type menu annotations
Currently we have member menu annotations. These are used for adding additional functionality for a setting, usually in the form of a context menu item.

For example, we have 'assign output' as a menu annotation.  These are currently widely supported. 

It would be nice to also have support for 'type' menu annotations. Instead of the functionality being available for a given setting, it would be the object itself that has it. Currently, this could be useful for:

- Exporting/Importing test plan settings as a CSV file (current feature in KS8400, it could be improved with this feature).
- Dynamic test step settings. Currently this has to be implemented by the right click menu of a property. Additing a context menu for the step itself or at least locating the functionality there would make more sense.

rmadsen-ks:
Added solution for type menu annotations. Currently this will only work for test plans and test steps.

There is no implementation for this in OpenTAP itself, but there is a unit test that shows how it works for Test Plans.

Close #1037

### DIFF
--- a/Engine/MenuAnnotation.cs
+++ b/Engine/MenuAnnotation.cs
@@ -11,12 +11,20 @@ namespace OpenTap
     /// </summary>
     public sealed class MenuAnnotation : IOwnedAnnotation
     {
-        /// <summary> Creates a new instance. </summary>
+        /// <summary> Creates a 'member' menu annotation. </summary>
         /// <param name="member"> Which member of an object should the menu be generated for</param>
-        internal MenuAnnotation(IMemberData member) => this.member = member; 
+        internal MenuAnnotation(IMemberData member) => this.member = member;
+        
+        /// <summary> Create a 'type' menu annotation. </summary>
+        /// <param name="type"> The type for which the menu should be generated. </param>
+        internal MenuAnnotation(ITypeData type) => this.type = type;
 
-        /// <summary> The model used to construct menu items. </summary>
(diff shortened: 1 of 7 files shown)


Engine/MenuAnnotation.cs
```diff
@@ -11,12 +11,25 @@ namespace OpenTap
     /// </summary>
     public sealed class MenuAnnotation : IOwnedAnnotation
     {
-        /// <summary> Creates a new instance. </summary>
-        /// <param name="member"> Which member of an object should the menu be generated for</param>
-        internal MenuAnnotation(IMemberData member) => this.member = member; 
+        /// <summary> Creates a 'member' menu annotation. </summary>
+        /// <param name="member"> Which member of an object should the menu be generated for.</param>
+        /// <param name="type"> Set this if menu items for the type should be generated. </param>
+        internal MenuAnnotation(IMemberData member, ITypeData type)
```
StefanHolst, line 17: This looks breaking!

StefanHolst (changes requested):
Looks fine, just a minor breaking change

Engine/MenuAnnotation.cs
rmadsen-ks: This is an internal API, so in the case we can change the argument types.

StefanHolst (approved):
Looks good

PR #1055, opened 2023-03-16
Add userinput when wiping image installations

alnlarsen:


Package/PackageActions/ImageInstall.cs
```diff
@@ -131,5 +144,22 @@ protected override int LockedExecute(CancellationToken cancellationToken)
...
+            No,
+            Yes,
+        }
+
+        [Display("Completely wipe installation?")]
+        class WipeInstallationQuestion
+        {
+            [Browsable(true)]
+            [Layout(LayoutMode.FullRow)]
+            public string Message { get; } = "You are about to completely wipe your installation. Do you wish to Contineu?";
```
AsgerIversen, line 159: Typo..

Package/PackageActions/ImageInstall.cs
```diff
@@ -106,7 +106,20 @@ protected override int LockedExecute(CancellationToken cancellationToken)
                 log.Debug(sw, "Image resolution done");
 
                 if (image.Packages.Count == 0)
-                    throw new InvalidOperationException("Resolved image is empty.");
+                {
+                    // If we are running interactively, we should prompt the user and warn them that they are
+                    // about to completely wipe their installation.
+                    if (NonInteractiveUserInputInterface.IsSet() == false)
```
AsgerIversen, line 112: Is this if really necessary? I would think UserInput.Request would just result in the default answer when using NonInteractiveUserInputInterface, so we should still be able to call that.

... or is this because the default answer is not to continue? I think it is probably best that you get the default, if you specify --non-interactive.

AsgerIversen (changes requested):
Good solution to the smart installer issue. Just some details...

Package/PackageActions/ImageInstall.cs
alnlarsen: I agree. I simplified it

AsgerIversen (approved):

PR #1068, opened 2023-03-24
Create option for packages to not fail when no exe was found

### LINKED ISSUE #1063: PackageActionExtension feature: Pass when ExeFile does not exist
Currently, we support specifying what exit codes are acceptable, but there is no way to say that an action should pass if the ExeFile does not exist. 

This is needed for the SDK plugin because it needs to install `dotnet new` templates, but we don't want the package installation to fail when `dotnet sdk` is not installed, or if `dotnet` is not in PATH.

frederikja163:
closes #1063

### DIFF
--- a/Package/PluginInstaller.cs
+++ b/Package/PluginInstaller.cs
@@ -13,6 +13,7 @@ using System.IO.Compression;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Runtime.InteropServices;
+using System.ComponentModel;
 
 namespace OpenTap.Package
 {
@@ -130,7 +131,7 @@ namespace OpenTap.Package
                 // If OPENTAP_COLOR is set, the escape symbols for colors in the child process will break the parsing of the forwarded logs.
                 // Ensure color is never set in the child process. Colors will still be set in the parent process.
                 pi.Environment["OPENTAP_COLOR"] = "never";
-
+                
                 try
                 {
                     Process p;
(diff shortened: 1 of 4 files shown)


Package/PluginInstaller.cs
```diff
@@ -130,23 +131,59 @@ IEnumerable<string> possiblePaths(string file)
...
-                        p = Process.Start(pi);
+                        try
+                        {
+                            p = Process.Start(pi);
+                        }
+                        catch (Exception e)
+                        {
+                            // Win32Exception is when EXE was not found.
+                            // InvalidOperationException is when no file name was specified.
+                            // ArgumentException is when the file path is not valid.
```
alnlarsen, line 150: I think we should explicitly handle these two cases before we try to start the program, and just catch the `Win32Exception` for when the executable was not in PATH

alnlarsen: In fact, it might be better if we handle those two cases at `Package Create` time, rather than at `package install` time.

Package/PluginInstaller.cs
```diff
@@ -192,6 +193,10 @@ IEnumerable<string> possiblePaths(string file)
                 }
                 catch (Exception e)
                 {
+                    if (step.FailOnNoExe && e is Win32Exception)
```
alnlarsen, line 196: It looks like you flipped the condition, and the action will still fail even if the condition is met. 

We probably want something like

```csharp
catch (Win32Exception) when (step.FailOnNoExe == false)
{
    log.Warning($"'{step.ExeFile}' not found, skipping action.");
}
```

rmadsen-ks (approved):
Looks good

alnlarsen (approved):
Excellent

PR #1084, opened 2023-04-12
Download system-wide packages in main process and install from local files

### LINKED ISSUE #1043: Resolver error: can't install system-wide package from local .TapPackage
I reproduced this in 9.20 RC and 9.19.5

I tried installing a system-wide package from a local .TapPackage file, but the resolver failed to find it (after correctly resolving the image). 

There are two workarounds:
1. Run the CLI action as admin
2. Copy the .TapPackage into the package cache

It seems to be related to elevation.

alnlarsen:
This fixes an issue causing system-wide packages to fail to install from local files if they are not available in the configured repositories.

Closes #1043

### DIFF
--- a/Package/PackageActions/Install.cs
+++ b/Package/PackageActions/Install.cs
@@ -323,15 +323,31 @@ namespace OpenTap.Package
                         }
                     }
                 }
+                
+                // Download the packages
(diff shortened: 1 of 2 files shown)


StefanHolst (approved):
Looks good to me

PR #1098, opened 2023-04-21
1097 Allow more flexible ordering of StringConverters

### LINKED ISSUE #1097: Allow more flexibility in ordering of IStringConvertProviders
Currently we have an IConvertible StringConvertProvider. This is normally fine, but since Pythonnet/PyObject has a pretty broken IConvertible implementation, it makes the object impossible to clone with the current system.

If we had a way to specify the order, then this could be done by checking PyObjectconvertProvider before ConvirtibleConvertProvider.
This could be done simply with an 'IOrderedStringConvertProviderPlugin that contains a single double property.

But maybe we could think of some more generic way of doing ordering in our plugins? something like double Order(ITypeData otherType). to specify where something comes relative to another plugin. or with an attribube like: [Before(Type)] .

This is needed in order to fix https://github.com/opentap/OpenTap.Python/issues/108

rmadsen-ks:
- Added BeforeAttribute, which adds an ordering constraint to plugin types (currently only supported by string convert providers).
- Added ordering to plugin resolver for IStringConvertProvider.
- Added unit test that verifies the new behavior.

Close #1097

### DIFF
--- a/Engine.UnitTests/BeforeAttributeTest.cs
+++ b/Engine.UnitTests/BeforeAttributeTest.cs
@@ -12,7 +12,7 @@ namespace OpenTap.Engine.UnitTests
     {
         public class StringConvertProviderA : IStringConvertProvider
         {
-            public static List<Type> UseOrder = new List<Type>();
+            public static readonly List<Type> UseOrder = new List<Type>();
             public string GetString(object value, CultureInfo culture)
             {
                 if (UseOrder.Contains(GetType()) == false)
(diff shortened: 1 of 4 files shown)


Engine/BeforeAttribute.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+namespace OpenTap
+{
+    /// <summary>
+    /// Ordering constraint for plugin types. This should be used 'before' something else.
+    /// This is currently only support by implementations of IStringConvertProvider.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
alnlarsen, line 8: Can we add `AllowMultiple`?

Engine/IStringConvertProvider.cs
```diff
@@ -66,20 +66,33 @@ public static T[] GetPlugins()
...
                                     try
                                     {
                                         var newthing = (T)type.CreateInstance();
                                         saveTypes.Add(type, newthing);
-                                        builder.Add(newthing);
+                                        
+                                        // if BeforeAttribute is used, insert somewhere before the marked type.
+                                        // otherwise insert at the end of the list.
+                                        int addIndex = builder.Count;
+                                        if (type.GetAttribute<BeforeAttribute>() is BeforeAttribute before)
```
alnlarsen, line 80: Can we add support for multiple `BeforeAttribute` instances?

alnlarsen (changes requested):
I'm a big fan of this feature. But I would like support for multiple `BeforeAttribute` instances.

Also, there is a similar implementation used in our `Dynamic Package XML` code. The internal attribute `DependsOnAttribute`. Can we update that implementation to use `BeforeAttribute` instead?

Engine.UnitTests/BeforeAttributeTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            StringConvertProvider.GetString(new StringConvertProviderD());
+            int ia, ib, ic, id;
+            ia = StringConvertProviderA.UseOrder.IndexOf(typeof(StringConvertProviderA));
+            ib = StringConvertProviderA.UseOrder.IndexOf(typeof(StringConvertProviderB));
+            ic = StringConvertProviderA.UseOrder.IndexOf(typeof(StringConvertProviderC));
+            id = StringConvertProviderA.UseOrder.IndexOf(typeof(StringConvertProviderD));
+            Assert.IsTrue(ib < ic);
+            Assert.IsTrue(ib < id);
+            Assert.IsTrue(ic < id);
+            Assert.IsTrue(ia < ic);
```
alnlarsen, line 60: How does this make sense?

`ib < ic && ib < id` makes sense, because IStringConvertProviderB has Before(C,D)
`ic < id` does not make sense, because ID has Before(C)

It seems to me the test should be like this:
`ib < ic`
`ib < id`
`ic < ia`
`id < ic`

alnlarsen (approved):
Looks good

PR #1109, opened 2023-04-28
1093 cli name clash

### LINKED ISSUE #1093: Handle name-clashes in CLI actions
It has always been possible to define CLI actions with the same name / group. OpenTAP has always handled this case by listing all of the options, and executing whichever one was loaded first. See the following help output in a build wheere I have defined multiple `sdk gitversion` actions:

```
❯ ./tap -h
"tap -h" is not a recognized command.
OpenTAP Command Line Interface (9.21.0-alpha.12+ccbcf336.1082-sweep-changing-child-reso)
Usage: "tap <command> [<subcommand(s)>] [<args>]"

Valid commands are:

package
......
sdk                       
   gitversion             Calculate the semantic version number for a specific git commit.
   gitversion             Calculate the semantic version number for a specific git commit.
   gitversion             Example 3
   gitversion             Example 2
.......
Run "tap <command> [<subcommand(s)>] -h" to get additional help for a specific command.
```

OpenTAP should provide a deterministic way of using each of these actions.

frederikja163:
closes #1093

### DIFF
--- a/Engine/Cli/CliActionExecutor.cs
+++ b/Engine/Cli/CliActionExecutor.cs
@@ -65,23 +65,21 @@ namespace OpenTap.Cli
             }
         }
 
-        public CliActionTree GetSubCommand(string[] args)
+        public CliActionTree GetSubCommand(IEnumerable<string> args)
         {
-            if (args.Length == 0)
-                return null;
+            if (!args.Any())
+                return this;
 
             foreach (var item in SubCommands)
             {
-                if (item.Name == args[0])
+                if (item.Name == args.First())
                 {
-                    if (args.Length == 1 || item.SubCommands.Any() == false)
-                       return item;
-                    var subCmd = item.GetSubCommand(args.Skip(1).ToArray());
-                    return subCmd ?? item;
+                    var subCmd = item.GetSubCommand(args.Skip(1));
+                    return subCmd;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Engine/Cli/CliActionExecutor.cs
```diff
@@ -200,9 +228,25 @@ public static int Execute(params string[] args)
...
-            var selectedcmd = actionTree.GetSubCommand(args);
-            if (selectedcmd?.Type != null && selectedcmd?.SubCommands.Any() != true)
-                SelectedAction = selectedcmd.Type;
+            var commands = actionTree.GetSubCommand(args).ToList();
+            if (commands.Count > 1)
+            {
+                if (args.Contains("--non-interactive") || Environment.GetEnvironmentVariable("OPENTAP_NON_INTERACTIVE") != null)
+                {
+                    throw new Exception("Multiple commands with the same name found. Run without --non-interactive to select a command, or without env variable OPENTAP_NON_INTERACTIVE set.");
+                }
```
alnlarsen, line 237: This changes existing behavior. Before this change, this would have executed the first command. It is probably okay to log a debug message, but I don't think we should throw

alnlarsen (changes requested):
See remark. I'm not sure if we want this or not. Let's discuss with Rolf.

PR #1114, opened 2023-05-02
CliUserInputInterface: Read from stdin pipe if no terminal is connected

### LINKED ISSUE #1092: CliUserInputImplementation: Handle case when no TTY is connected
When running in an environment with no stdin connected, the current implementation just hangs. We should detect if there is no stdin, and read from the stdin pipe.

I saw this exception in a GitHub pipeline: `InvalidOperationException: Cannot read keys when either application does not have a console or when console input has been redirected. Try Console.Read.
`

alnlarsen:
This adds a separate reader mode for the CLI user input handler. This reader mode is automatically used when `Console.IsInputRedirected` is set. This will be set whenever the the stdin of the process is not attached to a terminal. 

| Input Mode | Description |
| --- | --- |
| Piped | Stdin is read line by line. UserInput request `n` gets the value of line `n`. If EOF is reached, and there are more UserInput requests, an EOF exception is thrown. |
| Stdin | (unchanged) Stdin is continously read line by line by a background thread. When a user input arrives, it blocks until it times out, or a line is produced by the worker thread. |

Closes #1092

### DIFF
--- a/Engine/UserInput.cs
+++ b/Engine/UserInput.cs
@@ -122,63 +122,129 @@ namespace OpenTap
     /// <summary> Standard implementation of UserInputInterface for Command Line interfaces</summary>
     public class CliUserInputInterface : IUserInputInterface
     {
+        /// <summary>
+        /// Thrown when userinput is requested when reading input from stdin and EOF is reached
+        /// </summary>
+        internal class EOFException : Exception
+        {
+            /// <summary>
+            /// Initializes a new instance of the <see cref="EOFException"/> class with the specified message.
+            /// </summary>
+            /// <param name="message"></param>
+            public EOFException(string message) : base(message)
+            {
+            }
+        }
+
+        private bool IsPipeInput { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CliUserInputInterface"/> class in interactive mode. 
+        /// </summary>
+        public CliUserInputInterface()
+        {
+            IsPipeInput = Console.IsInputRedirected;
+        }
+
         readonly Mutex userInputMutex = new Mutex();
         readonly object readerLock = new object();
 
-        void IUserInputInterface.RequestUserInput(object dataObject, TimeSpan timeout, bool modal)
+        private TapThread StartKeyboardReader()
         {
-            if(readerThread == null)
+            return TapThread.Start(() =>
             {
-                lock (readerLock)
+                try
                 {
-                    if (readerThread == null)
-                    {
-                        readerThread = TapThread.Start(() =>
-                        {
-                            try
-                            {
-                                var sb = new StringBuilder();
-                                
-                                while (true)
-                                {
-                                    
-                                    var chr = Console.ReadKey(true);
-                                    if (chr.KeyChar == 0)
-                                        continue;
-                                    if (chr.Key != ConsoleKey.Enter && chr.Key != ConsoleKey.Backspace)
-                                        Console.Write(IsSecure ? '*' : chr.KeyChar);
-                                    
-                                    if (chr.Key == ConsoleKey.Enter)
-                                    {
-                                        lines.Add(sb.ToString());
-                                        sb.Clear();
-                                        Console.WriteLine();
-                                    }
+                    var sb = new StringBuilder();
 
-                                    if (chr.Key == ConsoleKey.Backspace)
-                                    {
-                                        if(sb.Length > 0){
-                                            sb.Remove(sb.Length - 1, 1);
-                                            // delete current char and move back.
-                                            Console.Write("\b \b");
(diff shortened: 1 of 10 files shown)


Engine/UserInput.cs
```diff
@@ -127,7 +127,7 @@ public class CliUserInputInterface : IUserInputInterface
 
         void IUserInputInterface.RequestUserInput(object dataObject, TimeSpan timeout, bool modal)
         {
-            if(readerThread == null)
+            if(readerThread == null && Console.IsInputRedirected == false)
```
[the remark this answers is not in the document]
rmadsen-ks: Also, lets say i want to pipe some input into the program, there may be ways to do this and I think we should support that or it would be a somewhat difficult to explain behavior.

Alternatively, we could have an extra command line argument that can be used for when it is wanted to pipe input into the application. Just like if you do `git diff`, it behaves differently from when you do `git diff|grep something`.

rmadsen-ks (changes requested):

Engine/UserInput.cs
```diff
@@ -317,7 +316,12 @@ bool isBrowsable(IMemberData m)
                     else if(string.IsNullOrEmpty(str.Value) == false)
                         Console.Write($"({str.Value}): ");
                     else Console.WriteLine(":");
-                    
+
+                    if (Console.IsInputRedirected)
+                    {
+                        log.Warning("\nStandard Input is redirected. Default option chosen.\n");
```
rmadsen-ks, line 322: I don't think we should emit a warning. Whatabout writing something a bit more useful as an Info message.

e.g "Selecting '{value}' for '{property}'?

rmadsen-ks (changes requested):

Engine/Cli/RunCliAction.cs
```diff
@@ -148,9 +148,13 @@ public int Execute(CancellationToken cancellationToken)
             }
 
 
-            if (!NonInteractive && UserInput.Interface == null)
+            if (UserInput.Interface == null)
```
rmadsen-ks, line 151: I like the idea of the 4 modes in your description. Why don't we make those 4 modes arguments for CliUserInputInterface to make it explicit when the class is created and move that state more up?

rmadsen-ks (changes requested):

alnlarsen:
I have reverted the whole 'non-interactive cli reader' because I don't think the concept makes a lot of sense. I can't imagine this being used for CLI actions with more than a single user input anyway. In all other cases, 'non-interactive' makes more sense.

Engine/UserInput.cs
```diff
@@ -318,16 +373,15 @@ bool isBrowsable(IMemberData m)
                         Console.Write($"({str.Value}): ");
                     else Console.WriteLine(":");
                     
-                    
                     if (secure)
                     {
-                        var read2 = (awaitReadLine(timeoutAt, true) ?? "").Trim();
+                        var read2 = (awaitReadLine(timeoutAt, true)).Trim();
```
rmadsen-ks, line 378: looks like awaitReadLine could previously return null. This is no longer the case?

Engine/UserInput.cs
```diff
@@ -398,11 +467,14 @@ string awaitReadLine(DateTime timeOut, bool secure)
             }
         }
 
+        // These must both be static, because concurrent reader threads will wreak havoc by simulatenously reading from stdin
```
rmadsen-ks, line 470: They must be static but arent?

rmadsen-ks (changes requested):
Can we add a test to the "Test test plans" which verifies the new piped behavior?

Engine/UserInput.cs
alnlarsen: I will remove the outdated comment :)

Engine/UserInput.cs
alnlarsen: No, `awaitReadLine` was used in two places, and both places used null propagation, so I moved the null propagation into `awaitReadLine` instead

rmadsen-ks:
I think we need to investigate the surprising behavior with BOM and github pipelines before implementing code specifically to handle that.

alnlarsen:
@rmadsen-ks I have removed the special handling for utf-8 bom. I also discovered how to reproduce the problem locally. 

This is an old bug in .NET Framework which will be solved by moving to .NET 6/8. The default codepage in Windows is CP-437, but Github Runners use CP-65001 which is UTF-8.

By running `chcp 65001` in a terminal, and then running the `test-piped-input` plan, the issue can be reproduced locally. I verified that the issue does not reproduce if I run `tap` with .NET 6.

rmadsen-ks (approved):
Looks good!

PR #1123, opened 2023-05-15
1122 make locking package actions lock installations

### LINKED ISSUE #1122: Make LockingPackageActions locking
Currently, LockingPackageActions are only mutually exclusive amongst themselves. E.g. it is not possible to run two concurrent install operations. But it is possible to start e.g. a testplan while a locking package action is running. 

Package actions are typically locking because they need to write to the installation folder. An install action may fail if a test plan is started while it is writing the installation, and the test plan results may be invalid because the installation was partially modified or even corrupted.

I propose making all CLI actions wait for the lock on the installation folder, and release the lock immediately if it is not a locking package action.

alnlarsen:
This makes LockingPackageAction completely lock an installation until the lock is released. 

Closes #1122


Denrask (changes requested):
WAIT!

alnlarsen:
You are right! I have moved the lock check to an earlier stage before assemblies are loaded.

Denrask (dismissed):
This works magnificent!

Denrask (approved):
Lets get it in!

AsgerIversen (approved):

PR #1128, opened 2023-05-17
1127 Unparameterize Two Members: fix and test

rmadsen-ks:
Fixed the issue by defaulting to unparameterizing a member with the same name if the exact same property is not in the list.

Added unit test to verify the behavior.


alnlarsen (approved):
Works!

PR #1140, opened 2023-05-25
Respect OS override when installing packages

### LINKED ISSUE #1139: `tap package install` does not respect `--os` option
Before we updated the image resolver, it was possible to e.g. install a Windows package on Linux with `--os Windows`. The `tap image install` CLI action supports the `--os` option, so I see no reason why the install action shouldn't still support it.

alnlarsen:
Set the image OS if an OS is specified during package install

Closes #1139


rmadsen-ks (approved):
Looks good!

PR #1141, opened 2023-05-31
Parameterization bug fix 2

### LINKED ISSUE #1134: Parameterizing and unparameterizing the same property causes errors
This bug was introduced between 9.18.5 and 9.19.5. It looks like an outdated parameter instance is kept around in a cache somewhere.

rmadsen-ks:
Close #1134


alnlarsen (commented):
Everything seems to work. I have never looked into our ThreadField implementation. Do you think we can discuss it a bit? Why does the ResultListener implementation need a different cache depending on the TapThread?

alnlarsen (approved):
Looks good!

PR #1143, opened 2023-06-07
Added checked conversion from double to TimeSpan for internal code.

### LINKED ISSUE #1142: Test Step DelayStep configured with infinity seconds delay finishes instantly
Our number parse supports infinity. And infinity is also a valid IEEE floating point number.

DelayStep configured with infinity seconds should just wait infinitely, but actually it stops immediately.

rmadsen-ks:
Fixes the issue via overflow checking when converting from double to TimeSpan.

Close #1142

### DIFF
--- a/Shared/ReflectionHelper.cs
+++ b/Shared/ReflectionHelper.cs
@@ -1891,7 +1891,25 @@ namespace OpenTap
         /// <returns></returns>
         public static TimeSpan FromSeconds(double seconds)
         {
-            return TimeSpan.FromTicks((long)(seconds * 1E7));
+            try
+            {
+                checked
+                {
+                    // if the multiplication here creates a number greater than
+                    // long.Max (or overflows negatively), we just return
+                    // TimeSpan.MaxValue/MinValue.
+                    long ticks = (long)(seconds * 1e7);
+                    return TimeSpan.FromTicks(ticks);
+                }
+            }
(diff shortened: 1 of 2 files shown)


Shared/ReflectionHelper.cs
```diff
@@ -1891,7 +1891,25 @@ internal class Time
...
-            return TimeSpan.FromTicks((long)(seconds * 1E7));
+            try
+            {
+                checked
+                {
+                    // if the multiplication here creates a number greater than
+                    // long.Max (or overflows negatively), we just return
+                    // TimeSpan.MaxValue/MinValue.
+                    long ticks = (long)(seconds * 1e7);
+                    return TimeSpan.FromTicks(ticks);
```
alnlarsen, line 1902: Why not just use `TimeSpan.FromSeconds(seconds)`? That variant also throws an overflow exception. It has the same behavior as this checked computation, except it throws an `ArgumentException` if seconds is double.NaN. Should we handle the NaN case?

alnlarsen (changes requested):

Shared/ReflectionHelper.cs
rmadsen-ks: That would be awesome, but last I checked, TimeSpan.FromSeconds rounds to nearest whole millisecond, so we don't use that.

PR #1147, opened 2023-06-08
1146 Embedded Properties with validation rules: fix

rmadsen-ks:
- Modified the unit test to show the issue.
- Fixed the issue by looping through possibly multiple EmbeddedMemberData instances to look for the original source and error. When these are nested the members are arranged in a recursive data structure.
- Added support for when the upper layers are not ValidatingObject, but the lower layers are.

close #1147


alnlarsen (approved):
Fixes the linked issue

PR #1163, opened 2023-06-22
Disable const resource cache and update reflection implementation

alnlarsen:
This disables a cache that was not being invalidated. The surrounding logic makes it very difficult to invalidate the cache, so let's just remove the cache instead. The performance impact should be negligible.


rmadsen-ks (approved):
Looks good!

PR #1165, opened 2023-06-23
Revert locking package actions

alnlarsen:
This reverts the fix for #1122. I am removing this for now in order to de-risk the 9.21.1 patch, but it will be re-added in 9.22.0.


rmadsen-ks (commented):
There are some essential issues with this feature. 
For example it means that it is allowed to run the Editor and then start package installation, but it is not allowed to run package installation and then start the editor (in the meantime).

I think we'd be modifying the behavior to something we have not completely thought thorugh.

PR #1169, opened 2023-06-27
Fixed issue with package.xml that caused signing on linux

StefanHolst:


AsgerIversen (changes requested):
Looks like this also disables signing on windows. We're still going to need that.

Could we just set the `Sign` variable to false in the linux/mac build job?

StefanHolst:
> Looks like this also disables signing on windows. We're still going to need that.
> 
> Could we just set the `Sign` variable to false in the linux/mac build job?

It should only disable signing on .dll files. The `tap.exe` file is still signed on Windows.

alnlarsen:
Can we do this instead: https://github.com/opentap/opentap/pull/1170

alnlarsen (changes requested):
Let's do this instead: https://github.com/opentap/opentap/pull/1170

PR #1176, opened 2023-07-17
Add PosixSignalHandler for handling SigTerm

alnlarsen:
This adds a new handler for treating SIGTERM in the same way we currently treat SIGINT. We don't use the handler on Windows because Windows does not have signals.


rmadsen-ks (changes requested):
Looks good!

rmadsen-ks (approved):

PR #1177, opened 2023-07-17
Debug log installed packages when Installation is instantiated

### LINKED ISSUE #800: SessionLog include information about Installed Packages
It could be helpful in debugging issues, if the SessionLog included information about installed packages (same info as output from `tap pacakge list --installed`.

alnlarsen:
This will log a summary of installed packages at most once per OpenTAP instance when Installation.Current is instantiated.

This should make it easier for us to reproduce issues when people submit bug reports with session logs. 

In execution paths that don't instantiate Installation.Current, this information will not be logged, but that is probably okay. It will always be instantiated in test plan logs because of the ITestPlanRunMonitor which adds the currently installed packages to the run result set.

@witoldpiet the output looks something like this:

```
00:00:00.437 : Installation  : Debug       : Editor                           - 9.21.0+185bec68
00:00:00.437 : Installation  : Debug       : TestWithDependencies             - 2.3.0+9b084a8
00:00:00.437 : Installation  : Debug       : Keysight Licensing               - 1.4.3+4185d1cf
00:00:00.437 : Installation  : Debug       : OpenTAP                          - 9.22.0-alpha.4+89e79075.800-log-installed-packages
00:00:00.437 : Installation  : Debug       : OSIntegration                    - 1.4.2+15f32a31
00:00:00.437 : Installation  : Debug       : Test                             - 1.2.3-alpha+test
00:00:00.437 : Installation  : Debug       : Test1                            - 1.2.3+4
00:00:00.437 : Installation  : Debug       : Test2                            - 1.2.3+4
00:00:00.437 : Installation  : Debug       : Test3                            - 0.0.0
00:00:00.437 : Installation  : Debug       : TUI                              - 1.0.0+48ed6e30
00:00:00.437 : Installation  : Debug       : WPF Controls                     - 9.21.0+185bec68
00:00:00.437 : Installation  : Debug       : Keysight Fixed License Manager   - 1.6.2+d58f8a89
00:00:00.437 : Installation  : Debug       : PathWave License Manager         - 2.7.0+398d3a87
00:00:00.437 : Installation  : Debug       : PathWave License Manager Minimal - 1.1.2+0ff8f1b5
00:00:00.437 : Installation  : Debug       : VCRedist                         - 1.1.0+67bba1c4
```

Closes #800

### DIFF
--- a/Engine/SessionLogs.cs
+++ b/Engine/SessionLogs.cs
@@ -139,7 +139,7 @@ namespace OpenTap
                     // this ensures that System.Runtime.InteropServices.RuntimeInformation.dll is loaded. (issue #4000).
                     .StartNew(PluginManager.Load)
                     // then get the system info on a separate thread (it takes ~1s)
-                    .ContinueWith(tsk => SystemInfo()); 
+                    .ContinueWith(tsk => LogStartupInfo()); 
 
                 AppDomain.CurrentDomain.ProcessExit += FlushOnExit;
                 AppDomain.CurrentDomain.UnhandledException += FlushOnExit;
@@ -481,7 +481,7 @@ namespace OpenTap
         }
 
         static Task SystemInfoTask;
-        private static void SystemInfo()
+        private static void LogStartupInfo()
         {
             foreach (var td in TypeData.GetDerivedTypes<IStartupInfo>().Where(td => td.CanCreateInstance))
             {
@@ -508,7 +508,7 @@ namespace OpenTap
                 }
                 catch (Exception ex)
                 {
-                    log.Debug($"Unhandled exception in '{td.Name}.{nameof(td.LogStartupInfo)}': {ex.Message}");
+                    log.Debug($"Unhandled exception in '{td.Name}.{nameof(si.LogStartupInfo)}': {ex.Message}");
                     log.Debug(ex);
                 }
             }
--- a/Engine/SessionLogs.cs
+++ b/Engine/SessionLogs.cs
@@ -508,7 +508,7 @@ namespace OpenTap
                 }
                 catch (Exception ex)
                 {
-                    log.Debug($"Unhandled exception in '{td.Name}.LogStartupInfo': {ex.Message}");
+                    log.Debug($"Unhandled exception in '{td.Name}.{nameof(td.LogStartupInfo)}': {ex.Message}");
                     log.Debug(ex);
                 }
             }
--- a/Engine/SessionLogs.cs
+++ b/Engine/SessionLogs.cs
@@ -442,7 +442,7 @@ namespace OpenTap
                 }
             });
         }
-        
+
         static string addLogRotateNumber(string fullname, int cnt)
         {
             if (cnt == 0) return fullname;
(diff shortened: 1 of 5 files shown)


Package/Installation.cs
```diff
@@ -86,7 +86,31 @@ public Installation(string directory)
...
         /// </summary>
-        public static Installation Current => current ??= new Installation(ExecutorClient.ExeDir);
+        public static Installation Current
+        {
+            get
+            {
+                if (current == null)
+                {
+                    current = new Installation(ExecutorClient.ExeDir);
+                    current.PrintReadable();
```
rmadsen-ks, line 96: I think it is a nice addition, but also that we should add it a bit more deliberately.

Maybe add it together with 'SessionLog.SystemInfo'?

rmadsen-ks (changes requested):
Should we move it to SessionLog.SystemInfo?

Package/Installation.cs
rmadsen-ks: Then we can also move the system info out of SessionLogs.

alnlarsen: I added an interface and moved the startup logging. Was this what you had in mind?

Engine/SessionLogs.cs
```diff
@@ -482,15 +483,37 @@ public static void Rename(string path)
         static Task SystemInfoTask;
         private static void SystemInfo()
```
rmadsen-ks, line 484: Maybe change the name of this to LogStartupInfo.

Package/Installation.cs
```diff
@@ -10,6 +10,27 @@
 
 namespace OpenTap.Package
 {
+    /// <summary>
+    /// Log information about the current installation at startup.
+    /// </summary>
+    public class InstallationLoggerStartupInfo : IStartupInfo
```
rmadsen-ks, line 16: Make this internal. Move to a separate file.

Engine/SystemStartupInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    public interface IStartupInfo : ITapPlugin
+    {
+        /// <summary>
+        /// LogStartupInfo is called exactly once immediately after session logging has been initialized.
+        /// </summary>
+        void LogStartupInfo();
+    }
+    
+    internal class SystemStartupInfo : IStartupInfo
```
rmadsen-ks, line 18: Move this to a separate file.

Engine/SessionLogs.cs
```diff
@@ -482,15 +483,37 @@ public static void Rename(string path)
...
-                log.Debug(RuntimeInformation.FrameworkDescription); // This becomes something like ".NET Framework 4.6.1586.0"
-            var version = SemanticVersion.Parse(Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion);
-            log.Debug("OpenTAP Engine {0} {1}", version, RuntimeInformation.ProcessArchitecture);
+                try
+                {
+                    si.LogStartupInfo();
+                }
+                catch (Exception ex)
+                {
+                    log.Debug($"Unhandled exception in '{td.Name}.LogStartupInfo': {ex.Message}");
```
rmadsen-ks, line 511:
```suggestion
                    log.Debug($"Unhandled exception in '{td.Name}.{nameof(td.LogStartupInfo)}': {ex.Message}");
```

rmadsen-ks (changes requested):
I like it now, but I have a few ideas for more cosmetic improvements.

alnlarsen:
I agree with all your suggestions. I have made some changes

PR #1178, opened 2023-07-17
1175 Mixins: Add mixins

rmadsen-ks:
Note, this is just a draft.

- Moved embedded property tests. Added the mixin tests.
- Added support for two test step mixins and one resource mixin.


rmadsen-ks:
This will not be merged as other work replaced it.

PR #1179, opened 2023-07-18
Fix error preventing bundle downgrades

### LINKED ISSUE #1159: The version of plugins in DMM bundle does not change when bundle is downgraded
1. install DMM bundle v2.1.0
2. Downgrade DMM bundle to 2.0.3

Prior to 9.19, this would downgrade all the members of DMM, but now the image fails to install:

```
Could not resolve DMM v2.0.3
Image configuration:
  DMM: 2.0.3
  DMM API: ^2.1.3+26062d00
  DMM Instruments: ^2.1.3+26062d00
  DMM Test Steps: ^2.1.3+26062d00
  OpenTAP: ^9.23.2-beta.5+40fdd208
```

alnlarsen:
Closes #1159

### DIFF
--- a/Package/Image/ImageSpecifier.cs
+++ b/Package/Image/ImageSpecifier.cs
@@ -53,7 +53,15 @@ namespace OpenTap.Package
             {
                 var ext = installed.FirstOrDefault(x => x.Name == package.Name);
                 if (ext != null)
+                {
                     installed.Remove(ext);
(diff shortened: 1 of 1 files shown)


rmadsen-ks (changes requested):
It seems this does not solve the underlying problem. 

For example if a bundle has been force-uninstalled or one of the packages has been installed beforehand.

PR #1180, opened 2023-07-18
1052 fix nuget syswide install

### LINKED ISSUE #1052: NuGet package: Build fails when installing system-wide packages
We have this msbuild target for verifying that an image installation succeeded: 

```xml
<Target Name="VerifyInstalledPackages"
        DependsOnTargets="ProcessOpenTapPackageList"
        BeforeTargets="GenerateOpenTapReferenceProps">
  <ItemGroup>
    <PackageXmlFiles Include="@(OpenTapPackagesToInstall-> '$(OutDir)Packages/%(Identity)/package.xml')"/>
  </ItemGroup>
  <Error Condition="'@(PackageXmlFiles)' != '' AND !Exists('@(PackageXmlFiles)')" Text="Some packages failed to install."/>
</Target>
```

It checks if a `package.xml` file exists for all the packages specified in the project. This fails for system-wide packages which are not added to the debug installation, which causes all builds to fail.

alnlarsen:
This detects when a system-wide package is requested, and fails the build immediately if this is the case. This will also give an error message stating the exact problem, instead of cryptically saying 'Some packages failed to install' with no further information.

Example output:

```bash
> dotnet build
MSBuild version 17.7.0-preview-23251-02+59879b095 for .NET
...
project.csproj : error OpenTAP Install: Requested package 'PathWave License Manager' is system-wide, and should not be in
stalled as part of a build.
project.csproj : error OpenTAP Install: Failed to install packages.
    0 Warning(s)
    2 Error(s)
```
Closes #1052

### DIFF
--- a/sdk/OpenTap.Sdk.MSBuild/OpenTapImageInstaller.cs
+++ b/sdk/OpenTap.Sdk.MSBuild/OpenTapImageInstaller.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
-using System.Xml.Linq;
 using Microsoft.Build.Framework;
 using OpenTap;
 using OpenTap.Diagnostic;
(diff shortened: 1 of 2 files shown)


alnlarsen:
Tested with @Denrask, seems to work!

sdk/OpenTap.Sdk.MSBuild/OpenTapImageInstaller.cs
```diff
@@ -17,11 +16,41 @@ internal interface IImageDeployer
...
+            return package.Class.Equals("system-wide", StringComparison.OrdinalIgnoreCase);
+        }
+
+        public void Install(ImageSpecifier spec, Installation install, CancellationToken ct)
+        {
+            ImageIdentifier image = null;
+            // As a performance optimization, don't start resolving the image if a package
+            // which is known to be system-wide was requested.
+            var requestedPackages = spec.Packages.Select(p => p.Name).ToArray();
+            var requestedSystemWide = install.GetPackages().Where(IsSystemWide)
```
rmadsen-ks, line 30: It looks like this is only able to detect the error if the system wide package being installed is already installed and as such does not work in general.

Could we find a better solution?

alnlarsen:
This is not needed

PR #1181, opened 2023-07-18
Change default local repository path to use LocalPath format

### LINKED ISSUE #1160: Package Manager: The local repository path in Settings is not user friendly
The path to the current installation in the package manager settings is not human readable. On Windows it starts with "file://C:/" and uses URI escaping, e.g. %20 instead of spaces. It used to just be a normal unescaped string starting with "C:/".

alnlarsen:
Closes #1160


rmadsen-ks (approved):
Seems to work.

PR #1185, opened 2023-07-20
1184 Stream Many To: Added streaming for SaveManyTo.

### LINKED ISSUE #1184: Memory Optimize PackageDef.SaveManyTo
Currently SaveManyTo buffers all data before writing it to the stream.

We should stream it instead

rmadsen-ks:
Closes #1184


alnlarsen (approved):
Let's do this

PR #1188, opened 2023-07-24
1183 ParameterizeUnparameterizeBug: Fix

### LINKED ISSUE #1183: Error on test execution after removing and adding parameter on parent
Reproduced on both Editor X and Editor

OpenTAP 9.21.1

**Environment:**
 OS: Ubuntu/Windows

**To Reproduce**
1. Add a Sequence with a Delay child
2. Time Delay - > Parameterize on parent
3. Run the test plan - > works, no errors
4. Time Delay - > Unparameterize
5. Time Delay - > Parameterize on parent
6. Run the test plan - > Error. Log: 
[Session-fe6e60d7-fab0-4c38-bc4a-9a315d7acd3b-2023-18-07-14-13.txt](https://github.com/opentap/opentap/files/12103298/Session-fe6e60d7-fab0-4c38-bc4a-9a315d7acd3b-2023-18-07-14-13.txt)

image

rmadsen-ks:
Tried to fix the issue by using immutable data structures.

I modified the cache so that it is only present during the test plan run.

Also a couple of the tests depended on the order of the members which has been reversed. In the cases where it made sense I ordered them, but in general the order does not matter.

Close #1183

### DIFF
--- a/BasicSteps/SweepParameterStep.cs
+++ b/BasicSteps/SweepParameterStep.cs
@@ -120,13 +120,16 @@ namespace OpenTap.Plugins.BasicSteps
             if (SweepValues.Count == 0)
                 return "No rows selected to sweep.";
             var rowType = SweepValues.Select(TypeData.GetTypeData).FirstOrDefault();
-            var sb = new StringBuilder();
+            
             var numErrors = 0;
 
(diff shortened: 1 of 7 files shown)


StefanHolst (approved):
Looks good

PR #1195, opened 2023-07-27
1194 Expression And String interpolation: Fix

### LINKED ISSUE #1194: Expressions and string interpolations
- Allow basic math expressions that can work on local data (step members). 
- Allow basic string interpolation-like expression for settings with string type. e.g "The frequency is: {Frequency} hz"
More to come...

Related: #1161

rmadsen-ks:
- Added expression parsing and expression manager.
- Added Updating expressions in DoRun.
- Added a number of unit tests testing the functionality.
- Added context menu for adding, modifying and removing expressions.
- Added math expressions: sin,cos,tan,ata,abs,floor,ceiling,Pi,π. And misc false,true and isempty(str).
- Math operators: +,-,/,*
- logical operators: <, >, <=, >=, ==, !=, &&, ||

Close #1194


StefanHolst (changes requested):
Looks good overall and works pretty good. I do see some missing validation errors for string expressions.

.github/workflows/ci.yml
```diff
@@ -3,11 +3,9 @@ name: CI
 on:
   push:
     branches:
-      - main
-      - release9x
+      - '*'
```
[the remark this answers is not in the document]
rmadsen-ks: Yes, this was intentional to get proper alpha names.

Engine.UnitTests/ExpressionTest.cs
```diff
@@ -0,0 +1,282 @@
...
+            var ast0 = builder.ParseStringInterpolation(expression);
+            var ast = ast0.Unwrap();
+            var lmb = builder.GenerateLambda(ast, ParameterData.Empty, typeof(string));
+            var result = lmb.Unwrap().DynamicInvoke();
+            Assert.AreEqual(expectedResult, result);
+        }
+
+        [TestCase( "123 {floor(1.5} 321432", "Unexpected symbol '}'.", null)]
+        [TestCase( "123 {1.5) 321432", "Unexpected symbol ')'.", null)]
+        public void StringExpressionParseErrors(string errorExpression, string parseError, string compileError)
```
frederikja163, line 81: If both of these have null for the compile error, is there a reason its included as a parameter?

Engine/Annotations/HasExpressionAnnotation.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using OpenTap.Expressions;
+namespace OpenTap
+{
+    class HasExpressionAnnotation : IIconAnnotation, IEnabledAnnotation, IInteractiveIconAnnotation, IExpressionAnnotation, IOwnedAnnotation, IErrorAnnotation
```
frederikja163, line 6: Not sure if we do this elsewhere, but i would highly suggest adding sealed to classes, if no classes inherit from them.
Makes calling methods on the class roughly 10x faster, and uses a locked down by default style, meaning you have fewer breaking changes that needs to be considered in the future.
```suggestion
    class HasExpressionAnnotation : IIconAnnotation, IEnabledAnnotation, IInteractiveIconAnnotation, IExpressionAnnotation, IOwnedAnnotation, IErrorAnnotation
```

Engine/Expressions/AstNode.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace OpenTap.Expressions
+{
+    /// <summary> Abstract Syntax Tree Node. This is the output of the expression parser. </summary>
+    class AstNode
```
frederikja163, line 4:
```suggestion
    abstract class AstNode
```

Engine/Expressions/BinaryExpressionNode.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace OpenTap.Expressions
+{
+    /// <summary> Binary expressions AST node. This describes the application of a binary operator for example '*' in 5 * 4.</summary>
+    class BinaryExpressionNode : AstNode
```
frederikja163, line 4:
```suggestion
    sealed class BinaryExpressionNode : AstNode
```

Engine/Expressions/BinaryExpressionNode.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary> Left side of the operation. </summary>
+        public AstNode Left;
```
frederikja163, line 7: Again maybe more of a style thing. But im more used to public members being properties rather than fields. I also think its in the c# style guides somewhere.

Engine/Expressions/MathExpressionFunctionProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+            public static double atan(double v) => Math.Atan(v);
+            public static double abs(double v) => Math.Abs(v);
+            public static double floor(double v) => Math.Floor(v);
+            public static double ceiling(double v) => Math.Ceiling(v);
+            public static double round(double v) => Math.Round(v);
+            public static double round(double v, int decimals) => Math.Round(v, decimals);
+            public static double Pi => π;
+            public static double π => Math.PI;
+            public static bool @false => false;
+            public static bool @true => true;
```
frederikja163, line 24: Maybe we should include pow, sqrt, log, max, min, round, ceil, floor and sign as functions here aswell?

Engine/Expressions/Operators.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Linq;
+using System.Reflection;
+namespace OpenTap.Expressions
+{
+    class Operators
```
frederikja163, line 5:
```suggestion
    static class Operators
```

Engine/Expressions/ParameterData.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <param name="array"></param>
+        /// <returns>True if the list has been updated. </returns>
+        internal static bool UpdateParameterMembers(object obj, ref ImmutableArray<IMemberData> array)
+        {
+            var newArray = TypeData.GetTypeData(obj).GetMembers()
+                .Where(mem => mem.Readable && mem.IsBrowsable() && (mem.HasAttribute<SettingsIgnoreAttribute>() == false) && (mem.HasAttribute<AnnotationIgnoreAttribute>() == false))
+                .ToImmutableArray();
+            if (newArray.Equals(array))
```
frederikja163, line 65: I dont think the new array will ever be equal to the old array. IIRC ImmutableArray.Equals checks if the objects are the same. You might need to use SequenceEqual instead to check if the contained elements are updated.

frederikja163 (approved):
Very nice, and super useful

Engine/Annotations/HasExpressionAnnotation.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using OpenTap.Expressions;
+namespace OpenTap
+{
+    class ExpressionUpdaterAnnotation : IOwnedAnnotation
+    {
+        public static ExpressionUpdaterAnnotation Instance { get; } = new ExpressionUpdaterAnnotation();
+        public void Read(object source)
+        {
+            if (source is ITestStepParent parent)
```
StefanHolst, line 11: Looks unnecessary

StefanHolst (changes requested):
The code looks great 🗡️ 

But when I test it with the TUI, it seems expression errors are not cleared when fixed. It might be a TUI issue though.


image


Also, the build fails and there's conflicts

rmadsen-ks:
> The code looks great 🗡️
> 
> But when I test it with the TUI, it seems expression errors are not cleared when fixed. It might be a TUI issue though.
> 
> image
> 
> Also, the build fails and there's conflicts

I am not able to reproduce with Editor. I think something is not updating properly in the TUI.

StefanHolst (approved):
Beautiful!

PR #1197, opened 2023-07-28
1196 Open Close Order Lazy FIx

rmadsen-ks:
When the reference count of a resource is greater than 1, it fails to correctly decrement the reference count.

This only occur in test plans with steps running in parallel. And it may not be a problem for few counts since they are not guaranteed to run in parallel.

The problem is that two reference counts are maintained and both are counted up on open, but one is only counted down when the other becomes zero.


StefanHolst (approved):
Great !

PR #1198, opened 2023-07-31
Mixins with UI

rmadsen-ks:


Engine/Annotations/IStringOverlayAnnotation.cs
alnlarsen, line 1: Do we need this?

rmadsen-ks: Yes, that is so that we can provide a GUI overlay for the expression.

Engine/Mixins/IMixin.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            var emb = TypeData.GetTypeData(target).GetBaseType<EmbeddedTypeData>();
+            if (emb == null) return arg;
+            foreach (var mem in emb.GetEmbeddingMembers())
+            {
+                if (!mem.TypeDescriptor.DescendsTo(typeof(T2))) continue;
+                if (mem.Readable == false) continue;
+                if (mem.GetValue(target) is T2 mixin)
+                {
+                    f(mixin, arg);
```
alnlarsen, line 25: What does it look like if `f` throws? It's probably obvious which mixin threw, and what the `MixinEvent` was, but I don't think the trace will have any information about what step or resource the mixin is attached to, or what the input was? 

Should we add something like a `MixinException` with some additional information, and an inner exception?

Engine/SerializerPlugins/MixinSerializer.cs
```diff
@@ -0,0 +1,62 @@
...
+        
+        public override bool Deserialize(XElement node, ITypeData t, Action<object> setter)
+        {
+            var rootElem = node.Element(rootElemName);
+            if (rootElem != null)
+            {
+                var elements = rootElem.Elements();
+                void setter2(object targetObject)
+                {
+                    var targetType = TypeData.GetTypeData(targetObject);
```
alnlarsen, line 21:
```suggestion
```

Engine/SerializerPlugins/MixinSerializer.cs
```diff
@@ -0,0 +1,62 @@
...
+                        mixins = new XElement(rootElemName);
+                        node.Add(mixins);
+                    }
+                    XElement elem = new XElement(mixinElemName);
+                    mixins.Add(elem);
+                    Serializer.Serialize(elem, mixin.Source);
+                }
+            }
+
+            return false;
```
alnlarsen, line 59: This always returns false. Is that intended?

Shared/ReflectionHelper.cs
```diff
@@ -1913,6 +1909,21 @@ public static T PopAt<T>(this IList<T> list, int index)
...
+                    yield break;
+            var ctxValue = ctx();
+            while (true)
+            {
+                yield return (e.Current, ctxValue);
+                if (e.MoveNext() == false)
+                    break;
+            }
+        } 
+
```
alnlarsen, line 1926: It doesn't look like this is used anywhere, and I don't understand how it's useful?
```suggestion
```

alnlarsen (changes requested):
On the whole I think this looks good.

Engine/Mixins/IMixin.cs
rmadsen-ks: We dont need MixinException for now.

Shared/ReflectionHelper.cs
rmadsen-ks: 👍 lets delete that!

Engine/SerializerPlugins/MixinSerializer.cs
rmadsen-ks: I have added a comment.

alnlarsen (dismissed):
Looks good!

alnlarsen (approved):

Issue #1182, opened 2023-07-20
Multiple edit: Run Program: Environment Variables does not work


sebastian-pop:
**Describe the bug**

**Environment (please complete the following information):**

 - OS: Windows 10
 - OpenTAP Version 9.21.1

**To Reproduce**

1. Add two Run Program steps
2. Select both
3. Edit the `Environment variables` setting 

**Actual behavior**

Cannot add new rows

**Expected behavior**

If it's intended for this setting to be edited on multiple steps it should work

Otherwise it should be hidden on multiple selection

**Additional context**

Add any other context about the problem here. Please also attach a log file from <InstallDir>\SessionLogs\ that shows the issue when possible/relevant.

rmadsen-ks:
This has been slightly improved by #1202, but still some modifications are needed to make it work perfectly.

rmadsen-ks:
Actually this seems to work in the TUI now, so I will move this issue to the ks8400 issue tracker.

PR #1202, opened 2023-08-09
1182 Environment Variable bug Improvement:

rmadsen-ks:
Improved the situation when multi selecting Run Program step and editing environment variables.

It is still not perfect, but at least now it somewhat works.

Moves towards fixing #1182, but it is still not perfect.


StefanHolst (approved):
Looks good

PR #1224, opened 2023-08-30
1223 OpenTapPackageReference Token: Fix

rmadsen-ks:
- Added support for setting tokens from csproj files.
- Added cli action set-auth-token (Browsable(false)) for now to set the user token,

I verified that this works with or without using an environment variable.

```<OpenTapPackageReference Include="Test 2" Version="0.1.0-alpha" Token="$(OPENTAP_BUILD_TOKEN)" Repository="https://packages.opentap.io" />```

### DIFF
--- a/Package/PackageActions/SetUserToken.cs
+++ b/Package/PackageActions/SetUserToken.cs
@@ -1,0 +1,45 @@
+using System.ComponentModel;
+using System.Linq;
+using System.Threading;
+using OpenTap.Authentication;
+using OpenTap.Cli;
+namespace OpenTap.Package
+{
+    /// <summary> Sets the authentication token for use with a given domain.  </summary>
+    [Display("set-auth-token", Group: "package", Description: "Sets the authentication token for a given domain.")]
+    [Browsable(false)]
+    public class SetUserToken : ICliAction
+    {
+        /// <summary>  Gets or sets the domain. </summary>
+        [UnnamedCommandLineArgument("Domain")]
+        public string Domain { get; set; }
+
+        /// <summary>  Gets or sets the token. </summary>
+        [UnnamedCommandLineArgument("Token")]
+        public string Token { get; set; }
+        
+        /// <summary>  Gets or sets update - if existing settings should be updated. </summary>
+        [CommandLineArgument("update", ShortName = "u", Description = "Sets if existing settings should be updated.")]
+        public bool Update { get; set; }
+
+        static readonly TraceSource log = Log.CreateSource("set-auto-token");
+        
+        public int Execute(CancellationToken cancellationToken)
+        {
+            if (AuthenticationSettings.Current.Tokens.Any(x => x.Domain == Domain))
+            {
+                if (!Update)
+                {
+                    log.Error($"Settings already contains a token for {Domain}. Use --update to overwrite it.");
+                    return -1;
+                }
+            }
+            
+            AuthenticationSettings.Current.Tokens.RemoveIf(x => x.Domain == Domain);
+            AuthenticationSettings.Current.Tokens.Add(new TokenInfo(Token, "", Domain));
+            AuthenticationSettings.Current.Save();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by alnlarsen** [comment]
This is really cool. Much nicer than having to install Cloud Drive and completing a login flow.

Can we improve the usability a bit for `set-auth-token`? If I want to login in to the repository 'https://test-automation.pw.keysight.com/api/packages', I have to strip the https scheme and /api/packages subpath, otherwise I get his error: `Domain should only be the host part of a URI and not a full absolute URI.`

I think it would be more natural for users to be able to just use the repository URL instead of having to know the specifics of how OpenTAP compares URIs.

File: Package/PackageActions/SetUserToken.cs [hunk 1]
```diff
@@ -0,0 +1,45 @@
+using System.ComponentModel;
+using System.Linq;
+using System.Threading;
+using OpenTap.Authentication;
+using OpenTap.Cli;
+namespace OpenTap.Package
+{
+    /// <summary> Sets the authentication token for use with a given domain.  </summary>
+    [Display("set-auth-token", Group: "package", Description: "Sets the authentication token for a given domain.")]
+    [Browsable(false)]
```
alnlarsen, line 10: Can we make this browsable?

Package/PackageActions/SetUserToken.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>  Gets or sets the token. </summary>
+        [UnnamedCommandLineArgument("Token")]
+        public string Token { get; set; }
+        
+        /// <summary>  Gets or sets update - if existing settings should be updated. </summary>
+        [CommandLineArgument("update", ShortName = "u", Description = "Sets if existing settings should be updated.")]
+        public bool Update { get; set; }
+
+        static readonly TraceSource log = Log.CreateSource("set-auto-token");
```
alnlarsen, line 25:
```suggestion
        static readonly TraceSource log = Log.CreateSource("set-auth-token");
```

sdk/OpenTap.Sdk.MSBuild/InstallOpenTapPackages.cs
```diff
@@ -119,6 +120,88 @@ public void Cancel()
         {
             cts.Cancel();
         }
+
+        /// <summary> Load Authentication tokens from Items into settings. </summary>
+        /// <returns>True if successful.</returns>
+        bool ProcessAuthTokens()
```
alnlarsen, line 126: I feel like this is encouraing people to put user tokens directly in their .csproj / check it into source control. I know we will use environment variables for this, but I don't feel great about it.

Package/PackageActions/SetUserToken.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Linq;
+using System.Threading;
+using OpenTap.Authentication;
+using OpenTap.Cli;
+namespace OpenTap.Package
+{
+    /// <summary> Sets the authentication token for use with a given domain.  </summary>
+    [Display("set-auth-token", Group: "package", Description: "Sets the authentication token for a given domain.")]
+    [Browsable(false)]
+    public class SetUserToken : ICliAction
```
alnlarsen, line 11: Can we call it `login` so it is a single verb like the other package actions? `create`, `list` etc.

Package/PackageActions/SetUserToken.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        /// <summary>  Gets or sets the domain. </summary>
+        [UnnamedCommandLineArgument("Domain")]
+        public string Domain { get; set; }
```
alnlarsen, line 15: Rename this to repository and copy-paste the description / flags from the other package actions

alnlarsen (changes requested):
This is really good

rmadsen-ks:
@alnlarsen, please check it again.

alnlarsen (approved):
Looks good!

Issue #1225, opened 2023-09-05
Image fails to resolve even if required packages are already installed


alnlarsen:
Consider this chain of commands:
```bash
> ./tap image install "Runner"
> ./tap package list -i
Cloud Drive - 0.2.0+7e7541d6  - installed
OpenTAP     - 9.21.1+2a94acc7 - installed
Runner      - 1.4.0+feac26dd  - installed
# Copy runner package to local directory
> mv ~/.local/share/OpenTap/PackageCache/Runner.1.4.0+feac26dd.Windows,Linux,Macos.TapPackage ~/local-repo
# Delete package cache
> rm -rf ~/.local/share/OpenTap/PackageCache
# Install image of Runner from local repo
> ./tap image install "Runner" -r ~/local-repo 
Unable to resolve image 'Runner:1.4.0+feac26dd': OpenTAP: ^9.21.1+2a94acc7, Cloud Drive: ^0.2.0+7e7541d6 was not found
```
The image will fail to resolve even if all the required packages are already installed. 
Running the command with `packages.opentap.io` enabled succeeds with this message: `Target installation is already up to date.` 

We should succeed with the same message in the offline scenario.

PR #1226, opened 2023-09-06
1225 add installed packages to cache during image install

alnlarsen:
This fixes an issue causing the image resolution to fail if an already installed package is not available in any of the specified repositories or the package cache.

This issue also reproduces for `tap package install...` under the same circumstances, but I decided to only fix this for image installs, since they behave slightly differently: `image install` succeed telling you the target installation is up to date if no changes would be made, but `package install` always tries to uninstall and reinstall packages, which is not possible if the package is not available.

Cloes #1225

### DIFF
--- a/Package/PackageActions/ImageInstall.cs
+++ b/Package/PackageActions/ImageInstall.cs
@@ -94,6 +94,7 @@ namespace OpenTap.Package
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


rmadsen-ks (changes requested):
Can you add a comment about why this is needed?

rmadsen-ks (approved):
Looks good

PR #1231, opened 2023-09-07
1230 Expressions are not always a dependency: fix

rmadsen-ks:
Two things were wrong:

1. If the default value for a property was used - for example null in the case of ExpressionCollection. Then the XML would still be marked as having a dependency.
2. The ExpressionMember did not get marked as a dependency from the Expressions package because the ExpressionCollection type did not get loaded the normal way and hence TypeData.FromType(typeof(ExpressionObject)).Assembly was null and it did not recognize that it came from that package.

 1 was fixed in the serializer by checking if the property had default value and if so the type was not marked as being used.

2 was fixed by adding the Assembly reference when loading a TypeData 'FromType'. To do this the assembly searcher had to be modified a bit to make it possible to search for an assembly.

### DIFF
--- a/Engine/TypeData.cs
+++ b/Engine/TypeData.cs
@@ -241,6 +241,9 @@ namespace OpenTap
         TypeData elementType;
         ITypeData baseTypeCache;
 
+        // this value is used to mark a type code that has not been loaded
+        // all the normal values of TypeCode means something specifically
+        // so this value is used as something that does not have other meaning.
         const int UnloadedTypeCode = 100;
         
         TypeCode typeCode = (TypeCode)(UnloadedTypeCode);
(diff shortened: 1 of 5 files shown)


Engine/TypeData.cs
```diff
@@ -240,7 +240,10 @@ internal bool IsString
         bool failedLoad;
         TypeData elementType;
         ITypeData baseTypeCache;
-        TypeCode typeCode = TypeCode.Object;
+
+        const int UnloadedTypeCode = 100;
+        
+        TypeCode typeCode = (TypeCode)(UnloadedTypeCode);
```
alnlarsen, line 246: I don't really understand why you added this. When would `type` be non-null if the assembly wasn't loaded?

alnlarsen (commented):
The changes look fine but I am a bit confused about the typecode thing.

Engine/TypeData.cs
rmadsen-ks: All normal values of TypeCode means something, so we can detect that the typecode has not been set by checking if it is this value.

alnlarsen (approved):
Looks good!

PR #1234, opened 2023-09-08
1233 Outputs on Scpi step and Process Step: Fix

rmadsen-ks:
- Added output on both classes (optional).
- Added exit code for ProcessStep
- HideIfDisabled for many settings.


alnlarsen (approved):
Looks like some nice improvements across the board. LGTM!

PR #1238, opened 2023-09-11
Look for system-wide files in system-wide directory

### LINKED ISSUE #1083: 'Tap package verify' command fails for OpenTAP: 'tap.exe' has non-matching checksum
**Describe the bug**

Package verify fails for OpenTAP: 'tap.exe' has non-matching checksum

 - OS: Windows 10 x64
 - OpenTAP: 9.20.4

**To Reproduce**

1. Install OpenTAP
2. Open Command Prompt in the OpenTAP installation folder
3. Run "tap package verify OpenTAP" 

**Actual behavior**

The output returned is:

Package 'OpenTAP' not verified.
File 'tap.exe' has non-matching checksum

**Additional context**

tap_package_verify

This issue also occurs on OpenTAP 9.18 & 9.19.

### LINKED ISSUE #1237: 'Tap package verify' command fails for system-wide packages
It seems OpenTAP is looking for files from system-wide packages in the current installation for the purposes of `tap package verify`. We should check the system-wide directory instead.

alnlarsen:
This fixes a couple of `tap package verify` issues:

* It tried looking for system-wide files in the current installation. Now it looks in the system-wide directory
* The `Tap.Upgrader` logic tried to replace `tap.exe.new` first, and then `Packages/OpenTAP/tap.exe.new` afterwards, which has a non-matching checksum because `tap.exe` and `tap.exe.new` are signed separately although they are copies of each other.
By swapping the order, we prefer `./tap.exe.new` over `./Packages/OpenTAP/tap.exe.new` if it is present (this is the case when upgrading from OpenTAP 9.17 or later)

Closes #1237
Closes #1083

### DIFF
--- a/Package/CreatePackage/FileHashPackageAction.cs
+++ b/Package/CreatePackage/FileHashPackageAction.cs
@@ -139,7 +139,11 @@ namespace OpenTap.Package
                         log.Debug("Replacing '\\' with '/' in {0}", file.FileName);
                         file.FileName = repl;
                     }
-                    string fullpath = Path.Combine(Target, file.FileName);
+
+                    string fullpath =
+                        Path.Combine(pkg.IsSystemWide() ? PackageDef.SystemWideInstallationDirectory : Target,
+                            file.FileName);
+                    
                     var hash2 = new FileHashPackageAction.Hash(FileHashPackageAction.hashFile(fullpath));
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks good!

PR #1241, opened 2023-09-12
Remove VerifyInstalledPackages check

### LINKED ISSUE #1052: NuGet package: Build fails when installing system-wide packages

alnlarsen:
I don't think this check is actually needed anymore because Image installs are actually reliable. We needed it when packages were installed in sequence which is no longer the case.

I tried this in Visual Studio, and it seems to behave well.

Closes #1052


rmadsen-ks (approved):
Looks awesome!

PR #1246, opened 2023-09-14
1245 Mixins Settings Results bug: Fix

rmadsen-ks:
- Moved slightly the place at which the mixins gets evaluated. This was done to facilitate expressions setting properties marked with ResultAttribute.
- Added unit tests that verifies the behavior.


alnlarsen (approved):
Nice

PR #1248, opened 2023-09-18
Made MixinMemberData SetValue and GetValue virtual.

rmadsen-ks:
This means that mixins should be able to implement advanced setter/getter behavior if necessary.


alnlarsen (approved):
Looks good!

PR #1249, opened 2023-09-20
Added support for validation rules for mixin builders.

rmadsen-ks:
- Added validation rules for mixin builders
- Added test that verifies the funcitonality.
- Added a fix that caused only basic type info to be shown in the MenuAnnotations.

### DIFF
--- a/Engine.UnitTests/EmbeddedPropertyTests.cs
+++ b/Engine.UnitTests/EmbeddedPropertyTests.cs
@@ -146,11 +146,10 @@ namespace OpenTap.UnitTests
             }
         }
         
-        public class RepeatMixin : ITestStepPostRunMixin, ITestStepPreRunMixin
+        public class RepeatMixin2 : ITestStepPostRunMixin, ITestStepPreRunMixin
         {
             [Display("Repeat Count")]
             public int Count { get; set; } = 1;
-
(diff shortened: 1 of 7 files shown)


Engine.UnitTests/EmbeddedPropertyTests.cs
```diff
@@ -146,12 +146,11 @@ public void OnPostRun(TestStepPostRunEventArgs step)
             }
         }
         
-        public class RepeatMixin : ITestStepPostRunMixin, ITestStepPreRunMixin
+        public class RepeatMixin2 : ITestStepPostRunMixin, ITestStepPreRunMixin
```
alnlarsen, line 149: Why was this renamed?

alnlarsen (commented):
This all looks good but I'm wondering why the test mixin was renamed

Engine.UnitTests/EmbeddedPropertyTests.cs
rmadsen-ks: Just to avoid clashing with a mixin defined by Basic Mixins. Its just a unit test, so it should not cause a problem.

alnlarsen (approved):
Looks good :)

PR #1253, opened 2023-09-26
Small cleanups:

rmadsen-ks:
- Removed LoginAction
- Changed process and scpi step so that they always save the output in a property.


alnlarsen (approved):
Cool :)

PR #1261, opened 2023-09-29
1259 Disabled Ambiguity Warning

rmadsen-ks:
- Added a few extra cases to the test.
- Removed the warning since we dont think it has actually helped anybody.


alnlarsen (approved):
I concur

PR #1266, opened 2023-10-02
Avoid assigning console property if it is not needed

### LINKED ISSUE #1265: Linux: `tap` does not behave well as a background job in shells
When starting `tap` as a background job with e.g. `tap package list &`, the process gets suspended before producing any output. The command will not proceed until it is foregrounded.

Interestingly, this does not happen when stdin is redirected, e.g.: `tap package list <<< '' &`

This happens for any tap command, so it is not related to user input.

alnlarsen:
This fixes an issue causing tap to become suspended when run as a background process.

Closes #1265


rmadsen-ks (approved):
Great fix!

PR #1267, opened 2023-10-03
Update documentation about installing on Linux

### LINKED ISSUE #1263: Documentation not accurate for Linux install
A customer pointed out to me that the documentation here refers to the previous method of unpacking the .tar, while the current install method is launching an executable. This should be updated to reflect the new installation method.

Documentation I'm referring to: https://doc.opentap.io/#_2-install-opentap

alnlarsen:
Update Linux install documentation. Update Linux smartinstaller.

Smartinstaller changes:
* Fix dotnet installation on Ubuntu 22.04 which was broken
* Fix bug which prevented --quiet installs on platforms without webkit2gtk installed
* Fix bug which caused installation to fail if target path contains spaces

Closes #1263

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -792,7 +792,7 @@ jobs:
     container:
       #This sha hash points to beta.7 (which was also force-overwritten). beta.7 contains a fix which allows the installer
       #to run in quiet mode when webkit2gtk is not installed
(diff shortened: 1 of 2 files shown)


sebastian-pop:
@alnlarsen 9.22.0-alpha.60.3+17c13c1e.1263-update-linux-install-docs

Confirmed fix on Ubuntu 22.04. It's now a one-click installer. Confirmed it still works on Ubuntu 20.04

Confirmed fix for installation path with spaces

rmadsen-ks (approved):
Looks good!

PR #1273, opened 2023-10-04
dotnet new ... --verbosity quiet

rmadsen-ks:
Set verbosity quiet when installing the nuget package with sdk. Otherwise it will generate warnings when installing on top of an existing nuget installation.


alnlarsen (approved):
Awesome

PR #1274, opened 2023-10-05
Update Linux smartinstaller

### LINKED ISSUE #1275: Linux: SmartInstaller fails when webkit2gtk is not installed
This was fixed in a newer version of smartinstaller, but reintroduced here when the installer was downgraded to fix a signing issue.

alnlarsen:
This updates the Linux smartinstaller to a version which contains a fix for #1275 

Closes #1275


rmadsen-ks (approved):
Looks good.

PR #1281, opened 2023-10-06
1280 Clone Mixins Using Serializers: fix

rmadsen-ks:
- Using TapSerializere to clone mixin builders instead of using their own implementation - if possible IClonable can be implemented.

- Removed Clone from IMixinBuilder (minor Breaking change from OpenTAP 9.22 RC).

- Added calling IMixinBuilder.Initialize also when modifying an existing mixin.

### DIFF
--- a/Shared/ReflectionHelper.cs
+++ b/Shared/ReflectionHelper.cs
@@ -882,7 +882,16 @@ namespace OpenTap
             del += () => f(v); 
             return del;
         }
-
+        
+        /// <summary> Clones a object. creates a new serializer if needed, but reuses an existing one if possible.</summary>
+        public static T Clone<T>(T obj, ref TapSerializer cachedSerializer)
+        {
+            if (obj is ICloneable cloneable)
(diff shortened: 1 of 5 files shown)


Shared/ReflectionHelper.cs
```diff
@@ -882,7 +882,16 @@ public static Action Bind<T>(this Action del, Action<T> f, T v)
...
         }
-
+        
+        /// <summary> Clones a object. creates a new serializer if needed, but reuses an existing one if possible.</summary>
+        public static T Clone<T>(T obj, ref TapSerializer cachedSerializer)
+        {
+            if (obj is ICloneable cloneable)
+                return (T)cloneable.Clone();
+            
+            return (T)(cachedSerializer ??= new TapSerializer()).Clone(obj);
```
alnlarsen, line 892: This cached serializer is a bit crazy. Could the caller be responsible for providing a (potentially cached) serializer instead?

alnlarsen (approved):
Looks good

PR #1283, opened 2023-10-09
1282-GetSettingsLookupRaceFix: Test

### LINKED ISSUE #1282: Race Condition in GetSettingsLookup
GetSettingsLookup has a cache that consists of essentially a dictionary. In some cases (e.g ParallelStep) that cache can get updated in parallel. Since its just a cache it should be ok, if two threads adds the same value to it at the same time, but since the Add function is being used, that will throw an exception if the key is already inserted. Instead, the new value should be updated or ignored.

This can be reproduced with a parallel step, but its quite hard to do reliably.

rmadsen-ks:
- Fixed the issue by using ImmutableDictionary.SetItem instead of ImmutableDictionary.Add.
   The difference is that SetItem updates values of existing keys while Add throws an exception if the key exists in the dictionary.

- Added unit test to show the issue.

Close #1282

### DIFF
--- a/Engine.UnitTests/TestStepTests.cs
+++ b/Engine.UnitTests/TestStepTests.cs
@@ -1,5 +1,9 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
 using NUnit.Framework;
 using OpenTap.Plugins.BasicSteps;
+using OpenTap.UnitTests;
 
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
Looks good, lgtm

PR #1284, opened 2023-10-09
1279 Result Property String: Test

rmadsen-ks:
- Added unit test to show the issue.
- Fixed the issue by focusing on numeric and string types.

### DIFF
--- a/Engine/TestStepRun.cs
+++ b/Engine/TestStepRun.cs
@@ -381,7 +381,9 @@ namespace OpenTap
             {
                 if (member.HasAttribute<ResultAttribute>())
                 {
-                    if (member.TypeDescriptor.IsPrimitive())
+                    var td = member.TypeDescriptor.AsTypeData();
+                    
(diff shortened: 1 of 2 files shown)


Engine/TestStepRun.cs
```diff
@@ -381,7 +381,9 @@ internal void AfterRun(ITestStep step)
             {
                 if (member.HasAttribute<ResultAttribute>())
                 {
-                    if (member.TypeDescriptor.IsPrimitive())
+                    var td = member.TypeDescriptor.AsTypeData();
+                    
+                    if (td != null && (td.IsNumeric || td.IsString))
```
alnlarsen, line 386: I'm not sure exactly what is included in `IsPrimitive` compared to `IsNumeric`. Is a boolean value numeric? 

To be safe, I think we should relax the check to allow all primitives in addition to strings.

alnlarsen (changes requested):
See my comment

PR #1287, opened 2023-10-10
Update from Ubuntu 20.04 base to 22.04 base

alnlarsen:


rmadsen-ks:
We dont need to do this at the moment. The security updates should come automatically.

PR #1288, opened 2023-10-10
1278 Parameters Race Condition fix

### LINKED ISSUE #1278: MembersAnnotation throwing InvalidOperationException
Since upgrading from 9.21 to 9.22 RC, I've got unittests that intermittently start to fail:

image

I believe this "Collection was modified; enumeration operation may not execute." exception has happened before, but was fixed in an earlier version.

This is the code that throws the exception (line 138):
image

rmadsen-ks:
Made parameter member data less prone to data races by making the members themselves immutable. This means one can iterate through the parameters while another thread updates them.

There is also a potential for dataloss if two objects add members at the same time. But lets assume that this will never happen. Worst case the user will have to add the parameters twice.

Close #1278


alnlarsen (approved):
I like this :)

Issue #1292, opened 2023-10-11
Error when loading a test plan with a mixin as a test plan parameter


Razvan-Muresan:
**Environment (please complete the following information):**

 - OS: Windows 10
 - Test Automation 9.22.0-rc.3

**To Reproduce**

1. Add any step
2. Open test plan parameters
3. Add a mixin
4. Save the test plan
5. Reload the test plan

**Actual behavior**

Error in log [Latest (1).txt](https://github.com/opentap/opentap/files/12867854/Latest.1.txt)

![image](https://github.com/opentap/opentap/assets/132340651/38db5d99-6d87-42db-854f-b6ff227c3188)

rmadsen-ks:
This was closed by #1295.

Issue #1291, opened 2023-10-11
Parameterized mixin remains after removing the mixin itself


Razvan-Muresan:
**Environment (please complete the following information):**

 - OS: Windows 10, 64-bit

 - Test Automation 9.22.0-rc.3

**To Reproduce**

1. Add a mixin
2.  Parameterize on Test Plan
3. Remove mixin

**Actual behavior**

Mixin was removed but it's still a parameter

**Expected behavior**

Parameterized mixin should be removed

![parammix](https://github.com/opentap/opentap/assets/132340651/ff66de42-a1e9-4dba-a962-56176d0ce7a4)

rmadsen-ks:
This was fixed by #1295

PR #1295, opened 2023-10-11
1293 TestPlanReference With Mixins

### LINKED ISSUE #1293: Test Plan Reference: Can add/edit/remove expressions and mixins
**Environment:**

Windows x, y-bit

Test Automation 9.22.0-rc.3

**Steps to reproduce:**

1. Add a Delay step
2. Add an expression to Time Delay
3. Load the test plan into a Test Plan Reference
4. Change Time Delay value on loaded test plan

**Actual result:**

Expression on loaded Delay step can be edited

**Expected result:**

Test plan loaded in Test Plan Reference should be read only

**Other remarks:**

rmadsen-ks:
Fixed a bunch of issues with mixins and parameters.

Also added unit test for each case.

Close #1293, #1292, #1291

### DIFF
--- a/BasicSteps/TestPlanReference.cs
+++ b/BasicSteps/TestPlanReference.cs
@@ -214,7 +214,14 @@ namespace OpenTap.Plugins.BasicSteps
                 MaxNumberOfElements = 100
             };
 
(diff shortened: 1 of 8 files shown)


alnlarsen (approved):
Looks good

PR #1298, opened 2023-10-12
1297 multi select and multiple enabled annotations

### LINKED ISSUE #1297: Multi-select and multiple IEnabledAnnotations are not compatible
This causes things to be enabled that should not be.

To verify this, assign an expression to the same property on two test steps and try to multi select it. Normally this causes the property to be disabled:

image

Then try to do the same with the Timeout property of a dialog step (it is enabled by a bool):

image
(It should have been disabled)

rmadsen-ks:
Fixed the issue by modifying some minor behavior in multi-select (many-to-one) annotation.

Close #1297


alnlarsen:
Should this be linked to an issue?

alnlarsen (approved):
Looks good. I linked the issue which I assume this is related to

PR #1314, opened 2023-10-17
1313 artifact position bug

### LINKED ISSUE #1313: Artifacts from Stream should not need to use Position
Streams from a HttpClient does not support getting Position. This causes an exception inside artifact handling.

rmadsen-ks:
Close #1313

- Removed the use of position from teeStream since this is not always supported.
- Publish artifact is blocking unless the Async versions gets used.
- ArtifactZipper now needs use the async call. Otherwise a deadlock occurs.
- Added HttpArtifactStep as a test


alnlarsen (approved):
Let's merge

PR #1323, opened 2023-10-23
1322 Missing Enabled Rule For TestPlan Context

### LINKED ISSUE #1322: Mixin can be added while a test plan is running
**Describe the bug**
The user have the possibility to add a mixin while a test plan is running. However the mixin can't be edited or removed while the test plan is running

 - OS:  Windows 10
 - OpenTAP Version 9.22.2+1f88c8b1

**To Reproduce**

1. Add a Dialog step and run the test plan, do not press anything on the Dialog box
2. Add a mixin while test plan is running

**Actual behavior**
The user should is able to add a mixin while a test plan is running

**Expected behavior**
The user should not be able to add a mixin while a test plan is running

**Additional context**
mixin while running

rmadsen-ks:
AddMixin never got disabled for the test plan because the code did not support it directly on the test plan.

Also added a unit test which verifies that the enabled/disabled rule works for AddMixin in general.

Note, there was also another part to issue #1322, but that was fixed in KS8400 itself.

Close #1322

### DIFF
--- a/Engine.UnitTests/MixinTests.cs
+++ b/Engine.UnitTests/MixinTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
+using System.Threading;
 using NUnit.Framework;
 using OpenTap.Plugins.BasicSteps;
 namespace OpenTap.UnitTests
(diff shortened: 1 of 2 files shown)


Engine.UnitTests/MixinTests.cs
```diff
@@ -260,7 +261,50 @@ public void TestParameterizeAndRemoveMixin()
...
+            var planRun = plan.ExecuteAsync();
+            started.WaitOne();
+            // ok, now the test plan is waiting until we set continueEvt.
 
+            var a = AnnotationCollection.Annotate(step);
+            var addMixin = a.GetIcon(IconNames.AddMixin);
+            Assert.IsNotNull(addMixin);
+            var enabled = addMixin.GetAll<IEnabledAnnotation>();
+            Assert.IsTrue(enabled.Any());
+            var isDisabled = enabled.Any(x => x.IsEnabled == false);
```
alnlarsen, line 301:
```suggestion
            var isDisabled = enabled.All(x => x.IsEnabled == false);
```

Engine.UnitTests/MixinTests.cs
```diff
@@ -260,7 +261,50 @@ public void TestParameterizeAndRemoveMixin()
...
+            Assert.IsTrue(isDisabled);
+            continueEvt.Set();
+            planRun.Wait();
+            a.Read();
+            var isDisabled2 = enabled.Any(x => x.IsEnabled == false);
```
alnlarsen, line 306:
```suggestion
            var isDisabled2 = enabled.All(x => x.IsEnabled == false);
```

alnlarsen (changes requested):
Minor nitpicks. Otherwise this looks good

Engine.UnitTests/MixinTests.cs
rmadsen-ks: Actually it is disabled if any one of these are false, so the Any is more correct.

alnlarsen (approved):
Looks good!

PR #1325, opened 2023-10-24
1324 Improve Result listener example.

rmadsen-ks:
Improved the result listener example by cleaning it a bit up and adding an example of how to get the TestStepRun from OnResultPublished.


alnlarsen (approved):
Looks fine

PR #1332, opened 2023-11-02
1329 Modify Mixin Not Working Fix

### LINKED ISSUE #1329: "Modify Mixin..." context menu is missing from the first step in the TestPlan
Steps to reproduce:

- Add any teststep
- Add mixin to the teststep
- Right click mixin
- Observe that "Modify Mixin..." is missing.

This is only the case for the first teststep in the testplan.

rmadsen-ks:
This is hard to reproduce without installing the Expressions package.

The expressions wants to hide modify mixin if an expression is already assigned to that member. To do this, it has to get the member of the mixin member in the menu model. In this case the member was null, which meant that the expressions plugin just checked if any property of the step had an expression rather than the specific property.

Close #1329

### DIFF
--- a/Engine.UnitTests/MixinTests.cs
+++ b/Engine.UnitTests/MixinTests.cs
@@ -307,6 +307,27 @@ namespace OpenTap.UnitTests
             Assert.IsFalse(isDisabled2);
 
         }
+
+
+        [Test]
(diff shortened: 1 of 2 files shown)


Denrask (approved):
Tested

PR #1333, opened 2023-11-03
1331 ChildStps Serialized as guids

rmadsen-ks:
Added a slight more safe way to check that the currently serializing type is a TestStepList.


alnlarsen (approved):
Looks good 🤘

PR #1343, opened 2023-11-13
Added documentation about mixins

rmadsen-ks:


doc/Developer Guide/Test Step/Readme.md
```diff
@@ -627,3 +627,103 @@ Remember, adhering to the correct syntax is crucial for the expressions to evalu
 
    Example: `The number is {1 + 2}.` becomes "The number is 3."
 
+## Mixins
+
+Mixins, short for "mix-ins," are small units of functionality that can be integrated or 'mixed in' with an object.
```
bdirenzo-keysight, line 632: I think this uses the word MixIn to describe what a MixIn is too frequently. Maybe something like. 
The term MixIn comes from object-oriented programming where a mix-in is a class that contains extensions for use by other classes.

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -627,3 +627,103 @@ Remember, adhering to the correct syntax is crucial for the expressions to evalu
 
    Example: `The number is {1 + 2}.` becomes "The number is 3."
 
+## Mixins
+
+Mixins, short for "mix-ins," are small units of functionality that can be integrated or 'mixed in' with an object. 
+In the context of OpenTAP, mixins are can be used for extending the capabilities of a test step.
```
bdirenzo-keysight, line 633: Don't we also support DUTs, Instruments, and Result Listeners? At least the prompt comes up for those

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -627,3 +627,103 @@ Remember, adhering to the correct syntax is crucial for the expressions to evalu
...
+
+### Getting Started
+
+By default, OpenTAP contains no mixins, but they can be added with packages. Notable packages includes:
+- Expressions: Adds Number and Text mixins.
+- Basic Mixins: Contains a number of basic mixins. For example the Repeat mixins, which enables a test step to repeat itself. Or Artifact which allows adding a file path pointing to a file that should be considered an artifact of the step.
```
bdirenzo-keysight, line 639: And Limits to apply a limit check based on a value calculated in the test step

doc/Developer Guide/Test Step/Readme.md
```diff
@@ -627,3 +627,103 @@ Remember, adhering to the correct syntax is crucial for the expressions to evalu
...
+
+### Adding Mixins as an End User
+
+The mixins are available through the settings context menu. Right-click a setting or in the settings area and click "Add Mixin".
```
bdirenzo-keysight, line 643: Should this part (+ some examples/screenshots) be in the 8400 documentation instead?

bdirenzo-keysight (changes requested):

doc/Developer Guide/Test Step/Readme.md
rmadsen-ks: Ill add test step as an example. We still dont have any mixins for those types.

doc/Developer Guide/Test Step/Readme.md
rmadsen-ks: I think it is generic enough to have as a paragraph here. I should have more info in the KS8400 documentation.

PR #1345, opened 2023-11-13
1344 Global Mutex Fix

### LINKED ISSUE #1344: LockingPackageAction should use a global mutex
On Windows, a global mutex should be prefixed with Global\, but this is not the case in the Win32FileLock.

This causes issues if two different users tries to install packages at the same time.

rmadsen-ks:
Prefixed the file lock with Global\ to make it a global mutex.

Close #1344

### DIFF
--- a/Shared/FileLocks.cs
+++ b/Shared/FileLocks.cs
@@ -191,7 +191,8 @@ namespace OpenTap
         public Win32FileLock(string name)
         {
             // Having backslashes in the mutex name seems to cause issues for some reason. Replace them with slashes.
-            _mutex = new Mutex(false, name.Replace("\\", "/") + "_opentap_named_mutex_");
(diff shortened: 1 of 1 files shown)


Denrask:
Hi @rmadsen-ks ! Thanks for doing this one! I did some testing and came to an interesting result:

I had a tap process holding the global mutex. My process was successfully able to wait for this using the mutex without the "global" prefix!

Furthermore, while my tap process was holding the global mutex, I tried calling a lockingpackageaction from another user:

image

?!?!?

rmadsen-ks:
After discussing this with Dennis we hve decided not to implement this.

The Global namespace causes other issues related to mutex access which is not available in .net standard. We can look into it again when we have .net 8 support.

PR #1335, opened 2023-11-07
Fix Python plugins not added to package xml

alnlarsen:
This feature causes plugins added from e.g. a Python file with the Python plugin (or any ITypeDataSource implementations) to be added to the <File> section in the package.xml of a plugin

PR #1350, opened 2023-11-16
1349 plugin type selector mixin

### LINKED ISSUE #1349: Use PluginTypeSelectorAttribute for Mixin selector
This would enable OpenTAP  to know that this can be extended with more plugins.

To do this we need to add an ItemSourceProperty to the PluginTypeSelectorAttribute.

rmadsen-ks:
Close #1349 

This PR also includes the changes from #1335

## Testing
1. Verify that Engine Settings / Resource Strategy behave like in the previous version.
2. Mixins (select type, edit, add, remove) behave like before with regards to naming and functionality.

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -1209,6 +1209,17 @@ namespace OpenTap
             this.AvailableMember = annotation;
         }
     }
+    
+    /// <summary>
+    /// This is interface is used for generating something from a string.
+    /// Currently it is used only for PluginTypeSelectorAttribute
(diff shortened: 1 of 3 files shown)


rmadsen-ks (commented):
We need to verify that EngineSettings dropdown has not changed due to this.

alnlarsen (commented):
I don't see anything alarming about these changes, but I'm unsure how to test it.

alnlarsen:
Engine settings are not affected by this change 👍

alnlarsen (approved):
Let's merge

PR #1354, opened 2023-11-20
1353 macos arm64 nuget

rmadsen-ks:


rmadsen-ks:
@alnlarsen, what do you think about this?

alnlarsen:
@rmadsen-ks did you test this? I think we also need to make some changes to `nuget/build/OpenTap.targets` so the new payload gets copied instead of the `linux-x64` payload.

StefanHolst (approved):
Looks good

PR #1356, opened 2023-11-20
Package self-depencency fix

AsgerIversen:
Added a check to in the non-dotnet dependency adding code of tap package create to prevent a package from depending on itself.


alan-copeland-keysight:
Self dependency is resolved! package.xml can be used both at the root folder and embedded at Packages\<package>\package.xml

alnlarsen (approved):
Looks good

PR #1357, opened 2023-11-21
Don't recurse any further when base type is null in addTypeDataDepend…

AsgerIversen:
…encies.


alnlarsen (approved):
Works, let's get this in

PR #1359, opened 2023-11-21
Check if a typedata source was already handled

### LINKED ISSUE #1358: Duplicate plugin types added to package xml
If multiple there are multiple equal ITypeDataSource instances for the same file, their plugins will be duplicated.

alnlarsen:
Closes #1358


AsgerIversen:
Thanks @alnlarsen I did one as well here: https://github.com/opentap/opentap/pull/1360

alnlarsen:
@AsgerIversen let's use your solution

PR #1360, opened 2023-11-21
Distinct list of plugin sources to avoid duplicate plugins in package

AsgerIversen:
This could happen if a python file defines several plugins, then the sourceList would contain the PythonTypeDataSource several times...


alnlarsen (approved):
Works, let's merge

PR #1361, opened 2023-11-21
Another fix for python plugin info in package xml

AsgerIversen:
Make sure to filter out abstract types early, so we don't erroneously pick a provider that can only provide those.


alan-copeland-keysight:
Looks good. PowerSupply, Operator UI and PythonExamples all build correctly with metadata

PR #1366, opened 2023-12-05
Remove Package step from workflow

creotronics:
This step is not needed anymore because the package is already created during the Build process.

At least if next is added in the VS project file (which by default is if I'm correct):
```
<PropertyGroup Condition="'$(Configuration)' == 'Release'">
    <CreateOpenTapPackage>true</CreateOpenTapPackage>
</PropertyGroup>
```

### DIFF
--- a/doc/Developer Guide/Continuous Integration/Readme.md
+++ b/doc/Developer Guide/Continuous Integration/Readme.md
@@ -151,10 +151,6 @@ jobs:
       # Build your project
       - name: Build
         run: dotnet build -c Release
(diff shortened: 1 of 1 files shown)


rmadsen-ks (changes requested):
You are right, although that assumes a specific configuration of the .NET project. 

Maybe we should comment it out instead of removing it and then add comment that the default project configuration builds the TapPackage?

rmadsen-ks (approved):
Looks good

PR #1373, opened 2024-01-04
1372 modify mixin bug

### LINKED ISSUE #1372: Mixin properties are not updated after modifying it via "Modfiy Mixin" menu item.
.**Describe the bug**
Mixin properties not updated after modifying it. 

**Environment:**

 - OS: Windows 10
 - OpenTAP Version: 9.22.3

**To Reproduce**

1. Add a mixin to a test step

2. Change one of its properties using "Modfiy Mixin" menu item

3. Observe that mixin's property is not updated after modifying it. 

**Actual behavior**
When a property is changed from the MixinBuilderUI, mixin's property is not updated, See image below:

image

**Expected behavior**

Property added via a mixin must be updated after modifying the mixin.

**Additional context**
You can access OpenTAP forum topic from [here](https://forum.opentap.io/t/enabledif-attribute-in-mixins/1333)

rmadsen-ks:
Close #1372


alnlarsen (approved):
I retract my statement. Everything seems to work with Expression Mixin, but there seems to be a separate issue with Limit Mixin from Basic Mixins, though that could be an issue with the specific implementation.

PR #1381, opened 2024-01-16
Short Lived Connections test

rmadsen-ks:
Added unit test that resouces are not connected at the same time when using LazyResourceManager (Short Lived Connections)


alnlarsen (approved):
Awesome!

PR #1382, opened 2024-01-18
Skip tap.exe when verifying OpenTAP

### LINKED ISSUE #1252: 'Tap package verify' command fails for OpenTAP if it was installed from the CLI
**Describe the bug**

Looking at these fixed scenarios https://github.com/opentap/opentap/pull/1238, it works for OpenTAP installed using the installer but not when OpenTAP is installed using the CLI

 - OS: Windows 10
 - OpenTAP Version 9.22.0-beta.43

**To Reproduce**

1. Install OpenTAP using the installer
2. Run `tap package verify` - > OpenTAP is verified
3. Run `tap package install OpenTAP --version beta` - > This will reinstall the same version of OpenTAP in this case
4.  Run `tap package verify` - > OpenTAP is not verified

**Actual behavior**

image

**Expected behavior**

OpenTAP should be verified in this scenario too

**Additional context**

Log: [Latest.txt](https://github.com/opentap/opentap/files/12725429/Latest.txt)

alnlarsen:
Closes #1252

### DIFF
--- a/Package/CreatePackage/FileHashPackageAction.cs
+++ b/Package/CreatePackage/FileHashPackageAction.cs
@@ -123,6 +123,10 @@ namespace OpenTap.Package
             var brokenFiles = new List<(PackageFile, string)>();
             foreach (var file in pkg.Files)
             {
+                var fn = Path.GetFileName(file.FileName);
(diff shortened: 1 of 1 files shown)


rmadsen-ks (approved):
Makes sense!

PR #1384, opened 2024-01-18
Count actual number of commits since version change

### LINKED ISSUE #1321: Gitversion Macro incorrect calculation of precedence
**Describe the bug**

I’ve been experiencing the Gitversion macro calculating a number that is lower precedence than a version calculated in a previous commit.

[[ information omitted ]]
My expectation would be that the Gitversion precedence would be 133 < 134 < 135.  But what actually happened was 134 < 135 < 133

 - OS: Windows 10
 - OpenTAP Version 9.21.1

**To Reproduce**

1. Make a commit on branch A
2. make a new branch (B) off of branch A, make several commits on branch B
3. make another commit on branch A
4. Merge branch A into branch B
5. Merge branch B into branch A

**Actual behavior**

The Gitversion calculated on branch A of step 5 will be a smaller number (less precedence) than the Gitversion calculated on branch A of step 4

**Expected behavior**

The Gitversion calculated on branch A of step 5 should be bigger number (higher precedence) than the Gitversion calculated on branch A of step 4

alnlarsen:
This solves an issue where the version number of a branch can be reset in some circumstances after a merge. This can happen in the following scenario:

1. A feature branch is created branching off from the current main branch
2. Some some commits are added to main
3. Some commits are added to feature branch
4. main is merged into the feature branch with a merge commit (`git merge --no-ff`)
5. The feature branch is fast-forwarded into main (`git merge --ff-only`)

Because the merge commit from merging main into the feature branch was fast forwarded back onto main without creating a new merge commit, the gitversion calculation would now skip all commits from between that merge commit, and the first parent of that merge commit. This would cause the prerelease counter to potentially rewind.

Instead, with this commit, we count the actual number of commits. This has a couple of side effects. For example, when a feature branch is merged into main with a merge commit, the prerelease counter of the main branch will increment once for each commit that was merged, and once for the merge commit. Previously, the prerelease counter would have only incremented once.

There is no way for us to detect if a problematic merge commit was fast forwarded onto main because a merge commit only contains the following information:
1. The head of the branch being merged
2. The head of the branch being merged into

In other words, there is no way to know what branches are actually being merged. Only what the head of each branch was when the merge commit was created.

This is a bit unfortunate, but unavoidable. But at least the prerelease counter is now guaranteed to be monotonically increasing, except when force pushing to a detached head.

See the changes I made to the unit tests for more details about the expected behavior change

Closes #1321

### DIFF
--- a/Package.UnitTests/GitHelperTests.cs
+++ b/Package.UnitTests/GitHelperTests.cs
@@ -173,7 +173,7 @@ namespace OpenTap.Package.UnitTests
             // |/
             // *     (tag: v2.0, release2x) Bumped version to 2.0
             new GitVersionAction() { PrintLog = "10", RepoPath = repo.Info.WorkingDirectory }.Execute(new System.Threading.CancellationToken());
-            verifyVersion(repo, 2, 1, 0, "rc.3");
+            verifyVersion(repo, 2, 1, 0, "rc.2");
 
             // Now we will try to go back to a previous commit (not the HEAD):
             // This is what GitLab CI does, so important for us to support
--- a/Package.UnitTests/GitHelperTests.cs
+++ b/Package.UnitTests/GitHelperTests.cs
@@ -103,7 +103,7 @@ namespace OpenTap.Package.UnitTests
             Commands.Checkout(repo, "master");
             repo.Merge(featureBranch, me, noFastForward);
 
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
This looks good

PR #1385, opened 2024-01-18
Use spinning file lock on Windows

### LINKED ISSUE #1344: LockingPackageAction should use a global mutex

alnlarsen:
Global mutexes on Windows have issues on netstandard. Use a spinning file lock instead

Closes #1344


alnlarsen:
Rejecting this since some plugins depend on the current behavior

PR #1390, opened 2024-01-22
Update README.md to state VS 2022 requirement

### LINKED ISSUE #1346: OpenTAP fails to build locally
I just made a fresh clone of the main branch and attempted to build it using VS2019:

OpenTAP_Build_Fails

Looks like _evt_ being declared in the if statement it is not longer available when used in _SetElementValues_.

Possible fix:

```csharp
            if (obj is Event)
            {
                Event evt = (Event)obj;
                node.SetElementValue(nameof(evt.Timestamp), evt.Timestamp);
                node.SetElementValue(nameof(evt.Message), evt.Message);
                node.SetElementValue(nameof(evt.Source), evt.Source);
                node.SetElementValue(nameof(evt.EventType), evt.EventType);
                return true;
            }
            else
            {
                return false;
            }

Same issue if I take the source code from the zip file for the 9.22.3 release.

rmadsen-ks:
We no longer support building on VS 2017 since sinde 2022 is required for building .net 6 projects.

Close #1346

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For a guide on how to develop using OpenTAP, check out our __[Developer Guide](h
 Most users build plugins for OpenTAP but if you are interested in building OpenTAP yourself you can clone the git repository at https://github.com/opentap/opentap and build the OpenTAP.sln solution file.
 
 ### Microsoft Windows 10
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
Looks good

PR #1398, opened 2024-01-24
Added retry to SetAsmInfo

### LINKED ISSUE #1397: Retry SetAsmInfo
In some cases SetAsmInfo can fail. This can be due to some antivirus software locking the file while it is doing a scan. We could consider retrying this in a loop a given number of times.

rmadsen-ks:
Close #1397

### DIFF
--- a/Shared/ReflectionHelper.cs
+++ b/Shared/ReflectionHelper.cs
@@ -15,6 +15,7 @@ using System.Runtime.InteropServices;
 using System.IO.MemoryMappedFiles;
 using System.Threading.Tasks;
 using System.Runtime.CompilerServices;
+using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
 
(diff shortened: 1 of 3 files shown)


Shared/ReflectionHelper.cs
```diff
@@ -1629,6 +1630,51 @@ public static string BytesToReadable(long bytes)
...
+                return 0;
+            }, retryOn, maxRetries, moreExceptions, sleepBaseMs);
+        }
+
+        public static T Retry<T>(Func<T> func, Type retryOn = null, int maxRetries = 5, Type[] moreExceptions = null, int sleepBaseMs = 100)
+        {
+            if (retryOn == null)
+                retryOn = typeof(Exception);
+            var allExceptions = new[] { retryOn }.Concat(moreExceptions ?? Array.Empty<Type>()).ToArray();
+            int retryCount = 0;
```
alnlarsen, line 1655: Can we use a for loop instead?

Shared/ReflectionHelper.cs
```diff
@@ -1629,6 +1630,51 @@ public static string BytesToReadable(long bytes)
...
+                }
+                var exType = ex.GetType();
+                if (retryCount < maxRetries && allExceptions.Any(x => x.IsAssignableFrom(exType)))
+                {
+                    retryCount++;
+                    Backoff(retryCount, sleepBaseMs);
+                    goto retryPoint;
+                }
+                ExceptionDispatchInfo.Capture(ex).Throw();
+                throw;
```
alnlarsen, line 1675: Why two throws?
Also, we already have an `ex.Rethrow()` extension method I think

alnlarsen (changes requested):

alnlarsen (dismissed):
Looks good!

alnlarsen (approved):
Looks great!

PR #1402, opened 2024-01-25
Tried to improve the isolation server for isolated package actions.

rmadsen-ks:


Shared/ExecutorInterop.cs
```diff
@@ -44,43 +44,22 @@ public ExecutorSubProcess(ProcessStartInfo start)
...
         }
 
-        void pipeConnected()
+        void ProcessPipe()
         {
-            
             byte[] buffer = new byte[1024];
 
-            void readMessage()
+            while (true)
```
alnlarsen, line 51: Maybe use `while (!tokenSource.Token.IsCancellationRequested)` here?

alnlarsen (changes requested):
This is an awesome refactor! I only have one small remark

alnlarsen:
Not needed. We no longer use isolation.

PR #1405, opened 2024-01-26
1400 fix multiple test plan reference giving duplicate warnings

### LINKED ISSUE #1400: Using two levels of test plan reference leads to duplicate ID warnings
**Describe the bug**

Windows 10, 64-bit
Test Automation 9.19.0 (reproduced on 9.18.4)

**To Reproduce**

1. Add a test plan reference step
2. Load this test plan [TR_Level1.txt](https://github.com/opentap/opentap/files/14053580/TR_Level1.txt)
3. Save the test plan
4. Load the saved test plan

**Actual behavior**

Warnings in the log [SessionLog_2022-11-08_09-07-53.txt](https://github.com/opentap/opentap/files/14053582/SessionLog_2022-11-08_09-07-53.txt)

image

**Expected behavior**

**Additional context**

Add any other context about the problem here. Please also attach a log file from <InstallDir>\SessionLogs\ that shows the issue when possible/relevant.

alnlarsen:
This demotes warnings about duplicate step ids if the offending steps are from two different TestPlanReference steps.

Closes #1400

### DIFF
--- a/Engine/SerializerPlugins/TestStepSerializer.cs
+++ b/Engine/SerializerPlugins/TestStepSerializer.cs
@@ -56,10 +56,19 @@ namespace OpenTap.Plugins
         {
             if (stepLookup.TryGetValue(step.Id, out ITestStep currentStep) && currentStep != step && !ignoredGuids.Contains(step.Id))
             {
+                ITestStepParent getTestPlanReferenceParent(ITestStep s) =>
+                    s.GetParents().FirstOrDefault(p => p.GetType().Name == "TestPlanReference");
+
                 step.Id = Guid.NewGuid();
                 if (step is IDynamicStep)
                 {   // if newStep is an IDynamicStep, we just print in debug.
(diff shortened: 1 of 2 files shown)


Engine/SerializerPlugins/TestStepSerializer.cs
```diff
@@ -56,11 +56,20 @@ public void FixupStep(ITestStep step, bool recurse)
         {
             if (stepLookup.TryGetValue(step.Id, out ITestStep currentStep) && currentStep != step && !ignoredGuids.Contains(step.Id))
             {
+                ITestStepParent getTestPlanReferenceParent(ITestStep s) =>
+                    s.GetParents().FirstOrDefault(p => p.GetType().Name == "TestPlanReference");
```
rmadsen-ks, line 60: Maybe we should add an interface type for steps that load their own child steps? That way we won't have to do this string comparison.

rmadsen-ks (changes requested):

rmadsen-ks (approved):
Looks good!

PR #1406, opened 2024-01-29
Allow setting repo token from CLI

### LINKED ISSUE #1383: Allow setting a repository token from the CLI
We support adding repository authentication in `AuthenticationSettings.xml`, but there is no simple way to specify a token on the CLI when using e.g. `tap package download` or `tap package install`. We should support e.g. `--token <user_token>`. This will enable users to easily install non-public packages uploaded to `packages.opentap.io`.

alnlarsen:
This adds the option to specify repository tokens from the CLI

Closes #1383

### DIFF
--- a/Package.UnitTests/PackageActionTest.cs
+++ b/Package.UnitTests/PackageActionTest.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Reflection;
 using System.Threading;
 using NUnit.Framework;
+using OpenTap.Authentication;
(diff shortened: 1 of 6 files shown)


rmadsen-ks (changes requested):
Should we  change this so that there is no extra command line argument, but instead the repostitory string has an additional syntax. e.g:

```--repository "https://example.com;token=myToken1234567890"```

rmadsen-ks (approved):
Seems to work

PR #1412, opened 2024-01-31
The current read implementation is needlessly complicated. This seems better

### LINKED ISSUE #1415: Clean up IEEE Block related code in ScpiInstrument
The code in ScpiInstrument could be made simpler.

alnlarsen:
This has not been tested, and we have no regression tests interfacing with instruments. Before we can merge this we need to do a manual test with a real instrument.

Close #1415


rmadsen-ks (approved):
Looks great!

PR #1414, opened 2024-01-31
1411 Parameterization Breaks After PluginManager search: Fix

### LINKED ISSUE #1411: Parameterizations breaks after PluginManager.Search
After plugin search external parameters breaks. 

This is probably due to cache invalidation and that every type gets new objects.

Reproduced by:

```cs
            var plan = new TestPlan();
            var delay = new DelayStep();
            plan.ChildTestSteps.Add(delay);
            
            TypeData.GetTypeData(delay).GetMember(nameof(DelayStep.DelaySecs))
                .Parameterize(plan, delay, nameof(DelayStep.DelaySecs));
            
            var xmlPreSearch = plan.SerializeToString();
            
            // searching for plugins should not affect serialization.
            PluginManager.Search();
            var xmlPostSearch = plan.SerializeToString();
            
            Assert.AreEqual(xmlPreSearch, xmlPostSearch);

```

rmadsen-ks:
Fixed the issue by implementing Equality for MemberData. This is required for a lookup table inside the ExternalParameterSerializer. Otherwise when serialization is done it will not recognize the new MemberData instance as the same one as in the parameter.

Added a unit test to verify the fix.

Close #1411

### DIFF
--- a/Engine.UnitTests/ExternalTestPlanParameterTest.cs
+++ b/Engine.UnitTests/ExternalTestPlanParameterTest.cs
@@ -494,5 +494,24 @@ namespace OpenTap.Engine.UnitTests
                 Assert.AreEqual(access.IsVisible, stepModel.Length == 1);
             }            
         }
+        
+        [Test]
+        public void ExternalParametersPlusReloadPlugins()
+        {
+            var plan = new TestPlan();
+            var delay = new DelayStep();
+            plan.ChildTestSteps.Add(delay);
+            
+            var preMember = TypeData.GetTypeData(delay).GetMember(nameof(DelayStep.DelaySecs))
+                .Parameterize(plan, delay, nameof(DelayStep.DelaySecs));
+            
+            var xmlPreSearch = plan.SerializeToString();
+            
+            // searching for plugins should not affect serialization.
+            PluginManager.Search();
+            var xmlPostSearch = plan.SerializeToString();
+            
+            Assert.AreEqual(xmlPreSearch, xmlPostSearch);
+        }
     }
 }
--- a/Engine/DynamicMemberTypeDataProvider.cs
+++ b/Engine/DynamicMemberTypeDataProvider.cs
@@ -467,6 +467,7 @@ namespace OpenTap
             if (member == null) throw new ArgumentNullException(nameof(member));
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (name == null) throw new ArgumentNullException(nameof(name));
+            if (source == target) throw new InvalidOperationException("Source and target cannot be the same object.");
             if (name.Length == 0) throw new ArgumentException("Cannot be an empty string.", nameof(name));
             { // Verify that the member belongs to the type.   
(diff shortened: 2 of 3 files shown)


Engine.UnitTests/ExternalTestPlanParameterTest.cs
```diff
@@ -494,5 +494,26 @@ public void MultiSelectEditParametersDisabled()
...
+            
+            var preMember = TypeData.GetTypeData(delay).GetMember(nameof(DelayStep.DelaySecs))
+                .Parameterize(plan, delay, nameof(DelayStep.DelaySecs));
+            
+            var xmlPreSearch = plan.SerializeToString();
+            
+            // searching for plugins should not affect serialization.
+            PluginManager.Search();
+            var xmlPostSearch = plan.SerializeToString();
+            TypeData.GetTypeData(plan).GetMember(preMember.Name);
```
alnlarsen, line 513: Does this lookup serve a purpose?

Engine/Reflection/DotNetTypeData.cs
```diff
@@ -184,5 +184,18 @@ ITypeData getTypeDescriptor()
...
+        public override bool Equals(object obj)
+        {
+            if (obj is MemberData m && m.Member == Member)
+                return true;
+            return false;
+        }
+        /// <summary> GetHash for MemberData. </summary>
+        public override int GetHashCode()
+        {
+            return Member.GetHashCode() * 32143211;
```
alnlarsen, line 198: What is the point of this multiplication?

Engine/DynamicMemberTypeDataProvider.cs
```diff
@@ -467,6 +467,7 @@ internal static ParameterMemberData ParameterizeMember(object target, IMemberDat
             if (member == null) throw new ArgumentNullException(nameof(member));
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (name == null) throw new ArgumentNullException(nameof(name));
+            if (source == target) throw new InvalidOperationException("Source and target cannot be the same object.");
```
alnlarsen, line 470: This fix looks unrelated to this issue. Did you encounter this scenario while working on this?

alnlarsen (changes requested):
I think this looks like a good solution, but I have a few questions

Engine/DynamicMemberTypeDataProvider.cs
rmadsen-ks: This happened when I tried to implement the test. If the check is not done a stack overflow exception will occur.

Engine/Reflection/DotNetTypeData.cs
rmadsen-ks: I scramble the hash to reduce the risk of collisions. Otherwise they'd have the same hash value as the Member and this would not be every "random".

rmadsen-ks: It's not super important. If there were more things being hashed together it would be.

alnlarsen (approved):
Looks great!

PR #1417, opened 2024-02-01
Added "disabled" annotation when parent child step lists are readonly

### LINKED ISSUE #1352: Properties of TestSteps added by TestPlanReference is not readonly in Annotations
**Describe the bug**

The Runner and TUI does not display properties of ReadOnly TestSteps as Readonly.

**To Reproduce**

1. Create a TestPlan containing a Delay step
2. Load that plan using a TestPlanReference Step
3. The "Time Delay" property is not readonly from an Annotation perspective

rmadsen-ks:
Added unit test to verify the behavior.

Close #1352

- I tested this on KS8400 by removing the special annotation there that made steps settings disabled.
- And tested using TUI.

### DIFF
--- a/Engine.UnitTests/AnnotationTest.cs
+++ b/Engine.UnitTests/AnnotationTest.cs
@@ -2347,5 +2347,34 @@ namespace OpenTap.UnitTests
                 }
             }
         }
+
+        [AllowAnyChild]
+        public class ReadOnlyDelays : TestStep
+        {
+            public ReadOnlyDelays()
+            {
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
This seems to work really well, but I discovered another problem while testing this:

It is possible to add child test steps to `ChildTestSteps` using `ChildTestSteps.Add(step)` with the C# API, even when `ChildTestSteps` is readonly. This is possible in e.g. TUI. 

Not really related to this issue. I will open a separate issue.

PR #1421, opened 2024-02-01
1416 fix flaky tests

### LINKED ISSUE #1416: Flaky unit test improvements
- Engine test times out intermittently after 20 minutes during CI builds. We need to figure out which test causes this. Could it be related to user input
- CyclicDependenciesTest during test packages can also fail.
- Set the test timeout to 5 minutes(normally it takes around 2m30s for it to pass)

alnlarsen:
This fixes flaky unittests

1. The timeout for engine / package unittests was shortened significantly
2. A deadlock in Engine unittests was fixed
3. Minor cleanups

I was unable to reproduce the issue with `CyclicDependenciesTest`, even in the pipeline after repeating it over 100 times. It may have been fixed, or just hiding well. I rearranged the asserts so we should get logs if it fails again

I also looked for a way to get information from vstest/nunit about which test actually timed out, but it does not seem to be possible. However, since tests currently run sequentially, I was able to find the offending tests by cross referencing a successful run and checking the test after the final test in the run that timed out.

Closes #1416

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,8 +288,8 @@ jobs:
           path: bin/
       - name: Test
         run: |
-          $ErrorActionPreference = "Stop"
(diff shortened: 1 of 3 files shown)


rmadsen-ks (approved):
Looks great!

PR #1424, opened 2024-02-05
1423 make tap.sh portable

### LINKED ISSUE #1423: Make tap.sh shell independent
The current tap.sh script has a few issues:

1. It assumes bash is present (shebang on first line is #!/bin/bash). Should be sh instead
2. It has two execution paths: one for macos and one for everything else
3. It uses bash constructs (`[[ thing ]]` for test expressions), which is not necessary and not posix

The two execution paths assume that anything which isn't macos has GNU readlink. This may not be the case

I think we could both simplify and improve the script by using posix syntax, and dropping the dependency on readlink

alnlarsen:
This makes a couple of changes to make tap.sh portable

1. use sh in shebang
2. use realpath over readlink
3. stop using bash control flow constructs

Closes #1423

### DIFF
--- a/LinuxInstall/tap.sh
+++ b/LinuxInstall/tap.sh
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/sh
+
+# set -e: makes any non-zero exit code fatal
+set -e
 
 # umask sets the calling process' file mode creation mask
 # 002 is the default mask for users. The default mask for root is 0022.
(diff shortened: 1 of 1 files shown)


MikaelElkiaer (dismissed):
LGTM

rmadsen-ks (changes requested):
This does not work on mac. (realname)

rmadsen-ks (approved):
Looks great!

PR #1427, opened 2024-02-06
Incremental work on metrics

rmadsen-ks:


rmadsen-ks:
Improvements:
- Add an IMetricProducer interface, which is a class that generates metrics without being associated with any resource.
- Add support for ComponentSettings to publish metrics.
- Add runtime metrics for OpenTAP as an IMetricProducer.

PR #1432, opened 2024-02-14
Add hamfisted check to see if a repository exists

### LINKED ISSUE #1431: CLI repository determined incorrectly
When a base address is set (`AuthenticationSettings.xml`), and a repo is specified without a scheme on the CLI (`-r packages.opentap.io`), it will be determined as a relative URL. When no base address is set, it is correctly interpreted as a scheme-less URL

alnlarsen:
Closes #1431


rmadsen-ks (approved):
Looks good!

PR #1433, opened 2024-02-14
1430 instrument trailing space

alnlarsen:
This fixes an issue where a warning is given during serialization if a resource name has trailing whitespace.

This change should be safe since we are already trimming in the places where we do the actual lookup. The only place we don't trim is when we compare the names of what we found versus what we were looking for (and we also trimmed the value we were looking for)


rmadsen-ks (approved):
Looks great!

PR #1438, opened 2024-02-15
Disable gui sdk example in CE and add editorx launch options

### LINKED ISSUE #1437: Update SDK examples to use editor x and stop using editor CE
We should update our SDK examples to use Editor X (under active development) instead of Editor CE (not actively maintained).

alnlarsen:
This fixes a build error that occurred when selecting the Debug CE configuration. It also adds launch options for editor x

Closes #1437


rmadsen-ks (approved):
Looks great!

PR #1444, opened 2024-02-22
fix-filerepo-updatecheck

alnlarsen:
This fixes an issue which caused update checks to fail for local package repositories.


rmadsen-ks (approved):
Looks good!

PR #1468, opened 2024-03-14
Add currently installed packages to the list of additional packages in image resolver

### LINKED ISSUE #1469: Offline image resolution problems
When deploying an image into an installation which already has some package installed, the image resolution will fail if all of the following conditions are met:

1. Package `X` is specified in the image
2. Package `X` is already installed in the target installation
3. Package `X` is not available in any of the specified repositories

This will mostly happen in debug scenarios, but it can also be triggered when trying to build an OpenTAP plugin where the version of the OpenTAP NuGet package does not have the corresponding `OpenTAP.TapPackage` in the local package cache.

alnlarsen:
This fixes an issue where image resolutions will sometimes fail while offline, even if the required packages are available in the target installation.

Closes #1469

### DIFF
--- a/Package/Image/ImageSpecifier.cs
+++ b/Package/Image/ImageSpecifier.cs
@@ -49,11 +49,16 @@ namespace OpenTap.Package
         {
             var toInstall = new List<PackageSpecifier>();
             var installed = installation.GetPackages().ToList();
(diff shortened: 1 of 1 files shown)


rmadsen-ks (approved):
Nice!

PR #1470, opened 2024-03-15
1464 fix empty logs rollover

### LINKED ISSUE #1464: Empty logs while logs rollover
**Describe the bug**

An extra empty log is created

 - OS: Windows 10
 - OpenTAP Version 9.23.2

**To Reproduce**

1. Delete all logs in the SessionLogs folder
1. Run the attached test plan: [OneMaxLog.TapPlan.txt](https://github.com/opentap/opentap/files/14538061/OneMaxLog.TapPlan.txt)
2. Verify the SessionLogs folder

**Actual behavior**

First log, max size 994 MB
Second log _1 365MB
Third log _2 and Latest both same content at 1KB

image

**Expected behavior**

Thrid log _2 should not be created and Latest should be the same as the Second log _1

### LINKED ISSUE #1471: Log rollover does not work when producing data too fast
Since the `FileTraceListener` implementation uses an auto flushing `StreamWriter` with an underlying `FileStream`, session logs cannot roll if session log data is being produced faster than we can write to the underlying block device. See the test plan included in #1464 for an example of how to reproduce it.

This happens because `FileTraceListener` attempts to flush pending messages before detaching itself from the set of LogListeners, and then re-attaching a new log listener. I propose doing the renaming within `FileTraceListener` itself.

alnlarsen:
This fixes an issue causing session logs not to roll if messages are produced faster than FileTraceListener can flush.

This happens because `FileTraceListener` attempts to flush pending messages before detaching itself from the set of LogListeners, and then attaching a new log listener with a new file name.
Flushing the pending messages will never finish while new messages are being produced faster than the messages can be written to disk.

Moving the renaming logic into the FileTraceListener implementation was significantly simpler than any other solution I was able to come up with, and the previous implementation seemed a bit overengineered. 

This of course affects the session log, but also the test plan log. But the test plan log does not set a max file size, so the default size of `ulong.MaxValue` is used, so the renaming logic is unlikely to be triggered for test plans (in this universe).

Closes #1464 
Closes #1471

### DIFF
--- a/Engine/FileTraceListener.cs
+++ b/Engine/FileTraceListener.cs
@@ -33,8 +33,6 @@ namespace OpenTap
             _fileName = Path.GetFullPath(fileName);
         }
 
-        public event EventHandler FileSizeLimitReached;
-
         /// <summary> Installs a file limit. When the limit is reached FileSIzeLimitReached is invoked. </summary>
         public ulong FileSizeLimit = ulong.MaxValue;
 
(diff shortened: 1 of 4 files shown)


rmadsen-ks:
The issue was fixed by #1474

PR #1477, opened 2024-03-20
fixed a deadlock occuring when a lot of messages are logged when the …

rmadsen-ks:
…log file rolls over


alnlarsen (approved):
Verified the deadlock no longer occurs (on linux)

PR #1480, opened 2024-03-20
1117 add new run parameters

### LINKED ISSUE #1117: Add TestPlanRun / TestStepRun property: Status
Currently, there is no reliable way to check if a teststep or testplan ran to completion, was aborted due to a break condition, or was aborted by the user. I think we should add a public status Enum to stepruns / planruns indicating which of the following is the case:

1. Finished normally
2. Was cancelled by the operator
3. Was cancelled due to a break condition

This status should be written to results etc.

alnlarsen:
This adds some run parameters to provide additional details about a test plan run

1. CancelledDueToBreakCondition to indicate whether a test plan run finished normally, or was terminated early due to a break condition being satisfied
2. StepListOverridden to indicate that only a portion of the test plan was run

Closes #1117

### DIFF
--- a/Engine/TestPlanExecution.cs
+++ b/Engine/TestPlanExecution.cs
@@ -185,7 +185,7 @@ namespace OpenTap
             {
                 if (didBreak) return;
                 didBreak = true;
-                execStage.Parameters.Add(new ResultParameter(breakConditionParameterName, run.Id.ToString())); 
+                execStage.Parameters.Add(new ResultParameter(TestPlanRun.SpecialParameterNames.BreakIssuedFrom, run.Id.ToString())); 
             }
 
             try
@@ -605,7 +605,7 @@ namespace OpenTap
                     // The order of the guids does not really matter. 
                     // Only that the order is the same across runs
                     Array.Sort(overrides); 
-                    execStage.Parameters.Add(new ResultParameter(stepsOverrideParameterName, string.Join(",", overrides)));
+                    execStage.Parameters.Add(new ResultParameter(TestPlanRun.SpecialParameterNames.StepOverrideList, string.Join(",", overrides)));
                 }
 
                 execStage.Start();
@@ -719,10 +719,7 @@ namespace OpenTap
             return Execute(resultListeners, metaDataParameters, null);
         }
 
-        TestPlanRun currentExecutionState = null;
-        internal const string breakConditionParameterName = "BreakIssuedFrom";
-        internal const string stepsOverrideParameterName = "StepOverrideList";
-
+        TestPlanRun currentExecutionState = null; 
 
         /// <summary> true if the plan is in its open state. </summary>
         [AnnotationIgnore]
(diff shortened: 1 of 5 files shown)


Engine/TestPlanExecution.cs
```diff
@@ -589,6 +599,8 @@ private TestPlanRun DoExecute(IEnumerable<IResultListener> resultListeners, IEnu
             else
             {
                 execStage = new TestPlanRun(this, resultListeners.ToList(), initTime, initTimeStamp);
+                if (stepsOverride != null)
+                    execStage.Parameters.Add(new ResultParameter("StepListOverridden", "Yes"));
```
rmadsen-ks, line 603: Maybe this should be the list of guids in stepsOverride?

Engine/TestPlanExecution.cs
```diff
@@ -180,6 +180,14 @@ failState execTestPlan(TestPlanRun execStage, IList<ITestStep> steps)
             Stopwatch planRunOnlyTimer = Stopwatch.StartNew();
             var runs = new List<TestStepRun>();
 
+            bool didBreak = false;
+            void addBreakResult()
+            {
+                if (didBreak) return;
+                didBreak = true;
+                execStage.Parameters.Add(new ResultParameter("CancelledDueToBreakCondition", "Yes"));
```
rmadsen-ks, line 188: Lets move this property name to a static and internal variable in TestPlanRun. Also we could use the GUID of the step causing the plan to cancel?

rmadsen-ks (changes requested):
Looks good, but I had minor suggestions for improvements.

Engine/TestPlanExecution.cs
alnlarsen: I agree!

Engine/TestPlanExecution.cs
alnlarsen: Good call!

Engine/TestPlanExecution.cs
```diff
@@ -701,8 +720,10 @@ public TestPlanRun Execute(IEnumerable<IResultListener> resultListeners, IEnumer
         }
 
         TestPlanRun currentExecutionState = null;
+        private const string breakConditionParameterName = "BreakIssuedFrom";
```
rmadsen-ks, line 723: If you make these internal, we can use them in the unit tests too.

rmadsen-ks (changes requested):

Engine/TestPlanExecution.cs
```diff
@@ -701,8 +720,10 @@ public TestPlanRun Execute(IEnumerable<IResultListener> resultListeners, IEnumer
         }
 
         TestPlanRun currentExecutionState = null;
+        internal const string breakConditionParameterName = "BreakIssuedFrom";
```
rmadsen-ks, line 723: Lets move these to a class called SpecialParameterNames. And make them public.

rmadsen-ks (changes requested):
Lets make the names more easilty discoverable.

Also should we use spaces?

Engine/TestPlanExecution.cs
alnlarsen: Great idea. I have added it.

rmadsen-ks (approved):
Looks great!

PR #1484, opened 2024-03-26
1368 Output Type Compatibility: Fix

### LINKED ISSUE #1368: Allow Assignment of Outputs to Members with Compatible Types
At present, the assignment of outputs is restricted to members with identical types. It would be beneficial to extend this capability to include types that are compatible through casting. For instance:

**Scenario 1:**

Output Type: int
Input Type: double
Assignable: Yes

**Scenario 2:**

Output Type: float
Input Type: double
Assignable: Yes

**Scenario 3:**

Output Type: double
Input Type: int
Assignable: No

This enhancement would improve flexibility and usability by allowing assignments between types that can be safely casted, expanding the range of scenarios where outputs can be effectively assigned to members.

rmadsen-ks:
Allow outputs to be assigned to more types of inputs. Some conversion is done between them. Sometimes this conversion might fail. This should result in a runtime error.

Sub classes are allowed to be assigned to parent classes, but not the other way around.

Unit tests were added to verify the behavior.

Added documentation about inputs and outputs in general.

Close #1368, #1486


Engine/TestStep.cs
```diff
@@ -955,6 +963,9 @@ internal static TestStepRun DoRun(this ITestStep Step, TestPlanRun planRun, Test
                         TapThread.Current.AbortToken);
                     try
                     {
+                        if (readInputsError != null)
+                            throw readInputsError;
```
alnlarsen, line 967: The exception stack trace is overwritten here. Is this intentional? If so, please add a comment. Otherwise, let's re-throw it instead.

Engine/TestStep.cs
```diff
@@ -1064,6 +1075,7 @@ void completeAction(Task runTask)
                         }
 
                         planRun.AddTestStepRunCompleted(stepRun);
+                        stepRun.SignalCompleted();
```
alnlarsen, line 1078: Will the Completed event be set if this throws?

alnlarsen (changes requested):
Looks good! One minor concern

alnlarsen (approved):
Great! Thanks!

Engine/TestStep.cs
rmadsen-ks: It is not likely that an exception will be thrown.

PR #1489, opened 2024-03-27
Added Iteration output to repeat step.

### LINKED ISSUE #1455: Provide the iteration number in the Repeat Step
This information could be useful in child steps, so this should probably be an `[Output]`.

rmadsen-ks:
Close #1455


alnlarsen (approved):
Works, though it is a bit strange that `Current Iteration` uses the same display name. I tried using the new `Iteration` variable with the `Expression` plugin, and it uses this new uint value, and not the old `x / n` value.

PR #1491, opened 2024-04-03
Ensure innermost step id is used in break result

alnlarsen:
* Add a unit test verifying the break condition step id is propagated correctly
* Unwrap exceptions thrown in deferred runs
* Correctly propagate stepRun break exceptions


rmadsen-ks (approved):
Looks great!

PR #1495, opened 2024-04-05
Fix Input<> being cleared when moving steps around

### LINKED ISSUE #1456: Test Plan Editor: "If" parameter from "If Verdict" step is cleared after parent step is moved as child of another step
**Environment**
KS8400 9.22
KS8500 DEV 1.15

**Describe the bug**
"If" parameter from "If Verdict" step is cleared after parent step is moved as child of another step

**To Reproduce**

Add a Parallel test step
Add a Repeat test step
Add a Dialog step as a child of Repeat
Add an If Verdict step as a child of Repeat
Set "If" parameter to "Verdict from Dialog"
Move Repeat step as a child of Parallel

**Actual behavior**
The "If" parameter of "If Verdict" is cleared

**Screenshots**
if_verdict_step_if_parameter_is_empty_after_adding_parent_step_as_child_to_another_step

alnlarsen:
Add a lookup to check if the step id of an input is still in the test plan when.

Closes #1456

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
         password: ${{ secrets.GITLAB_REGISTRY_TOKEN}}
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - run: |
           mkdir Help API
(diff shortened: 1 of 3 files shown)


rmadsen-ks (approved):
Works beautifully!

rmadsen-ks (approved):
Looks great!

PR #1498, opened 2024-04-08
Adding instrument setting SendIDNOnConnect to ScpiInstrument

### LINKED ISSUE #1497: Allow ScpiInstrument to not send IDN on connect
I have had two main issues while sending *IDN? during base.Open():

1. Non compliant SCPI Instruments. For this case the instruments are old instruments and/or instruments that use RS-232. For this instruments the visa connection string is something like: ASRL1::INSTR and typically this instruments won't support *IDN?
2. Multiple connections to an instrument using the same GPIB Visa Address. In this case if there is an instrument like the DAQ970A which has the option to install multiple cards on the mainframe. The user would like to have an instrument for each one of the cards within the mainframe and address individually even though they share the same GPIB address.

For instruments that use a TCPIP device with a visa string like: TCPIP0::localhost::5025::SOCKET, there is no issue and there can be multiple instruments with the same visa address:

Instrument 1: PNA-X1
image

Instrument 2: PNA-X2
image

Running Test Plan that uses two instruments with the same VISA Address: TCPIP0::localhost::5025::SOCKET
image

If we want to repeat the same using an instrument with GPIB

Instrument SCPI 1: GPIB::5::INSTR

image

Instrument SCPI 2: GPIB::5::INSTR

image

Running same test plan that only queries *IDN?:

image

Adding an option to Send *IDN? on connect will help these two cases, similar to the Debug \\ Send VIClear On Connect option in the Scpi Instrument.

c-montes:
This allows to bypass sending IDN query on Open() for instruments that do not support and/or multiple GPIB instruments that have the same visa address string

This PR adds an option on the Debug section of the ScpiInstrument Settings, which is enabled by default:

image

If the user disables this option for two GPIB instruments:

image

image

After running a test plan that uses SCPI to query *IDN? on both GPIB instruments, the test plan now completes successfully:

image

Close #1497

### DIFF
--- a/Engine/ScpiInstrument.cs
+++ b/Engine/ScpiInstrument.cs
@@ -230,6 +230,12 @@ namespace OpenTap
         [Display("Send *IDN? On Connect", Groups: new[] { "VISA", "Debug" }, Order: 4.3, Collapsed: true, Description: "Send *IDN? when opening the connection to the instrument.")]
         public bool SendIDNOnConnect { get; set; }
 
+        /// <summary>
+        /// When true, will send *CLS right after establishing a connection.
+        /// </summary>
+        [Display("Send *CLS On Connect", Groups: new[] { "VISA", "Debug" }, Order: 4.4, Collapsed: true, Description: "Send *CLS when opening the connection to the instrument.")]
+        public bool SendCLSOnConnect { get; set; }
+
         #endregion
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


rmadsen-ks (approved):
Looks good to me!

Engine/ScpiInstrument.cs
```diff
@@ -369,11 +376,14 @@ public override void Open()
...
-                        Log.Info("Now connected to: " + IdnString);
+                        if (SendIDNOnConnect)
+                        {
+                            IdnString = QueryIdn();
+                            IdnString = IdnString.Trim();
+                            Log.Info("Now connected to: " + IdnString);
 
-                        CommandCls(); // Empty error log
+                            CommandCls(); // Empty error log
+                        }
```
rmadsen-ks, line 386: Actually I think CommandCls should be outside the if?
```suggestion

                        }
                        CommandCls(); // Empty error log
```

rmadsen-ks (changes requested):

Engine/ScpiInstrument.cs
c-montes: Good suggestion, I added a new boolean for SendCLSOnConnect, more flexibility for the user

rmadsen-ks (approved):
Looks good!

PR #1501, opened 2024-04-10
Use dotnet sdk in favor of old tap sdk new templates

### LINKED ISSUE #1319: Align tap sdk new and dotnet new opentap
If I create a new project using `tap sdk new project <name>`, it will correctly create a new project.
But the template for the included TestStep will be named the same as the project. So if I call the project "SomePlugin", the TestStep inside will look like this:

```
namespace SomePlugin
{
    [Display("SomePlugin")]
    public class SomePlugin : TestStep
    {
        // The rest of the template ignored
    }
}
```

If I, on the other hand, create the template with `dotnet new opentap`, it will create a TestStep that looks like this:

```
namespace SomePlugin
{
    [Display("MyTestStep", Description: "Insert a description here", Group: "SomePlugin")]
    public class MyTestStep : TestStep
    {
        // The rest of the template ignored
    }
}
```

I think the one created with `dotnet new` is more correct, as it is a little misleading for new users of OpenTAP, that the class created in `SomePlugin.cs` is not related to the overall plugin, but actually a TestStep that is just named incorrectly. 

In general, I think that the templates generated should be the identical, and not generate two different results. Another example of this is the `DisplayAttribute` added, it has a lot more details on the `dotnet new` template.

Was created using OpenTAP/SDK version 9.22.2+1f88c8b1.

/Pelle

alnlarsen:
Use the new `dotnet new` template in favor of the old `tap sdk new project` logic.

The dotnet template supports a bunch of configuration flags that I have not added, but should be pretty simple to add as CLI options if we want:

This is part of an ongoing effort to maintain fewer SDKs. We are currently maintaining 3 SDKs. We would like to maintain only one.

Closes #1319

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,7 +322,9 @@ jobs:
 
   TestWindowsPlan:
     runs-on: windows-2022
-    needs: Build-Win
+    needs: 
+      - Build-Win
+      - Package-Templates
     strategy:
       matrix:
         retries: [ 0, 1 ]
@@ -339,6 +341,13 @@ jobs:
         with:
           name: win-x64-bin
           path: bin/Release/
+      - name: Download Template package
+        uses: actions/download-artifact@v3
+        with:
+          name: package-templates
+          path: sdk-template
+      - name: Install Templates
+        run: dotnet new --install ./sdk-template/*.nupkg
       - name: Test
         working-directory: bin/Release
         run: ./tap.exe run ../../tests/regression.TapPlan --verbose --color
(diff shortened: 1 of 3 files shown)


rmadsen-ks (changes requested):
```
Directory.Build.props	Properties		TestProj.csproj		package.xml
MyTestStep.cs		TestProj		TestProj.sln
```

I think it should put the sln file in a separate folder. e.g TestProj.sln and TestProj/[the rest]

Why is there a folder called TestProj?

alnlarsen:
> ```
> Directory.Build.props	Properties		TestProj.csproj		package.xml
> MyTestStep.cs		TestProj		TestProj.sln
> ```
> 
> I think it should put the sln file in a separate folder. e.g TestProj.sln and TestProj/[the rest]
> 
> Why is there a folder called TestProj?

What command did you run to get this? I just tested this:

`> tap sdk new project TestProj`

image

This output looks similar to what you requested. I agree the output you listed is not good

rmadsen-ks (approved):
Works well!

PR #1510, opened 2024-04-18
Clarify break condition behavior

### LINKED ISSUE #1452: Update `Break Conditions` documentation
https://doc.opentap.io/User%20Guide/Introduction/#verdicts

```
When a break condition is encountered, execution of all child steps is stopped, and the test plan resumes from the point where the break was encountered.
```

It should be clear that break conditions affect the parent step.

alnlarsen:
Closes #1452


witoldpiet (approved):
Looks good to me.

PR #1522, opened 2024-04-24
1521 improve resolver errors

### LINKED ISSUE #1521: Image resolution from http repositories needs better error messages
The image resolver gives the same cryptic error message if:

1. An incorrect or expired user token is used
2. The specified http repository does not exist

In the first case, we should check if the user token is invalid by calling e.g. `ListTokens`.
In the second case, we should show an error message indicating the repository is not responding

alnlarsen:
Fix bug causing tokens not to be extracted in image installs.
Also adds nice error messages when resolution goes wrong due error response from repository.
Closes #1521

### DIFF
--- a/Package.UnitTests/Image/Deploy.cs
+++ b/Package.UnitTests/Image/Deploy.cs
@@ -289,43 +289,6 @@ namespace OpenTap.Image.Tests
             }
         }
 
-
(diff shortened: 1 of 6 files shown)


rmadsen-ks (approved):
Looks good

PR #1526, opened 2024-04-25
1524 reject results reserved by steprun

### LINKED ISSUE #1524: It is possible to override result parameters that OpenTAP depends on
Publishing a result in a test step called `Verdict` causes OpenTAP to throw an exception later if the verdict of a test step run is queried. This happens because the getter has this logic:

```csharp
case string r:
  return (Verdict) StringConvertProvider.FromString(r, TypeData.FromType(typeof(Verdict)), null);    
```

This obviously crashes if Verdict is a string, and it is not a Verdict.

Errors caused due to this are hard to track down because

1. They are only triggered if a result listener happens to check the step verdict
2. They are triggered in logic completely unrelated to where the result was published

I think `Results.Publish(...)` should fail immediately if a result column that OpenTAP depends on would be overwritten.

alnlarsen:
Error out immediately if a built-in stepRun result is overridden.

The error message looks like this:

`Error running "VerdictResultPublisher": Result name 'Verdict' is reserved by OpenTAP.`
Closes #1524

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -947,7 +947,7 @@ jobs:
         run: nuget push OpenTAP.*.nupkg ${{ secrets.NUGET_KEY }} -NonInteractive -Source https://api.nuget.org/v3/index.json
 
   Publish-TapPackages:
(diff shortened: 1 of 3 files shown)


alnlarsen:
This is not the right approach

PR #1536, opened 2024-04-26
1531 dont allow assigning strings to numbers

### LINKED ISSUE #1531: Dont allow assigning strings to numbers
This can go well sometimes, but since this is a user-facing part of OpenTAP applications it is probably better to be a bit more restrictive in which types can be assigned to which.

If it is wanted to assign a string to a number, expressions can be used.

alnlarsen:
Closes #1531


rmadsen-ks (approved):
Looks great!

PR #1537, opened 2024-04-26
Include group names in output selection dialog

### LINKED ISSUE #1517: Repeat step has 2 Iteration outputs
**Environment**
KS8400 v9.23.0-release
OpenTAP v9.24.0-rc.1

**Describe the bug**
Repeat step has 2 Iteration outputs

**To Reproduce**
1. Upgrade OpenTAP to 9.24.0-rc.1
2. Launch KS8400
3. Add a Repeat step
4. Add a Delay child for the Repeat step
5. Add a Log Output as a child of Repeat step
6. Select the Log Output step, right-click Log Message and click Assign Output
7. Check the Output dialog

**Actual behavior**
There are 2 Outputs with the same name (e.g. Iteration from Repeat)

**Expected behavior**
There should not be 2 different settings that have the same name

**Screenshots**
image

alnlarsen:
This adds the group names to outputs to avoid confusion when duplicate names are used:

image

I verified that this does not cause issues for serialization since we serialize the member name of the C# property, and not the display name.

Closes #1517

### DIFF
--- a/Engine.UnitTests/InputOutputRelationTest.cs
+++ b/Engine.UnitTests/InputOutputRelationTest.cs
@@ -467,6 +467,31 @@ namespace OpenTap.UnitTests
         }
 
         [Test]
+        [TestCase("Iteration", "3 of 3")]
+        [TestCase("Outputs \\ Iteration", "3")]
+        public void TestInputOutputNameClash(string output, string expected)
+        {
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks great!

PR #1539, opened 2024-04-26
Flip arguments to CanConvert

alnlarsen:
Arguments to CanConvert are now in the correct order. This did not cause issues before because CanConvert was reflexive for primitive types (but not for complex types).


rmadsen-ks (approved):
Looks great!

PR #1546, opened 2024-04-29
1544 fix result listener race condition

### LINKED ISSUE #1544: Race condition in OnTestStepRunStart
If a Result Listener adds steprun parameters in the OnTestStepRunStart event, it can mess with the internal lookups of other concurrent result listeners. This can be triggered with any test plan using the latest beta of the postgres plugin.

alnlarsen:
This adds a couple of steprun related unit tests and fixes the issue by cloning the stepRun object for each result listener.

Closes #1544


rmadsen-ks (approved):
Looks good!

PR #1566, opened 2024-05-13
Improve resolver errors again

### LINKED ISSUE #1525: improve resolver errors in trivial error cases
## Multiple packages with the same name but different versions specified.

This can happen in build scenarios where for example a OpenTapPackageReference is made in the Directory.Build.props file and the csproj file. The error message can be confusing:
```
 error OpenTAP Install: Unable to resolve image 'Sign:1.4.0, Sign:1.5.0-alpha.4.9+6a5355a8.sign-apple-apps, 
Dotfuscator:2.0.2-beta.6+11a5bc9a, OpenTAP:9.23.2+5a906a86': 
Sign: 1.4.0, Sign: 1.5.0-alpha.4.9+6a5355a8.sign-apple-apps was not found
```

Actually it was found, but the constraint cannot be solved. because both Sign package references are required to be that version.

## The package names does not exist

```
PS C:\git\opentap> ~/tap/tap image install --dry-run abc:1.2.3
Unable to resolve image 'abc:1.2.3': abc: 1.2.3 was not found
```

In this case, the problem is not the version, but no package exists with that name. It would be better to point that out.

### LINKED ISSUE #1563: OpenTAP Image Resolve Message
The image resolver attempts to resolve these references from a .csproj file:
```
 <AdditionalOpenTapPackage Include="Developer's System" Version="9.23.0" />
 <PackageReference Include="OpenTAP" Version="9.23.0" />
```

Should the resolver be able to bail out early from this as "Developer's System" Version="9.23.0" depends on OpenTAP Version="^9.23.2"?

Could we improve the error message when this happens? It can be difficult to figure out the root cause from this message:
 
image

alnlarsen:
This is part of an ongoing effort to make our resolver errors make sense.

The checks I have added are fairly ad-hoc, but they should mostly be correct in identifying issues.

Even if they are not completely correct, the errors we have today are mostly useless

Closes #1563 
Closes #1525

### DIFF
--- a/Package/Image/ImageResolution.cs
+++ b/Package/Image/ImageResolution.cs
@@ -14,6 +14,7 @@ namespace OpenTap.Package
 
         public virtual bool Success => true;
 
-        public override string ToString() => string.Format("[ImageResolution: {0}", string.Join(", ", Packages.Select(x => $"{x.Name} {x.Version}")));
+        public override string ToString() =>
+            $"[ImageResolution: {string.Join(", ", Packages.Select(x => $"{x.Name} {x.Version}"))}";
     }
 }
--- a/Package/Image/ImageResolveException.cs
+++ b/Package/Image/ImageResolveException.cs
@@ -1,0 +1,96 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Text;
+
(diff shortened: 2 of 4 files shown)


Oeh1995 (approved):
Looks very nice!

PR #1573, opened 2024-05-14
Remove redundant if statement

### LINKED ISSUE #1572: Cannot install images from network drive
Installing packages from a network drive (on windows) does not work. I was able to reproduce the issue by trying to install a package from my wsl package cache:

`./tap image install 'Python' -r '\\wsl.localhost\Arch\home\alex\.local\share\OpenTap\PackageCache'`

alnlarsen:
I don't know why this if statement exists, but I don't see any reason for it.

Closes #1572


rmadsen-ks (approved):
Looks good lets merge it.

PR #1585, opened 2024-05-23
Make unreachable repositories a warning instead of an error

### LINKED ISSUE #1579: Cannot modify installation while offline
We were a bit too aggressive in #1522 

Trying to install locally available packages while offline now fails.

alnlarsen:
This fix makes it once again possible to install images while offline (as long as the packages are cached)
Closes #1579


rmadsen-ks (approved):
Looks great!

PR #1587, opened 2024-05-29
Update Readme.md

rmadsen-ks:
Removed comment about visa addresses in result listener documentation. This is already handled by escaping strings containing the delimiters.


alnlarsen (approved):
Great!

PR #1594, opened 2024-06-04
Only convert sweep result parameters to strings if they are not IConvertible

### LINKED ISSUE #1593: Result parameter type discrepancy in sweep loop
When sweeping parameters, the parameter is added as an additional result to all child steps. 

Sweep loop converts the swept value to a string, even if the value is already IConvertible. The swept value is also added as a result to the sweep loop itself, but here the actual type is used, and not the IConvertible type. This causes issues for some result listeners that do not expect result values to change type.

alnlarsen:
This fixes an issue for some result listeners that expect results with the same name not to switch types.

Closes #1593

### DIFF
--- a/BasicSteps/SweepParameterStep.cs
+++ b/BasicSteps/SweepParameterStep.cs
@@ -251,21 +251,25 @@ namespace OpenTap.Plugins.BasicSteps
                 var AdditionalParams = new ResultParameters();
 
 
-                foreach (var set in sets)
+                string getValueString(object value)
                 {
-                    var mem = rowType.GetMember(set.Name);
-                    var value = mem.GetValue(Value);
-
                     string valueString;
                     if (value == null)
                         valueString = "";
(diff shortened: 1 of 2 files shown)


rmadsen-ks:
We agreed not adding this improvement.

Denrask:
Hi @rmadsen-ks and @alnlarsen 

Why did you agree not to fix this issue?

rmadsen-ks:
@Denrask, this PR does not fix the root cause of the issue and the implementation was not "wrong", so we decided against changing the behavior.

The OpenTAP API does not guarantee a lot about the types of the parameters. If one test step has a parameter with the same name as another test step, there is no way we could guarantee that those would be the same type.

So we could change this in the sweep loop step it will occur again because of some other test step. 

If you are in doubt about how to handle parameter types in a given result listener, I suggest just converting it to a string.

PR #1598, opened 2024-06-06
Added a function to ShortTimeSpan that allows for formatting of timepans longer than 1 second.

Oeh1995:


rmadsen-ks (changes requested):

Engine.UnitTests/TestPlanCompositeRunTests.cs
```diff
@@ -114,14 +114,51 @@ public void ShortTimeSpanTest()
...
+                new TimeSpan(0, 0, 84600) , 
+                new TimeSpan(0, 0, 2031720), 
+                new TimeSpan(0,0, 86700), 
+                new TimeSpan(0,0, 18003),
+                new TimeSpan(0,0, 260470),
+                new TimeSpan(0,0, 86399999),
+                new TimeSpan(0,0, 86459),
+                new TimeSpan(0,0, 1),
+                new TimeSpan(0,0, 0)
+            };
```
rmadsen-ks, line 140: Instead of these seconds, can we use the more specific way of declaring the time?
```
                new TimeSpan(1,0, 0, 0), 
                new TimeSpan(0, 23, 30, 0) , 
                new TimeSpan(23, 12, 22, 0), 
                new TimeSpan(1,0, 5,0),  ...
```

Shared/ShortTimeSpan.cs
```diff
@@ -58,6 +62,32 @@ public static ShortTimeSpan FromSeconds(double seconds)
...
 
+        /// <summary>
+        /// For formating of timespans longer than 1 minute like "1 d 1 h 14m" instead of "90840 s"
+        /// </summary>
+        /// <param name="timeSpan"></param>
+        /// <returns></returns>
+        public static string LongTimeSpanFormat(TimeSpan timeSpan)
+        {
+            string result = "";
+            if (timeSpan.TotalSeconds == 0)
```
rmadsen-ks, line 73: This should probably be timeSpan.TotalSeconds < 1. Since TotalSeconds is a double.

rmadsen-ks (changes requested):
I have some minor suggestions.

rmadsen-ks (approved):
Beautiful!

rmadsen-ks:
After testing it a bit more, I noticed that the new time span is only used  for the total test plan time. I think it should also be used for the other items in the summary.

image

rmadsen-ks (changes requested):
See my latest comment

Engine/TestPlanRunSummaryListener.cs
```diff
@@ -121,12 +121,16 @@ static void printSummary(IDictionary<Guid, TestStepRunData> stepRuns, TestStepRu
             int indentcnt = stepRunIndent(stepRuns, run);
             string indent = new String(' ', indentcnt * 2);
             string inverseIndent = new String(' ', maxIndent * 2 - indentcnt * 2);
-            
+            if(run.Duration.TotalSeconds < 1)
+            {
+
+            }
```
rmadsen-ks, line 127: Did you mean to do something here?

Shared/ShortTimeSpan.cs
```diff
@@ -12,7 +12,11 @@ namespace OpenTap
     /// </summary>
     struct ShortTimeSpan
     {
+       
         public enum UnitKind{
+            Days,
+            Hours,
+            Minutes,
```
rmadsen-ks, line 19: It seems like we dont actually use these extra cases, right? Maybe we should leave them out.

rmadsen-ks (dismissed):
Looks good, but I think some things could be removed.

alnlarsen:
image

It looks like we are missing a space before the verdict.

alnlarsen (changes requested):
Please see comments

alnlarsen:
This looks really good now. There is a minor alignment issue when steps take more than an hour, but let's open a separate issue for calculating the required alignment for the timespan value.

alnlarsen (approved):

PR #1600, opened 2024-06-06
Upgrade System.Text.Encodings.Web to use the most secure version.

rmadsen-ks:
System.Text.Json depends on System.Text.Encodings.Web internally


alnlarsen (dismissed):
Looks good!

PR #1606, opened 2024-06-10
Determine if a relative url is an http repository by trying to connect to it

### LINKED ISSUE #1595: tap package install returns 405 Not Allowed
**Describe the bug**

Different "tap package" commands returns:

```html
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx/1.24.0</center>
</body>
</html>
```

 - OS: Linux
 - OpenTAP Version 9.24.2

**To Reproduce**

1. On Linux install OpenTAP 9.24.2

2. Run `./tap package install "N6705 DC Power Analyzers(v6)"`

**Actual behavior**

Installing a dependency to the `N6705 DC Power Analyzers(v6)` package and then try to install the package = 405
image

Downgrade OpenTAP to 9.20 and try to install the package = success!
image

This might also be a weird error message case:

1. tap package install package -> Could not resolve
2. install dependency of package
3. tap package install package -> success

image

alnlarsen:
This fixes an issue that makes it impossible to use a local absolute path on Linux and Mac.

Closes #1595


Denrask (approved):
Looks good. Don't we have unittests exercising this?

What happens now if the url is an absolute path is a directory that does not exist? I thought `DetermineRepositoryType` was handling this case before?

alnlarsen:
@Denrask if an absolute path is used which is determined to be an http repository it will be used as a file repository (which will of course have no packages).

You raise a good point. We have a lot of unit tests exercising this, but they don't use base address. We should test them all with and without a base address.

PR #1609, opened 2024-06-11
1568 allow parallel step children to be repeated by repeat mixin

### LINKED ISSUE #1568: Parallel step disables "suggested next step" functionality
**Describe the bug**

When setting the StepRun.SuggestedNextStep property to the Id of the current step, the test step will be repeated. This does not work inside a parallel step.

 - OS: Windows 10
 - OpenTAP Version 9.24.1

**To Reproduce**

1. Create a test step that sets StepRun.SuggestedNextStep to the Id of the current test step

2. Run inside and  outside of a parallel step

Alternative: 
1. Use any step (e.g. Delay), use Repeat Mixin from Basic Mixin Package 
2. Run inside and  outside of a parallel step

**Actual behavior**

Inside of a parallel step, the step is only executed once

**Expected behavior**

Inside of a parallel step, the step is repeated exactly as outside of a parallel step

Oeh1995:
This should allow child steps inside a parallel step to be repeated by repeat mixin.

Closes #1568


Engine/TestStep.cs
```diff
@@ -859,7 +859,10 @@ public static TestStepRun RunChildStep(this ITestStep step, ITestStep childStep,
                 if(run.Verdict == Verdict.Error && throwOnBreak)
                     run.ThrowDueToBreakConditions();
             }
-
+            if (run.SuggestedNextStep == childStep.Id)
+            {
+                RunChildStep(step, childStep, throwOnBreak, currentPlanRun, currentStepRun, attachedParameters);
```
alnlarsen, line 864: I think we need to return the run from the recursive call here. Imagine for example a Repeat mixin which repeats a step until the verdict is `Pass`. Then we should return the TestStepRun from the attempt that succeeded.

alnlarsen (changes requested):
Looks good, but I see a minor issue

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
+            plan.ChildTestSteps.Add(parallelStep);
+            plan.PrintTestPlanRunSummary = true;
+            PlanRunCollectorListener planRunListener = new PlanRunCollectorListener();
+        
+            var planRun = plan.Execute(ResultSettings.Current.Concat(new IResultListener[] { planRunListener }));
+            
+            Log.Flush();
+            Log.RemoveListener(trace);
+            var summaryLines = trace.allLog.ToString().Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries).Where(l => l.Contains(": Summary ")).ToArray();
+            Assert.AreEqual(14, summaryLines.Count(), "Did not find the expected number of summary lines in the log.");
```
alnlarsen, line 662: This does not seem very reliable. If we add a new line to the summary logger, this test is going to fail.

The thing we actually want to test is that the child step executed 10 times. We can verify this by checking `planRunListener.StepRuns`

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
             Assert.AreEqual(Verdict.NotSet, run.Verdict);
         }
+        /// <summary>
+        /// Tests that a parallel step where the SuggestedNextStep is set to the current ID will repeat the step. 
+        /// </summary>
+        [Test]
+        public void ParallelSuggestedNextStepTest()
+        {
+            TestTraceListener trace = new TestTraceListener();
+            Log.AddListener(trace);
```
alnlarsen, line 647: We don't need a TestTraceListener here

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
+            var plan = new TestPlan();
+            var parallelStep = new ParallelStep();
+            var delayStep = new ParallelStepSuggestedNextStep();
+            delayStep.RepeatCount = 10;
+            parallelStep.ChildTestSteps.Add(delayStep);
+            plan.ChildTestSteps.Add(parallelStep);
+            plan.PrintTestPlanRunSummary = true;
```
alnlarsen, line 654: There is no need to set this

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
+            var planRun = plan.Execute(ResultSettings.Current.Concat(new IResultListener[] { planRunListener }));
+            
+            Log.Flush();
+            Log.RemoveListener(trace);
+            var summaryLines = trace.allLog.ToString().Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries).Where(l => l.Contains(": Summary ")).ToArray();
+            Assert.AreEqual(14, summaryLines.Count(), "Did not find the expected number of summary lines in the log.");
+
+        }
 
+        public class ParallelStepSuggestedNextStep : TestStep
```
alnlarsen, line 666: Can we find a nicer name for this step? The step itself has nothing to do with ParallelStep

tap/Properties/launchSettings.json
```diff
@@ -29,6 +22,10 @@
     "tap (profile)": {
       "commandName": "Project",
       "commandLineArgs": "profile --test OpenTap.UnitTests.ScopeParametersTest.TestIsParameterized"
+    },
+    "Tap Editor": {
+      "commandName": "Executable",
+      "executablePath": "C:\\Users\\olivhans\\source\\repos\\opentap\\opentap\\bin\\Debug\\Editor.exe"
```
alnlarsen, line 28: Let's remove this

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
+            var plan = new TestPlan();
+            var parallelStep = new ParallelStep();
+            var delayStep = new ParallelStepSuggestedNextStep();
+            delayStep.RepeatCount = 10;
+            parallelStep.ChildTestSteps.Add(delayStep);
+            plan.ChildTestSteps.Add(parallelStep);
+            plan.PrintTestPlanRunSummary = true;
+            PlanRunCollectorListener planRunListener = new PlanRunCollectorListener();
+        
+            var planRun = plan.Execute(ResultSettings.Current.Concat(new IResultListener[] { planRunListener }));
```
alnlarsen, line 657: There is no need to append the current result settings. Overriding is fine in a unit test, unless there is a specific reason

Engine.UnitTests/TestPlanTest.cs
```diff
@@ -638,8 +637,46 @@ public void RepeatChildStepsFailure()
...
+        /// </summary>
+        [Test]
+        public void ParallelSuggestedNextStepTest()
+        {
+            TestTraceListener trace = new TestTraceListener();
+            Log.AddListener(trace);
+            var plan = new TestPlan();
+            var parallelStep = new ParallelStep();
+            var delayStep = new ParallelStepSuggestedNextStep();
+            delayStep.RepeatCount = 10;
```
alnlarsen, line 651: This is not a delay step. Let's rename it

alnlarsen (changes requested):
I think we can improve the test a bit

alnlarsen (approved):
Looks great!

PR #1612, opened 2024-06-18
Update newtonsoft

### LINKED ISSUE #1599: Update to Newtonsoft.Json 13.0.3
Newtonsoft.json 12.x had a known security vulnearability, namely possiblity of DDOS attack due to stack exhaustion (Many recursively nested objects).

This has been fixed in 13, where it has been limited to 64 levels, but at the same time that is a breaking change for people needing more that 64 levels.

alnlarsen:
The only breaking change in this Newtonsoft major bump is a max limit on recursion depth during serialization. It should be safe to upgrade.

Closes #1599


Oeh1995 (approved):
Looks good

PR #1614, opened 2024-06-19
Delete unfinished documentation

alnlarsen:
This deletes some documentation we never finished. It is currently live, but nothing links to this page. It is only accessible through the search function. Let's just delete it.


Oeh1995 (approved):
Looks fine!

PR #1617, opened 2024-06-19
1589 make tap image browsable

### LINKED ISSUE #1589: Make the 'tap image' subcommand browsable
'image install' has always been unbrowsable. I think we should just open it up. Internally, we use it a lot for creating installations, even from the CLI, so I think we should just open it up.

alnlarsen:
This will make `tap image install` appear in the output of e.g. `tap --help`, and most
importantly, reveals the possible arguments in `tap image install
--help` which is currently a bit too secret.

Closes #1589


sebastian-pop:
Implemented

image

PR #1619, opened 2024-06-19
1616 allow documenting unnamed arguments

### LINKED ISSUE #1616: CLI: Allow documenting [UnnamedCommandLineArgument]
It is currently possible to provide a description for `[CommandLineArgument]`, which is used in the auto-generated help output for e.g. `tap package list --help`, but it is not possible to provide a description.

We should add a Description property to `[UnnamedCommandLineArgument]`.

alnlarsen:
This adds help output for unnamed arguments. They are currently a bit hard to spot at the end of the current `Usage` text. See example:

```
Usage: install [-h] [-v] [-c] [-q] [--log <arg>] [--merge] [--non-interactive] [--os <arg>] [--architecture <arg>] [--dry-run] [-r <arg>] [-f] [-t <arg>] <image>
  <image>             The image to install. Supported formats:
                      A string describing the image. Example: 'OpenTAP:9.24,TUI:beta,CSV'
                      A path to a file containing an XML or JSON formatted image specification.
  -h, --help          Write help information.
  -v, --verbose       Show verbose/debug-level log messages.
  -c, --color         Color messages according to their severity.
  -q, --quiet         Quiet console logging.
  --log               Specify log file location. Default is ./SessionLogs.
  --merge             Merge the requested image with the existing installation instead of overwriting it.
  --non-interactive   Never prompt for user input.
  --os                The operating system to resolve packages for.
  --architecture      The architecture to resolve packages for.
  --dry-run           Only print the result, don't install the packages.
  -r, --repository    Repositories to use for resolving the image.
  -f, --force         Try to run in spite of errors.
  -t, --target        Override the location where the command is applied.
                      The default is the OpenTAP installation directory.

```

When I made `image install` public, I realized that the `<image>` argument is impossible to figure out without help, and the only alternative way to document it is with an error message which is not very friendly.

Closes #1616


StefanHolst (approved):
This is beautiful! I have been missing this a bunch of times

PR #1621, opened 2024-06-19
Add abort points while waiting for subprocess to exit

### LINKED ISSUE #1604: Run Program: Test Plan cannot be stopped while Firefox is open
**Environment**
KS8400 9.24.0-rc.2

**Describe the bug**
Test Plan cannot be stopped while Firefox is open

**To Reproduce**

1. Add a Run Program step
2. Set `firefox.exe` as application
3. Run the test plan
4. Wait for Firefox to be opened
5. Switch to KS8400
6. Press Stop
7. Actual behavior
8. The test plan cannot be stopped

**Additional notes**
This issue also occurs on older version of KS8400

**Screenshots**
run_program_testplan_cannot_be_stopped

alnlarsen:
This has some ugly workarounds because dotnet contains ugly bugs. The process API really is crazy.

In general, it's impossible to control a child process unless it cooperates. In the case of Firefox, it is spawning a child process that spawns a new window. So the initial process has actually exited, but because it forwarded its output / error pipes to the child process, they will never close, which causes issues in this case.

I think the best we can do in such scenarios it to allow aborting the test plan, and accept that the child process hangs instead of closing.

Closes #1604

### DIFF
--- a/BasicSteps/ProcessStep.cs
+++ b/BasicSteps/ProcessStep.cs
@@ -110,7 +110,6 @@ namespace OpenTap.Plugins.BasicSteps
         [EnabledIf(nameof(WaitForEnd), true, HideIfDisabled = true)]
         public int ExitCode { get; private set; }
         
-        ManualResetEvent outputWaitHandle, errorWaitHandle;
         StringBuilder output;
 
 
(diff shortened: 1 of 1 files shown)


Oeh1995 (approved):
Looks good!

PR #1623, opened 2024-06-21
Set assembly version to gitversion during build

### LINKED ISSUE #1556: Get rid of Mono.Cecil dependency
We are currently using Mono.Cecil to set assembly versions using GitVersion. We can do this at build time by using the `GenerateAssemblyInfo` option.

alnlarsen:
This adds a new target to the nuget package to set gitversion during build instead of during package creation.

This is nice for a couple of reasons:

* we can remove the dependency on mono.cecil
* shorter package.xml files since version injection no longer needs to be specified for all dlls
* The hash of a packaged DLL will match the hash of the input dll, unless signed / obfuscated of course
* dlls will be versioned correctly in debug builds
* we no longer need to update pdb files when we update the version info, since this is now done correctly by the compiler

I had to make GitVersionCalculator public. Let's consider if we want to actually do this. It would not be a big deal if it was in the 'SDK' package, but unfortunately it is in OpenTap.Package.

### DIFF
--- a/Package/CreatePackage/GitVersionCalculator.cs
+++ b/Package/CreatePackage/GitVersionCalculator.cs
@@ -18,7 +18,7 @@ namespace OpenTap.Package
     /// <summary>
     /// Calculates the version number of a commit in a git repository
     /// </summary>
-    internal class GitVersionCalulator : IDisposable
+    public class GitVersionCalulator : IDisposable
     {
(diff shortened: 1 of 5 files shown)


alnlarsen:
This feature was merged in #1772 and (issue #1687)

PR #1628, opened 2024-06-26
AssemblyFinder will skip a subdirectory if it appears to be a OpenTap installation.

### LINKED ISSUE #1571: Refuse to create new installations inside existing installations
This is not a new issue. We have known that this doesn't work for several years since we have a quite old check in the `sdk new project` template for this very reason.

Someone had a problem recently because they are building into a subdirectory of `$TAP_PATH`. It looks like the Serializer and the Assembly Resolver will not always agree on which instance of types it should use. This of course causes issues when we try to create an instance of a type from an assembly which is not loaded.

I tried reproducing this on 9.24, 9.23, and 9.22, but I suspect it goes back further. 

** Steps to reproduce **
1. `tap image install 'OpenTAP:9.24' -t .`
2. `tap image install 'OpenTAP:9.21' -t subdir`
3. `./tap package install 'TUI'`
4. `./tap package list -i`

Oeh1995:
Closes #1571

### DIFF
--- a/Engine/AssemblyFinder.cs
+++ b/Engine/AssemblyFinder.cs
@@ -97,7 +97,7 @@ namespace OpenTap
                             {
                                 if (StrEq(subDir.Name, "obj"))
                                     continue; // skip obj subfolder
-                                if (subDir.GetFiles("Tap.exe").Any())
+                                if (subDir.GetFiles("OpenTap.dll").Any())
                                     continue; // skip subdirectory if it appears to be an opentap installation
                                 var ignorePluginsInSubDir = dir.IgnorePlugins || StrEq(subDir.Name, "Dependencies");
                                 if (IncludeDependencies)
--- a/Engine/AssemblyFinder.cs
+++ b/Engine/AssemblyFinder.cs
@@ -96,12 +96,7 @@ namespace OpenTap
                             foreach (var subDir in dir.Info.EnumerateDirectories())
                             {
                                 if (StrEq(subDir.Name, "obj"))
-                                    continue; // skip obj subfolder
-                                if (subDir.GetFiles("OpenTap.dll").Any())
-                                {
-                                    log.Warning("Found OpenTap installation in subdirectory: '{0}'", subDir.Name);
-                                    continue; // skip subdirectory if it appears to be an opentap installation                                    
-                                }
+                                    continue; // skip obj subfolder                        
                                 var ignorePluginsInSubDir = dir.IgnorePlugins || StrEq(subDir.Name, "Dependencies");
                                 if (IncludeDependencies)
                                     ignorePluginsInSubDir = false;
--- a/Engine/AssemblyFinder.cs
+++ b/Engine/AssemblyFinder.cs
@@ -98,7 +98,10 @@ namespace OpenTap
                                 if (StrEq(subDir.Name, "obj"))
                                     continue; // skip obj subfolder
                                 if (subDir.GetFiles("OpenTap.dll").Any())
-                                    continue; // skip subdirectory if it appears to be an opentap installation
+                                {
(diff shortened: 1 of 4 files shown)


Engine/AssemblyFinder.cs
```diff
@@ -97,6 +97,8 @@ private void SyncFiles()
                             {
                                 if (StrEq(subDir.Name, "obj"))
                                     continue; // skip obj subfolder
+                                if (subDir.GetFiles("Tap.exe").Any())
+                                    continue; // skip subdirectory if it appears to be an opentap installation
```
alnlarsen, line 101: `tap.exe` only exists on Windows. We should probably check `OpenTap.dll` instead.

Engine/AssemblyFinder.cs
```diff
@@ -97,6 +97,8 @@ private void SyncFiles()
                             {
                                 if (StrEq(subDir.Name, "obj"))
                                     continue; // skip obj subfolder
+                                if (subDir.GetFiles("OpenTap.dll").Any())
+                                    continue; // skip subdirectory if it appears to be an opentap installation
```
alnlarsen, line 101: We should probably log a warning when this happens the first time for a given directory.

Engine/AssemblyFinder.cs
```diff
@@ -97,6 +97,11 @@ private void SyncFiles()
                             {
                                 if (StrEq(subDir.Name, "obj"))
                                     continue; // skip obj subfolder
+                                if (subDir.GetFiles("OpenTap.dll").Any())
+                                {
+                                    log.Warning("Found OpenTap installation in subdirectory: '{0}'", subDir.Name);
```
alnlarsen, line 102: We need to inform the user that plugins will not be loaded from this subdirectory since this is new behavior. Nested installations can cause unexpected behavior and are not supported.

alnlarsen:
I have been thinking about this a bit more. I think it is a risky change to completely ignore subfolders containing 'opentap.dll'. This is risky because it's possible to add additional search paths using `PluginManager.DirectoriesToSearch.Add(...)`, so I think the best option is just to ensure the serializer does not pick a different version of e.g. OpenTap types than the version which is currently loaded.

I think we should still follow Rolfs original suggestion of throwing an exception when trying to deploy a new installation into a subdirectory of OpenTAP.

Oeh1995:
I think we should create a new issue dealing with the Serializer and then keep going forward with throwing and exception if you try to install opentap in a subdirectory of another installation.

Package/PackageActions/Install.cs
```diff
@@ -109,8 +109,14 @@ private int DoExecute(CancellationToken cancellationToken)
         {
             if (Target == null)
                 Target = FileSystemHelper.GetCurrentInstallationDirectory();
+            if (NestedTapInstallation() == true)
```
alnlarsen, line 112: No need to check if a bool is == true

Package/PackageActions/IsolatedPackageAction.cs
```diff
@@ -78,6 +78,30 @@ public override int Execute(CancellationToken cancellationToken)
...
+                    filePath = Path.Combine(parentDir.FullName, "OpenTap.dll");
+                    if (File.Exists(filePath))
+                    {
+                        fileExists = true;
+                        break;
+                    }
+                    parentDir = parentDir.Parent;
+                }
+            }
+            return fileExists;
```
alnlarsen, line 101: We can make this a lot simpler. Maybe something like:

```cs
var dir = new DirectoryInfo(target).Parent;
while (dir != null)
{
    if (dir.EnumerateFiles("OpenTap.dll").Any())
        return true;
    dir = dir.Parent;
}
return false;
```

alnlarsen (approved):
This works great now!

PR #1629, opened 2024-06-26
Delete unused project

### LINKED ISSUE #1630: Delete Engine.TestModule/
This folder was checked in in the initial commit, and as far as I can tell, it is not used for anything. Let's delete it.

alnlarsen:
This does not seem to be used at all, and has not been changed since it was checked in. Let's get rid of it

Closes #1630


Oeh1995 (approved):
Looks great!

PR #1632, opened 2024-06-26
flush InitTraceListener before removal

### LINKED ISSUE #1631: Messages logged during tap initialization are sometimes lost
This happens because we don't flush the root log context before we detach the InitTraceListener.

This is likely not super critical. Mostly this can result in some diagnostic messages from OpenTAP not arriving sometimes.

alnlarsen:
This fixes an issue causing startup messages to disappear.

Because the flush method on the init trace listener dit not actually flush, messages from e.g. PluginSearcher during startup did not always arrive before the listener was removed.

Closes #1631

### DIFF
--- a/Cli/TapInitializer.cs
+++ b/Cli/TapInitializer.cs
@@ -23,7 +23,7 @@ namespace OpenTap
                     AllEvents.AddRange(events);
             }
             public void Flush(){
-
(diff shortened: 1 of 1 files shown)


Oeh1995 (approved):
Looks good

PR #1647, opened 2024-07-04
fix duplicate test steps not being reported

### LINKED ISSUE #1503: Duplicate Type names should produce a warning during startup
Example:
- OpenTap.Plugins.A from assembly PluginAssembly1.dll
- OpenTap.Plugins.A from assembly PluginAssembly2.dll

alnlarsen:
This also adds some duplicate types to our unit test project 

Closes #1503


Shared/DuplicateTestStep.cs
Oeh1995, line 1: If this is only used in Tap.Engine.UnitTests, then I think it would be good to move it into that project instead of Shared.

Oeh1995 (changes requested):
See comment

Oeh1995 (approved):
Looks good!

PR #1648, opened 2024-07-04
Remove byte-order-mark from template files

alnlarsen:
This causes issues for some people. Let's get rid of it.


AsgerIversen (approved):
Nice

PR #1660, opened 2024-07-05
Fix system-wide installs failing with exit code 0

### LINKED ISSUE #1655: Install a plugin: Order of messages is sometimes not correct
**Describe the bug**

 - OS:  Windows 10 (seen on Ubuntu as well)
 - OpenTAP Version 9.25.0-rc.1 

**To Reproduce**

1. Install PathWave License Manager

2. Verify CLI output

**Actual behavior**

The message `Installing to C:\Program Files\OpenTAP` is displayed after the plugin has been installed

image

**Expected behavior**

Does not reproduce for Demonstration for example

### LINKED ISSUE #1658: Package install action has exit code 0 when system wide installation fails
This happens because we don't do anything with the error in cases when system-wide installation fails.

alnlarsen:
This causes failing system-wide installs to actually fail the installation. Also fixes an edge case related to logging

Closes #1658
Closes #1655

### DIFF
--- a/Package/Installer.cs
+++ b/Package/Installer.cs
@@ -73,8 +73,7 @@ namespace OpenTap.Package
                         Stopwatch timer = Stopwatch.StartNew();
                         PackageDef pkg = PluginInstaller.InstallPluginPackage(TapDir, fileName, UnpackOnly);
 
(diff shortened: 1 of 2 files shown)


Oeh1995 (approved):
Looks good

PR #1664, opened 2024-07-08
1656 No Output for isolated process: Fix

### LINKED ISSUE #1656: Linux: No output after using a command with `-q` argument
**Describe the bug**

 - OS: Ubuntu 20.04 (not reproduced on Windows)
 - OpenTAP Version 9.25.0-rc.1

**To Reproduce**

1. `./tap package list -i -q` - > no output because of `-q`
2. `./tap package install -h` 

**Actual behavior**

No output
image

**Expected behavior**

`--quiet` was not used so the second command should have output

rmadsen-ks:
After some investigation, it turned out that the error can occur with isolated processes started with the -h argument. This is probably because they exit very quickly after printing the output, but in reality probably any output can get lost.

Somehow this only affected Linux. I think this might be  because the linux/.net 8 process is just faster at exiting than the windows one.

To fix it, I added waiting for the output writing tasks to finish before ending the parent process.

Close #1656

### DIFF
--- a/Shared/ExecutorInterop.cs
+++ b/Shared/ExecutorInterop.cs
@@ -27,12 +27,21 @@ namespace OpenTap
         Process Process { get; set; }
 
         ProcessStartInfo start;
+        
+        // tasks for processing child log output.
+        Task redirectErrorTask;
+        Task redirectOutputTask;
 
         public int WaitForExit()
         {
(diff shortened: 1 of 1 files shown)


Oeh1995 (approved):
Looks good!

PR #1667, opened 2024-07-09
Allow resources to be both DUTs and Instruments

### LINKED ISSUE #1666: Not possible to select an instrument instance of a resource that implements both IDut and IInstrument
See this example:

```cs
public class MyInstrumentDut : Resource, IDut, IInstrument
{
    
}
public class IDutUser : TestStep
{
    public IDut MyDut { get; set; }
    public IInstrument MyInstrument { get; set; }
    public MyInstrumentDut Subclass { get; set; }
    public override void Run()
    {
    }
}
```

If MyInstrumentDut is configured as an IInstrument on the bench, it is not possible to assign it to the Subclass property

alnlarsen:
This fixes an issue preventing selecting a configured Instrument if the concrete implementation is also a DUT

Closes #1666

### DIFF
--- a/Engine.UnitTests/AnnotationTest.cs
+++ b/Engine.UnitTests/AnnotationTest.cs
@@ -2167,6 +2167,53 @@ namespace OpenTap.UnitTests
             }
         }
 
+        public class DutAndInstrumentResource : Resource, IDut, IInstrument
+        { 
+        }
+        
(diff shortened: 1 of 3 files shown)


rmadsen-ks (approved):
Looks good. Lets get it in!

PR #1669, opened 2024-07-10
1668 package dependency caching

### LINKED ISSUE #1668: Caching of package dependencies
Sometimes package lookups can be a bit slow. Especially when the result contains 1000s of packages. 

To improve this we can add some caching mechanism the the package resolver, that it saves the result of the package request to the disk and only adds new packages the next time.

This can be done by restricting the package lookup by "UploadDate". The first time we do the query, we take all packages, but then we can register the current date and only search for newer dates the next time.

There are a few additional situations where the cache needs to be invalidated though.
- If the users permissions changes they will be able to see packages with an older upload date than the key.
- If a package is hidden or removed it will no longer be resolvable. 

The first problem can be solved by getting the name of all visible packages in the repo. If this name changes the cache needs to be invalidated.

The second problem only relates to pre-release package. However this will usually not be a problem because if the cache exists it most often means that the image was resolved and installed, which means the package exists in a local cache. 
Another solution might be to see that even though the package was resolved, installing the image fails. In this case we need to invalidate the cache.

rmadsen-ks:
Currently this does not improve the performance a lot since the database does not filter optimally when searching, but at least it reduces the traffic and C# overhead. I think its around 10% faster

Close #1668

### DIFF
--- a/Package/Image/PackageDependencyGraph.cs
+++ b/Package/Image/PackageDependencyGraph.cs
@@ -189,6 +189,39 @@ namespace OpenTap.Package
             }
         }
 
+        public void WriteToJson(Utf8JsonWriter jsonWriter, string arrayName)
+        {
+            if(arrayName != null)   
+                jsonWriter.WriteStartArray(arrayName);
+            else
+                jsonWriter.WriteStartArray();
+            
+            foreach (var pck in PackageSpecifiers())
+            {
+                jsonWriter.WriteStartObject();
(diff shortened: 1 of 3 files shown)


alnlarsen (commented):
I like this a lot. I am of course worried about invalidating the cache. Let's discuss how / when we should invalidate it. 

Maybe we should merge this early on after bumping to 9.26 so we can start evaluating.

PR #1671, opened 2024-07-11
Improve incremental builds

### LINKED ISSUE #1670: NuGet: speed up incremental builds
Because the dependencies of tasks in `OpenTap.Targets` are not perfectly mapped out, we are doing a lot of unnecessary work in builds without a lot of changes. Especially on slow network connections, and when using references to beta packages, we make a lot of network requests to packages.opentap.io that are not required.

If an image install is done (which it always is), package references must also be re-generated, which also triggers some linking in dotnet (opentap.g.props).

We can improve this drastically by detecting if packages are up to date.

alnlarsen:
This drastically improves incremental builds by introducing a marker file for installed packages. By skipping the InstallOpenTapPackages step, we avoid having to:

1. Make network requests
2. Re-generating opentap.g.props references

This saves a lot of time in solutions using OpenTapPackageReference, especially if the solution has many projects, or if it uses WPF.

This is especially noticable on a slow internet connection

Closes #1670


rmadsen-ks (approved):
Looks good. @alnlarsen is already using it for builds.

PR #1686, opened 2024-08-05
1685 Type Filter Attribute: Fix

### LINKED ISSUE #1685: Type Filter Attribute
Add additional constraints on a settings type.

```c#
[TypeFilter(typeof(IDMM))]
public IPSU Psu{get;set;}
```

rmadsen-ks:
- Added TypeFilterAttribute and implemented support in the annotations.
- Added a unit test to verify that it works as intended.
Close #1685


alnlarsen (commented):
The implementation looks great, but is this actually something we need?

PR #1702, opened 2024-08-19
1699 Connection GetOtherPort with Equals

### LINKED ISSUE #1699: Connection.GetOtherPort should use Equals
Connection.GetOtherPort currently uses `==`, however ports are often more marker-type objects, without much own state, so it would be more natural to use object.Equals

rmadsen-ks:
Fixed the issue by using Equals.

Added a unit test to show the issue.

Close #1699


alnlarsen (approved):
Looks great!

PR #1703, opened 2024-08-19
1698 ThreadSafe Scpi Query/Command: Fix

### LINKED ISSUE #1698: SCPI Instrument should be thread safe for ScpiQuery/Command
We support locking the SCPI resource across devices, but not inside the process itself. 

Let's make queries and commands thread safe by adding a lock.

To test this behavior, try this:

image

rmadsen-ks:
Fixed the issue by introducing locked commands and queries using an internal lock on the scpi instrument.

Before:
image

After:

image

Close #1698

### DIFF
--- a/Engine/ScpiInstrument.cs
+++ b/Engine/ScpiInstrument.cs
@@ -49,14 +49,19 @@ namespace OpenTap
         
         /// <summary> Some instruments does not support reading the status byte. This is detected during Open. </summary>
         bool readStbSupported = true;
+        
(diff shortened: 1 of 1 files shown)


alnlarsen (commented):
Looks good. We already have some locking mechanisms. (FineGrainedLock, DoLock, etc.)

This new lock looks a lot more coarse. Can you explain the difference between these different locking mechanisms? Do we need them all?

alnlarsen (approved):
Looks great

PR #1705, opened 2024-08-20
1701 RF Connection Get Intepolated Cable loss default: Fix

### LINKED ISSUE #1701: RfConnection.GetInterpolatedCableLoss throws an exception if there is no calibration data
RfConnection.GetInterpolatedCableLoss throws an exception if there is no calibration data in the loss files. 

This seems a bit extreme since you'd normally just assume a loss of 0 if you didnt know better.

I suggest we modify this to just return 0 when there is no calibration data defined.

rmadsen-ks:
- Added that the RF connection should return 0.0 dB loss if there are no loss points configured.

- Added unit tests to verify the behavior.
- Modified unit tests to test more interpolation cases.

Close #1701

### DIFF
--- a/Engine.UnitTests/ConnectionTest.cs
+++ b/Engine.UnitTests/ConnectionTest.cs
@@ -64,6 +64,13 @@ namespace OpenTap.Engine.UnitTests
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
Looks great

PR #1707, opened 2024-08-20
1705 Virtual Via Points: Fix

### LINKED ISSUE #1704: ViaPoint.IsActive should be virtual
If ViaPoint.IsActive was virtual it would be a lot easier to write code which dynamically creates via points based on instrument settings.

rmadsen-ks:
- Made ViaPoint.IsActive virtual.

- Added unit test to verify the behavior.

Close #1704


alnlarsen (approved):
I decompiled OpenTap.dll and I am pretty sure it is not breaking to make the property virtual since current property accessors are already using CallVirt even though the property is not currently Virtual. Let's get this merged.

PR #1708, opened 2024-08-20
1700 MultiSelect EmbedProperty error fix.

rmadsen-ks:
- Added unit test to show the issue, occuring when multi select is used with certain objects and errors are read out.


rmadsen-ks:
This breaks the functionality DateTime Annotation because it is both a string value annotation and a IErrorAnnotation.

PR #1717, opened 2024-08-28
Parallelize UpdatePrerelease queries

### LINKED ISSUE #1716: Parallelize graphql queries during image resolution
When updating prereleases for a package, we are currently querying all configured http repositories in sequence. We should parallelize this to improve performance.

alnlarsen:
This should speed up resolutions when using multiple http repositories

Closes #1716


rmadsen-ks (approved):
Looks great.

PR #1722, opened 2024-08-29
Serialize empty test step lists

### LINKED ISSUE #1721: Serializer does not set empty step lists
This is an issue in the edge case where a step adds child steps in its constructor, and the test plan is saved after the child steps are removed.

Because we do not serialize an empty <ChildTestSteps/> element, we do not call the setter for the `ChildTestSteps` property with an empty list.

This was likely introduced by https://github.com/opentap/opentap/pull/588

This doesn't seem to affect generic lists. Just TestStepList

alnlarsen:
This fixes a missing <ChildTestSteps/> element in serialized test steps. This solves a serializer issues that affects steps that add child steps in their constructor, which are re-added if a test plan is reloaded after removing those child steps.

Closes #1721


rmadsen-ks (approved):
Looks great!

PR #1736, opened 2024-09-09
1684 SetAssemblyVersion adds references: Fix

### LINKED ISSUE #1684: Linux builds: SetAssemblyInfo can pull in System.Private.CoreLib as a dependency
This happens if the XVersion attributes are not set inside the assembly and the mono assembly version injector needs to add them instead of modify ones already existing.

This means there is a workaround for all projects: just set `<GenerateAssemblyInfo>true</GenerateAssemblyInfo>` (default).

Setting this to false seems to be what caused the issue for the Unit-Test plugin.

rmadsen-ks:
Added code to prevent that adding e.g AssemblyFileVersionAttribute adds a reference to private.corelib.

Close #1684

### DIFF
--- a/Package/SetAsmInfo/SetAsmInfo.cs
+++ b/Package/SetAsmInfo/SetAsmInfo.cs
@@ -369,6 +369,9 @@ namespace OpenTap.Package.SetAsmInfo
                 asm.Name.Version = version;
                 anyWritten = true;
             }
+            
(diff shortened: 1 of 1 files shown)


alnlarsen (approved):
Looks good, I can't see how this could go wrong :fire: :fire: :fire:

PR #1738, opened 2024-09-10
Improve build architecture matching to allow cross compiling

alnlarsen:
This allows building e.g. a 32 bit plugin on an x64 machine, or an x64 plugin on an arm64 machine, by specifying `<PlatformTarget .../>`


rmadsen-ks (dismissed):
This looks good.

rmadsen-ks (approved):
Even better!

PR #1757, opened 2024-09-23
1756 CableLossMemoryLeak: Fix

### LINKED ISSUE #1756: Intermittent memory leak with Connections and Cable Losses
To reproduce the issue, create a connection and add 1000s of cable losses.

If the settings dialog is closed it gets cleaned up, but without closing it, these can leak due to an event handler inside Connection.

rmadsen-ks:
Close #1756 

- Fixed the issue by not registering an event handler for every connection created.
- Optimized the memory usage and initialization for cable loss points, by overriding GetError directly.


Oeh1995 (approved):
looks good!

PR #1759, opened 2024-09-24
1758 cable loss interpolation with binary search

rmadsen-ks:


rmadsen-ks:
### Testing

Unit tests already verify the behavior. However, there should be a test verifying cable loss calculations.

PR #1761, opened 2024-09-26
1737 Allow adding mixins to test plan reference

### LINKED ISSUE #1737: Can't add Mixins to Test Plan Reference steps
It is not possible to add Mixins to Test Plan Reference steps. This happens because we don't allow adding mixins to "locked" steps. The way we currently check if a step is "locked" is by checking `ChildTestSteps.IsReadOnly`. This value is always true for `TestPlanReference`, and any other step which needs to internally control its child steps. 

We need a different mechanism for determining if mixins can be added in order to support mixins on steps like `TestPlanReference`.

rmadsen-ks:
- Modified the rules on AddMixin to support adding mixins to test plan reference.
- Added unit test to verify the behavior.

Close #1737

### DIFF
--- a/Engine.UnitTests/MixinTests.cs
+++ b/Engine.UnitTests/MixinTests.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading;
 using NUnit.Framework;
 using OpenTap.Plugins.BasicSteps;
+using Tap.Shared;
 namespace OpenTap.UnitTests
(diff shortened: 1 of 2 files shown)


Oeh1995 (approved):
Looks good!

sebastian-pop:
Implemented. Verified on 9.27.0-rc.2

image

PR #1763, opened 2024-09-26
1762 Embed Test Plan Reference: fix

### LINKED ISSUE #1762: Add a button for converting test plan reference to a sequence step
Add a button to test plan reference to enable converting it to a sequence step, embedding the test steps directly in the test plan.

rmadsen-ks:
- Added code to convert a test plan reference to a sequence step.
- Added handling for mixins and parameters
- Not working: If a setting is parameterized to a parent.

image

image

image

Close #1762

### DIFF
--- a/BasicSteps/TestPlanReference.cs
+++ b/BasicSteps/TestPlanReference.cs
@@ -386,15 +386,11 @@ namespace OpenTap.Plugins.BasicSteps
             
             var refPlanPath = Filepath.Expand(testPlanDir: testplandir as string);
             refPlanPath = refPlanPath.Replace('\\', '/');
-            if (!File.Exists(refPlanPath))
-            {
-                Log.Warning("File does not exist: \"{0}\"", refPlanPath);
-                return null;
-            }
+            
             return refPlanPath;
         }
 
-        public bool anyStepsLoaded => ChildTestSteps.Any() && File.Exists(GetPath());
+        public bool anyStepsLoaded => ChildTestSteps.Any() && GetPath() is string path && File.Exists(path);
 
         
         
@@ -418,7 +414,7 @@ namespace OpenTap.Plugins.BasicSteps
             
         
         [Browsable(true)]
-        [Display("Convert to Sequence", "Convert  the test plan reference to a sequence step.", Order: 1.1)]
+        [Display("Convert to Sequence", "Convert the test plan reference to a sequence step.", Order: 1.1)]
         [EnabledIf(nameof(anyStepsLoaded), HideIfDisabled = true)]
         public void ConvertToSequence()
         {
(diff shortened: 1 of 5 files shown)


BasicSteps/TestPlanReference.cs
```diff
@@ -376,6 +376,97 @@ public void loadTestPlan()
...
+                Log.Warning("File does not exist: \"{0}\"", refPlanPath);
+                return null;
+            }
+            return refPlanPath;
+        }
+
+        public bool anyStepsLoaded => ChildTestSteps.Any() && File.Exists(GetPath());
+        
+        [Browsable(true)]
+        [Display("Convert to Sequence", "Convert  the test plan reference to a sequence step.", Order: 1.1)]
```
alnlarsen, line 400: Fix double space:
```suggestion
        [Display("Convert to Sequence", "Convert the test plan reference to a sequence step.", Order: 1.1)]
```

BasicSteps/TestPlanReference.cs
```diff
@@ -376,6 +376,122 @@ public void loadTestPlan()
...
+        [EnabledIf(nameof(anyStepsLoaded), HideIfDisabled = true)]
+        public void ConvertToSequence()
+        {
+            var warn = new ConvertWarning();
+            UserInput.Request(warn, true);
+            if (warn.Response == ConvertWarning.ConvertOrCancel.Cancel)
+                return;
+            var subPlan = TestPlan.Load(GetPath());
```
alnlarsen, line 429: `GetPath()` returns null if the test plan can't be found. We should give a prettier error message than whatever `TestPlan.Load()` does with a null argument.

alnlarsen: `GetPath()` already logs a warning in the null case. It seems more correct to log the warning here

BasicSteps/TestPlanReference.cs
```diff
@@ -376,6 +376,122 @@ public void loadTestPlan()
...
+            object testplandir = null;
+            var currentSerializer = TapSerializer.GetObjectDeserializer(this);
+            if (currentSerializer != null && currentSerializer.ReadPath != null)
+                testplandir = System.IO.Path.GetDirectoryName(currentSerializer.ReadPath);
+            
+            var refPlanPath = Filepath.Expand(testPlanDir: testplandir as string);
+            refPlanPath = refPlanPath.Replace('\\', '/');
+            if (!File.Exists(refPlanPath))
+            {
+                Log.Warning("File does not exist: \"{0}\"", refPlanPath);
```
alnlarsen, line 391: Remove this and log warnings at call sites?

BasicSteps/TestPlanReference.cs
```diff
@@ -376,6 +376,122 @@ public void loadTestPlan()
...
+        public void ConvertToSequence()
+        {
+            var warn = new ConvertWarning();
+            UserInput.Request(warn, true);
+            if (warn.Response == ConvertWarning.ConvertOrCancel.Cancel)
+                return;
+            var subPlan = TestPlan.Load(GetPath());
+            var seq = new SequenceStep
+            {
+                Name = Name != "Test Plan Reference" ? Name : "Sequence",
```
alnlarsen, line 432: It is potentially a bit confusing that a step named `Test Plan Reference (1)` does not get renamed. Should we update the logic a bit to handle our automatic name collision prevention?

alnlarsen (changes requested):
Tried it out, and it seems to work well. Regarding parameterizations from the test plan reference to e.g. parent steps not working, I guess this is acceptable, but currently it leaves the test plan in a weird state where the parameter appears on the parent, and there is no way to remove it.

Would it be very complicated to traverse the parent hierarchy and check if any members are controlled by the parent?

BasicSteps/TestPlanReference.cs
rmadsen-ks: This was implemented

rmadsen-ks:
@alnlarsen, can you check this again?

alnlarsen (approved):
This looks awesome, and works well!

Denrask:
This is confirmed to work flawlessly in KS8500. Well done!

PR #1767, opened 2024-09-30
Use Reference instead of ReferencePathWIthRefAssemblies

alnlarsen:
For unclear reasons, WPF intermediate builds don't pick up references added with ReferencePathWithRefAssemblies, but references added with the Reference element are added just fine.

This fixes an issue that causes WPF builds using OpenTapPackageReference to fail the first time.

We currently have workarounds in various WPF projects that involve setting up a dummy project that installs packages, and then manually referencing assemblies from those packages. This should no longer be needed.


rmadsen-ks (approved):
Looks good!

PR #1769, opened 2024-10-01
1768 TestStepList Raise Event On Set: Fix

### LINKED ISSUE #1768: TestStepList API does not raise an event when an element is set.
When items are added or removed from a TestStepList, no event gets raised. This gives trouble for UIs that want to notify the user when the test plan has changed.

rmadsen-ks:
Added that SetItem and MoveItem raises an event to signal that the list has changed.

Added unit tests to verify the functionality.

Close #1768


alnlarsen (approved):
Looks fine

PR #1772, opened 2024-10-04
1687 build time gitversion

### LINKED ISSUE #1687: Build-time Assembly Versions
We want to be able to automatically set the assembly version by opting into a git versioning feature in the csproj file. Like this
```xml
<PropertyGroup> 
    <OpenTapGitAssistedAssemblyVersion>
        <GitVersion><!-- this is optional--> $(VersionEnv)</GitVersion> <!-- if GitVersion is empty, compute it, otherwise use it -->
   </OpenTapGitAssistedAssemblyVersion>
    <GenerateAssemblyInfo>true</GenerateAssemblyInfo> <!-- required for the feature to work -->
</PropertyGroup>
```
```xml
<PropertyGroup>  <!-- the lazy variant -->
    <OpenTapGitAssistedAssemblyVersion>true</OpenTapGitAssistedAssemblyVersion>
    <GenerateAssemblyInfo>true</GenerateAssemblyInfo> <!-- required for the feature to work -->
</PropertyGroup>
```

The metrics plugin already does something like this, but in a more manual fashion.

Additional requirements: 
- if GenerateAssemblyInfo is set to false, fail with a user-friendly build error.
- If GitVersion is not set, and its not inside a git repository, fail with a user-friendly build error.

alnlarsen:
This adds support for versioning assemblies with GitVersion during build. This is an alternative to versioning assemblies in the packaging stage with `<File Path="my.dll"><SetAssemblyInfo Attributes="Version"/></File>`

This also correctly sets the version information in the appropriate PE sections (example from local test with the CSV plugin) on both Windows and Linux so it shows up in the Properties tab of a file. This is an improvement on our current `SetAssemblyInfo` implementation which only works on Windows because it relies on P/Invoke APIs.

image

Closes #1687

### DIFF
--- a/doc/Developer Guide/Getting Started in Visual Studio/Readme.md
+++ b/doc/Developer Guide/Getting Started in Visual Studio/Readme.md
@@ -120,7 +120,7 @@ conditionally control it through an environment variable:
   <OpenTapSetAssemblyVersion Condition="'$(SomeVersionFromEnvironment)' == ''">true</OpenTapSetAssemblyVersion>
   <!-- Otherwise use the version from the environment variable -->
   <OpenTapSetAssemblyVersion Condition="'$(SomeVersionFromEnvironment)' != ''">
-    <GitVersion>$(SomeVersionFromEnvironment)</GitVersion>
+    <Version>$(SomeVersionFromEnvironment)</Version>
   </OpenTapSetAssemblyVersion>
 </PropertyGroup>
 ```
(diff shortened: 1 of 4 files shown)


AsgerIversen (changes requested):
I see the point in the feature. Let's get that feature in. Only comment is related to naming... In the case where you specify a version string manually, the naming is a bit weird.

rmadsen-ks (approved):
Looks great!

PR #1774, opened 2024-10-04
Update step IDs on SetItem

### LINKED ISSUE #1773: TestStepList: Update ID Lookup on set item
Update the ID when SetItem is called on TestStepList.

rmadsen-ks:
Close #1773

### DIFF
--- a/Engine/TestStepList.cs
+++ b/Engine/TestStepList.cs
@@ -158,6 +158,20 @@ namespace OpenTap
                     idLookup[thing.Id] = thing;
                 }
             }
+            else if (Action == ChildStepsChangedAction.SetStep)
+            {
+                if (idLookup.TryGetValue(Object.Id, out var prev))
+                {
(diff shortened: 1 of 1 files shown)


alnlarsen (changes requested):
I can't really follow the logic here.

It seems to me that the `SetStep` action should

1. Remove the IDs in the replaced steps hierarchy
2. Add the IDs in the new steps hierarchy

But that is not what happens here.

rmadsen-ks:
> I can't really follow the logic here.
> 
> It seems to me that the `SetStep` action should
> 
> 1. Remove the IDs in the replaced steps hierarchy
> 2. Add the IDs in the new steps hierarchy
> 
> But that is not what happens here.

All we are doing here is updating the cache which relates test step IDs to objects. When SetItem is used, there is a chance that the new step has the same ID as the old one. That is OK, but in this case the cache needs to be updated.

alnlarsen (approved):
Looks good!

Issue #1775, opened 2024-10-05
TapAssemblyResolver.Invalidate does not invalidate assemblies


PetrJasek89:
`TapAssemblyResolver.Invalidate` method (inside `PluginManager.cs`) should either call `assemblyResolutionMemorizer.InvalidateAll()` or `assemblyResolutionMemorizer.InvalidateWhere((k,v) => v == null)`. I think `InvalidateAll` makes more sense as all assemblies loaded until now should no longer be valid because new `DirectoriesToSearch` are set.

Currently, if `PluginManager.DirectoriesToSearch` is changed and `PluginManager.Search` is called the assemblies will not get invalidated. This is because `TapAssemblyResolver.Invalidate` is called first and it sets `lastSearchedDirs = FileFinder.DirectoriesToSearch.ToHashSet()`. Then, in `TapAssemblyResolver.GetAssembliesToSearch` the condition `if (false == lastSearchedDirs.SetEquals(FileFinder.DirectoriesToSearch))` is never true so the `assemblyResolutionMemorizer.InvalidateWhere((k, v) => v == null)` is not called.

The issue might be potentially related to #108.

rmadsen-ks:
This was fixed by #1777.

PR #1777, opened 2024-10-05
1775 Fix TapAssemblyResolver.Invalidate

PetrJasek89:
Fix for #1775: Invalidate assemblies when TapAssemblyResolver.Invalidate is called


rmadsen-ks (approved):
This looks good. Lets get it in!.

Issue #1186, opened 2023-07-21
Error when selecting two `Sweep Loop` steps


sebastian-pop:
**Describe the bug**

**Environment (please complete the following information):**

 - OS: Windows 10
 - OpenTAP Version 9.21.1

**To Reproduce**

1. Add two Sweep Loop steps each with a Delay child
2. On one or both Sweep steps set Sweep Mode to Across Runs
3. Select both Sweep steps

**Actual behavior**

Error in log: [Latest.txt](https://github.com/opentap/opentap/files/12119128/Latest.txt)

![image](https://github.com/opentap/opentap/assets/97734470/47b278e8-f112-41b8-8651-55563b8e8857)

**Expected behavior**

**Additional context**

Add any other context about the problem here. Please also attach a log file from <InstallDir>\SessionLogs\ that shows the issue when possible/relevant.

PR #1784, opened 2024-10-09
Update OS detection logic in package identifier

### LINKED ISSUE #1783: Stop trying to parse OS information from RuntimeInformation.OSDescription
We are currently using this to guess the host OS, but it doesn't work in all cases. Microsoft does not recommend parsing this value, and on some platforms, this variable is based on the content of `/etc/os-release`, which could be modified.

We should stop using parsing this value.

See documentation of OSDescription for more details:

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osdescription?view=net-8.0

alnlarsen:
This fixes an issue caused by relying on the string output of `OSDescription`, which according to MSDN, cannot be relied on. This makes sense because on Linux, it just parses the file `/etc/os-release`, which any distribution can modify as they like.

Closes #1783

### DIFF
--- a/Package/IPackageIdentifier.cs
+++ b/Package/IPackageIdentifier.cs
@@ -4,7 +4,6 @@
 // file, you can obtain one at http://mozilla.org/MPL/2.0/.
 using System;
 using System.Linq;
-using System.Runtime.InteropServices;
 
 namespace OpenTap.Package
 {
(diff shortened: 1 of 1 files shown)


Oeh1995 (dismissed):
Looks good!

Oeh1995:
Also closes #1186

PR #1785, opened 2024-10-09
go back to self hosted windows runners

alnlarsen:
Windows code build runners take up to 10 minutes to start. Let's go back to the self-hosted windows runner.


rmadsen-ks (approved):
Looks good!

PR #1788, opened 2024-10-10
1765 improve system runtime exceptions

### LINKED ISSUE #1765: Improve error message when .NET 6 runtime dlls cannot be loaded
Currently, we get a stack trace which is not really understandable by most users. Let's add a special case to our resolver to let the user know why it cannot be loaded.

alnlarsen:
This adds special handling when `System.Runtime` cannot be loaded, since this always means we tried to load an assembly which depends on dotnet 6+ in a .NET Framework process

Closes #1765

### DIFF
--- a/Engine/TypeData.cs
+++ b/Engine/TypeData.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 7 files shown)


Engine/TypeData.cs
```diff
@@ -348,8 +349,31 @@ public Type Load()
...
+                    var versionPrefix = "System.Runtime, Version=";
+                    if (filename.StartsWith(versionPrefix, StringComparison.OrdinalIgnoreCase))
+                    {
+                        var commaIndex = filename.IndexOf(',', versionPrefix.Length);
+                        if (commaIndex == -1)
+                        {
+                            commaIndex = filename.Length;
+                        }
+                        var versionString = filename.Substring(versionPrefix.Length, commaIndex - versionPrefix.Length);
+                        dependencyVersion += $" {versionString}";
```
rmadsen-ks, line 370: if commaIndex is -1, the versionString will be empty, right?1
```suggestion
                        if (commaIndex == -1)
                        {
                            // no comma so it ends with "Version=X.Y.Z.W"
                            commaIndex = filename.Length;
                        }
                        var versionString = filename.Substring(versionPrefix.Length, commaIndex - versionPrefix.Length);
                        dependencyVersion += $" {versionString}";
```

Engine/TypeData.cs
```diff
@@ -348,8 +349,31 @@ public Type Load()
...
-                    ex.Message);
+                var reason = ex.Message;
+                // Create a special-case error message when the load error is due to a dotnet runtime mismatch.
+                // FileName is a qualified assembly name, e.g. System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
+                if (ex is FileNotFoundException notFound && notFound.FileName.StartsWith("system.runtime", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Build the reason string. I don't know if it's possible for the assembly to not be strong-named,
+                    // but let's handle that just in case.
+                    var filename = notFound.FileName;
+                    var dependencyVersion = ".NET";
```
rmadsen-ks, line 360: Maybe let this be null and default to writing another error the file is e.g "System.RuntimeInformation...."

rmadsen-ks (changes requested):

Engine/TypeData.cs
alnlarsen: No, if commaIndex is -1, then there was no trailing comma. In the line above we already checked that the line contains `Version=`. 

This is meant to handle the edge case where the AssemblyName was not strongnamed, though I'm not sure if that can even happen.

rmadsen-ks (approved):
Looks good!

PR #1790, opened 2024-10-11
1789 sweep iteration parameter

### LINKED ISSUE #1789: Sweeps should include an iteration parameter
To be able to distinguish iterations from a results perspective, various sweep loops could use a parameter to store the iteration number.

It could also be done from the result listener, but that would be a bit unsafe since the user could decide to jump to a given test step.

Having an iteration number would be the most robust solution.

rmadsen-ks:
Added Sweep parameter iteration Resultparameter.

It is not possible to distinguish between iterations from a result listener point of view.

Close #1789


alnlarsen (approved):
Looks great!

PR #1801, opened 2024-10-22
1798 element factory

### LINKED ISSUE #1798: Add IElementFactory for lists with special behavior
Sweep Parameter Step has a SweepRowCollection type which contains SweepRows. That is done because SweepRow needs a reference to the test step it belongs to. 

However, it does not know about the step until it is added to the collection, which makes it hard to figure out which columns SweepRows has when generating a GUI. It would be simpler if there was an API on the test step or list, which tells OpenTAP how to make an instance of the elements, given its context.

It could look like this:

```cs

    /// <summary>
    /// Defines how the class can create list elements of a given type Or for a given member.
    /// If its the class having the list property, elementContext will be a IMemberData
    /// If its the list containing the elements, it will be an ITypeData.
    /// </summary>
    public interface IElementFactory
    {
        /// <summary> Creates a new element for the list. </summary>
        object NewElement(IReflectionData elementContext);
    }
```

rmadsen-ks:
Close #1798


alnlarsen (changes requested):
This looks really awesome!

Engine/SerializerPlugins/ObjectSerializer.cs
```diff
@@ -80,10 +80,18 @@ public virtual bool TryDeserializeObject(XElement element, ITypeData t, Action<o
                     {
                         throw new Exception($"Cannot create instance of {t} and no default value exists.");
                     }
+                  
+                    if (ownerMember.HasAttribute<IFactoryAttribute>())
+                    {
+                        newobj = FactoryAttribute.Create(ownerObj, ownerMember.GetAttribute<IFactoryAttribute>());
+                    }
```
alnlarsen, line 87: It looks like `IFactoryAttribute` would match both a `FactoryAttribute` and `ElementFactoryAttribute`, right? In this example:

```csharp
[ElementFactory("SomeMethod")]
public List<object> MyList { get; set; }
```
We would end up trying to use the element factory to construct MyList, which does not seem correct.

I think we should only check FactoryAttribute here

PR #1806, opened 2024-10-29
1805 prerun mixin parameters not updated

### LINKED ISSUE #1805: Properties changed in TestStepPreRunEvent not reflected in result parameters
This is an issue for use cases where parameters are being updated by a mixin: Notice that parameters are already initialized before TestStepPreRunEvent is invoked:

image

I am a bit concerned about just swapping the order since mixins could expect the parameters to already be initialized. It would probably be safest to re-calculate the parameters after invoking the mixin.

alnlarsen:
This fixes an issue causing updates to test step parameters in pre-run mixins to not be reflected in `StepRun.Parameters` in the `OnTestStepRunStart` event in result listeners

Closes #1805


rmadsen-ks (approved):
Looks great!

PR #1809, opened 2024-10-31
Create timestamps based on stopwatches instead of UtcNow

### LINKED ISSUE #1808: Low clock resolution in session logs on Windows
Our current log timestamper (called `AccurateStamper`) is not at all accurate. It has an error of ~10ms on Windows (but not Linux). Log messages that occur within 10ms of each other will have identical timestamps in the log on Windows. We should use stopwatches to measure time instead of `Datetime.Utc.Now.Ticks`. See log file and notice identical timestamps:

image

alnlarsen:
Measure time with stopwatches instead of UtcNow.Ticks

See remarks: https://learn.microsoft.com/en-us/dotnet/api/system.datetime.utcnow?view=net-8.0#remarks

The system timer can be anything between 0.5ms to 15 ms. Session logs specify timestamps with 6 significant digits which is a bit misleading if we can only provide 2-3.

Closes #1808

### DIFF
--- a/Engine/Logging/diag_impl.cs
+++ b/Engine/Logging/diag_impl.cs
@@ -19,10 +19,16 @@ namespace OpenTap.Diagnostic
         /// <summary> Prints a friendly name. </summary>
         /// <returns></returns>
         public override string ToString() => "Local and Accurate";
+        
+        // DateTime.UtcNow is not very accurate (Roughly 10-15ms accuracy)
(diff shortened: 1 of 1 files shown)


witoldpiet:
Good stuff!

alnlarsen:
This is not really a good solution. For long running test plans, this will drift. DateTime should be preferred for accurate time stamps. Furthermore, the precision is greatly improved in .NET 6, so once we move this will be resolved.

PR #1822, opened 2024-11-13
1821 External Parameter and Column Display Name attribute should not …

### LINKED ISSUE #1821: ColumnDisplayNameAttribute and ExternalParameterAttribute should not follow parameters
These attributes are generally created because something should happen to that specific property, but once it has been parameterized it  should be considered a setting and not be affected by these attributes.

rmadsen-ks:
…follow parameterization

Fixed the issue by disabling the behavior.

Close #1821


alnlarsen (approved):
Seems to work great!

PR #1826, opened 2024-11-13
1824 Multi-Select Convert to Sequence: Fix

### LINKED ISSUE #1824: Allow multi-select Convert to sequence
When converting many test plan reference steps it is bothersome to continously click proceed.

Instead we should just ask for confirmation for the first time.

rmadsen-ks:
- Added support for multi-selecting test plan reference and converting many to sequences with just a couple of clicks.

- Added IMergedValueAnnotation to be able to show annotators when we are dealing with merged data.
- Added IMergedMethodAnnotation for exactly the case where we can manually call the method in a combined way.

Close #1824


alnlarsen (approved):
Looks great

PR #1836, opened 2024-11-19
Fix image installer unpacking non zip files

### LINKED ISSUE #1835: Image installer tries extracting local files with names matching a requested package
Consider this logic from the image installer:

```csharp
if (File.Exists(package.Name))
{
    var package2 = PackageDef.FromPackage(package.Name);
    toInstall.Add(new PackageSpecifier(package2.Name, package2.Version.AsExactSpecifier()));
    installed.Add(package2);
    continue;
}
toInstall.Add(package);
```

If there is a local file named 'OpenTAP', and an image is trying to install 'OpenTAP', then this logic will attempt to read the file 'OpenTAP' as a zip archive. This is obviously not going to end well.

This logic is intended to make the image installer support specifying local packages, e.g. `./packages/OpenTAP.TapPackage`. Let's make this check a bit smarter.

alnlarsen:
Closes #1835

### DIFF
--- a/Package/Image/ImageSpecifier.cs
+++ b/Package/Image/ImageSpecifier.cs
@@ -43,6 +43,16 @@ namespace OpenTap.Package
         public ImageSpecifier()
         {
             
+        } 
+        
+        private static bool IsZipFile(string path)
+        {
+            using var fs = File.OpenRead(path);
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Looks great!

PR #1840, opened 2024-11-20
1839 Allow Calling convert to sequence without a popup dialog

rmadsen-ks:
Added a check so that if user input is null, it will not show the popup for the user input dialog.

Close # 1839


alnlarsen (approved):
Looks good

PR #1844, opened 2024-11-21
1842 test plan reference with embed properties mixins

### LINKED ISSUE #1842: TestPlanReference "Convert to Sequence" button does not transfer mixins that use the EmbedPropertiesAttribute
**Describe the bug**

The "Convert to Sequence" button on TestPlanReference test step does not transfer mixins that have the EmbedPropertiesAttribute.

 - OS: Windows 10
 - OpenTAP Version: 9.27.0-rc.4
 - Basic Mixins: 0.2.0-alpha.1.1+248e8ea6.pull-14-merge

**To Reproduce**

1. Add a Test Plan Reference test step.

2. Select any tapplan to reference.

3. Add a Limit mixin.

4. Click the "Convert to Sequence" button on the Test Plan Reference.

**Actual behavior**

Before clicking "Convert to Sequence":
image

The referenced test plan in the above image only has a Delay step, which has its Time Delay parameterized on the Test Plan.
The Test Plan Reference test step has a Number mixin, and a Limit mixin.

After clicking "Convert to Sequence":
image

Only the Number mixin is transferred.

When testing with only the Limit mixin, no mixins are transferred.

**Expected behavior**

Both the Number mixin and the Limit mixin should be transferred.

rmadsen-ks:
Added unit tests.
Tested this with KS8400.

Close #1842

### DIFF
--- a/BasicSteps/TestPlanReference.cs
+++ b/BasicSteps/TestPlanReference.cs
@@ -354,8 +354,10 @@ namespace OpenTap.Plugins.BasicSteps
                                     }
                                     continue;
                                 }
+                                // transfer  the value from the test plan instance to this instance.
+                                var value = member.GetValue(tp);
                                 DynamicMember.AddDynamicMember(this, mem);
-                                mem.SetValue(this, member.GetValue(this));
+                                mem.SetValue(this, value);
                             }
                         }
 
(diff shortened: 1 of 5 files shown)


Oeh1995 (approved):
Looks good!

PR #1848, opened 2024-11-25
1847 modify dialog: use display name if present

### LINKED ISSUE #1847: Modify Mixin: Use Display Name if set
Currently, the `Modify Mixin` dialog uses the mixin member name as the dialog title. This is fine as a fallback, but I think we should prefer to use `[DisplayAttribute]` if it is available instead since this is more likely to contain a name intended to be presented to the user.

alnlarsen:
Closes #1847


rmadsen-ks (approved):
Looks great!

PR #1851, opened 2024-12-02
1846 multi level scope bug

### LINKED ISSUE #1846: Parameters with multiple levels of scope cannot be copied
In the following test plan, if the top sequence step is copied, the parameter does not follow.

image

rmadsen-ks:
Close #1846


alnlarsen (approved):
This seems to work well now!

PR #1854, opened 2024-12-05
stop deserializing parameterized members with factories

### LINKED ISSUE #1853: Serializer error when parameterizing Sweep Values on test plan
This is a regression introduced in 9.27.0 by #1801

alnlarsen:
Using factories for parameters should not be necessary in the first place since they should have already been created with factories in the source step.

Also add a unittest which ensures parameterized factory elements can be deserialized without errors

Closes #1853


rmadsen-ks (approved):
Looks good

PR #1859, opened 2024-12-11
Fixes an issue with duplicate packages in image resolution

### LINKED ISSUE #1858: Image Resolution Duplicate Package
**Describe the bug**

If I ask to have an image resolved with two identical packages, the resulting resolution also contains two identical packages.

 - OS: PoopOS
 - OpenTAP Version: 9.27.0

**To Reproduce**

1. Have Runner 1.3.1 in your local package cache `~/.local/share/OpenTap/PackageCache`

2. Run `./tap image install --target ../Testing "Runner:1.3.1,Runner:1.3.1" --dry-run -v --repository ~/.local/share/OpenTap/PackageCache`

**Actual behavior**

There are two Runner packages in the resulted image:

image

Denrask:
Fixes #1858

### DIFF
--- a/Package/Image/ImageResolver.cs
+++ b/Package/Image/ImageResolver.cs
@@ -74,6 +74,7 @@ namespace OpenTap.Package
                         {
                             // select pkg2 instead of pkg1
                             packages[i] = pkg2;
+                            pkg1 = pkg2;
                             packages.RemoveAt(j);
                             goto retry;
                         }
--- a/Package/Image/ImageResolver.cs
+++ b/Package/Image/ImageResolver.cs
@@ -42,24 +42,7 @@ namespace OpenTap.Package
                         {
                             packages.RemoveAt(j);
                             goto retry;
-                        }
-                        
-                        // In the case where the versions are the same, but the match behavior is different
-                        // If one is exact and the other is compatible, we can remove the compatible one.
-                        if (pkg1.Version.CompareTo(pkg2.Version) == 0)
-                        {
-                            if (pkg1.Version.MatchBehavior == VersionMatchBehavior.Exact && pkg2.Version.MatchBehavior == VersionMatchBehavior.Compatible)
-                            {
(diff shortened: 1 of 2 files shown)


Package/Image/ImageResolver.cs
```diff
@@ -37,6 +37,30 @@ public ImageResolution ResolveImage(ImageSpecifier image, PackageDependencyGraph
...
+                            if (pkg1.Version.MatchBehavior == VersionMatchBehavior.Exact && pkg2.Version.MatchBehavior == VersionMatchBehavior.Compatible)
+                            {
+                                packages.RemoveAt(j);
+                                goto retry;
+                            }
+                        
+                            if (pkg2.Version.MatchBehavior == VersionMatchBehavior.Exact && pkg1.Version.MatchBehavior == VersionMatchBehavior.Compatible)
+                            {
+                                packages.RemoveAt(i);
+                                goto retry;
```
alnlarsen, line 60: This looks like a bug. 

`pkg1` is assigned right before the `retry` label, and this goto will continue the inner loop without reassigning `pkg1`.

This should probably be:

```
packages[i] = packages[j];
packages.RemoveAt(j);
goto retry;
```

alnlarsen (dismissed):
I am not sure if this is the correct place to solve the problem. Otherwise the logic looks fine except for the bug I mentioned.

It also seems a bit weird that our Version comparison implementation does not check the match behavior, and I don't like the fact that this implementation relies on that behavior.

Package/Image/ImageResolver.cs
```diff
@@ -37,6 +37,30 @@ public ImageResolution ResolveImage(ImageSpecifier image, PackageDependencyGraph
...
+                            }
+                        }
+
```
alnlarsen, line 63: It doesn't look like this is needed at all?

rmadsen-ks (dismissed):
Looks good!

PR #1861, opened 2024-12-11
Fix bug causing image resolutions from local caches to fail

### LINKED ISSUE #1865: Image resolver version specifier behavior differs between local cache and http repositories
This happens because the XML parser appends `MatchBehavior.AnyPrerelease` to version specifiers using the 'compatible' (`^`) match behavior. This relaxes the constraints used by the image resolver to allow using an older (potentially incompatible) prerelease version of a plugin than the one required by a package.

Specifically, this can happen in the following image configuration, under specific circumstances:

* Image: `PlugA,PlugB:^0.1.0`
When the following conditions are met:
1. PlugA depends on PlugB version 0.2.0-beta.2
2. PlugA is cached locally
3. PlugB version 0.2.0-beta.1 is cached locally
4. The image is resolved with only local caches

In this case, the `AnyPrerelease` specifier added by the XML deserializer causes the image to succesfully resolve the incompatible beta version of PlugB. The correct behavior in this case is to fail the image resolution.

Let's figure out a way to resolve this without causing splash damage.

alnlarsen:
* Update VersionSpecifier compatibility check to now accept prereleases for underspecified majors
This is more in line with underspecified minors. Currently, ^1.0 matches 1.2.3-beta.1, but ^1 does not. With this fix, it will.
* Update PackageDefinitionSerializer to no longer add `MatchBehavior.AnyPrerelease` to `Compatible` specifiers.
I tracked down the original commit, and found that this behavior was added in order for OpenTAP to stop complaining about 
version compatibility issues when e.g. 9.27.0-beta.1 is installed in an installation with e.g. Demonstration, which depends on OpenTAP ^9. With this updated versin matching, this should no longer be an issue.

I also added a bunch of unit test to verify new behavior.

Closes #1865

### DIFF
--- a/Engine.UnitTests/SemanticVersionTests.cs
+++ b/Engine.UnitTests/SemanticVersionTests.cs
@@ -145,7 +145,15 @@ namespace OpenTap.Engine.UnitTests
         [TestCase("^9.0.0-alpha.2", "9.0.0-alpha.1", false)]
         [TestCase("^9.0.0-alpha.1", "9.0.0-alpha", false)]
         [TestCase("^9.0.1200-alpha.1.7", "9.0.1200-alpha.1.2", false)]
-        [TestCase("^9.1.0-alpha.1", "9.0.0-alpha.1", false)]
+        [TestCase("^9.1.0-alpha.1", "9.0.0-alpha.1", false)] 
+        [TestCase("^1", "0.1.0", false)] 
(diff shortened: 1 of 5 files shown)


rmadsen-ks (approved):
Looks good lets merge and test additionally with KS8400

PR #1862, opened 2024-12-13
fix resolver dependency fetching

### LINKED ISSUE #1863: Image resolver fails to populate prerelease caches
This bug occurs when the following conditions are met:

1. Package A depends on package B in a prerelease version
2. The input image specifies a release version of package B
3. A compatible version of B is not present in local caches

Under such circumstances, the resolver will deem the resolution a failure before the prerelease versions of B are fetched. We must make the dependency fetching slightly more eager.

alnlarsen:
This fixes an issue causing prerelease dependencies to not be fetched early enough during resolution.

I verified that this does not impact performance by an RC version of resolving Developer's System, which has quite a few prerelease dependencies.

Closes #1863


rmadsen-ks (approved):
Looks good!

rmadsen-ks (approved):

PR #1873, opened 2025-01-07
1872 fix deserializing factory elements 2

### LINKED ISSUE #1872: Sweep paramter as external paramters causes exception in a Test Plan Refernce
**Describe the bug**

Open a TAP plan with Test Plan Refernce where sweep paramters are external paramters:
image
When opening the plan you get 
15:27:20.357  Serializer   Caught error while finishing serialization: Object reference not set to an instance of an object.
15:27:20.363  Serializer   Caught error while finishing serialization: Object reference not set to an instance of an object.

 - OS: Windows
 - OpenTAP Version KS8400 9.27.1-RC.1

**To Reproduce**

I have attached an example plan here: [example.zip](https://github.com/user-attachments/files/18320553/example.zip)
Copy the delaySweep.TapPlan to C:\TEMP and open the masterPlan.TapPlan

**Additional context**
This is a followup to https://github.com/opentap/KS8400/issues/1309

alnlarsen:
This fixes an issue causing the factory logic to apply to all kinds of typedata implementations: ParameterMemberData, ExpandedMemberData, and any plugin-defined member data. MemberData from such typedata should have already been deserialized in the object that actually owns the member, and thus no special handling is required. 

This also allows removing the special-case for test plan parameters which was added in #1853

Closes #1872

### DIFF
--- a/Engine.UnitTests/ElementFactoryTest.cs
+++ b/Engine.UnitTests/ElementFactoryTest.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using NUnit.Framework;
 using OpenTap.Plugins.BasicSteps;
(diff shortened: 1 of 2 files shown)


Oeh1995 (approved):
Looks good!

PR #1881, opened 2025-01-14
Update version resolving

### LINKED ISSUE #1879: Image resolver: 'Basic Mixins:^0' resolves an alpha version instead of a beta version
This is an issue in the 9.27.1 rc which surfaced after we relaxed the resolver to allow e.g. '^0' to resolve prereleases. The version sorting logic considers alpha versions just as good as betas.

When this was discovered, the most recent versions of Basic Mixins was 0.2.0-alpha and 0.2.0-beta. In this scenario, we should prefer the beta version over the alpha version.

This was discovered by @tudor9507 in https://github.com/opentap/opentap/issues/1866#issuecomment-2579448751

### LINKED ISSUE #1880: Image resolver: '^0' only resolves alphas and release versions
Introduced in the rc version of 9.27.1

If for example a package only has a beta version (such as 'Docker Services'), installing it with `tap image install "Docker Services:^0"` fails because it cannot find any compatible versions, even though version `1.0.0-beta.9+3769e550` exists.

alnlarsen:
* Ensure '^0' resolves all prereleases, and not just alphas #1880
* Ensure release > rc > beta > alpha is preferred when no prerelease preference is specified #1879 

Closes #1879 
Closes #1880

### DIFF
--- a/Package.UnitTests/Image/DependencyGraphTest.cs
+++ b/Package.UnitTests/Image/DependencyGraphTest.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using NUnit.Framework.Constraints;
 
 namespace OpenTap.Image.Tests
 {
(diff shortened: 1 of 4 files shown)


Oeh1995 (approved):
looks good

PR #1885, opened 2025-01-16
1872 more edge cases

alnlarsen:
This fixes an edge case causing error messages in the log when loading a test plan on a test plan reference step with an empty Sweep Values list parameterized

This is another addon to #1872


Oeh1995 (approved):
looks good

PR #1886, opened 2025-01-16
1872 more edge cases final

alnlarsen:
Another regression related to #1872 

This fixes an issue causing changes made to sweep values parameterized on a test plan reference to not be persisted when clicking `Load Test Plan`, or when saving and reloading the test plan.


Oeh1995 (dismissed):
Amazing

PR #1891, opened 2025-01-20
1327 support XmlIgnore on embedded properties

### LINKED ISSUE #1327: Allow XmlIgnore on EmbedProperties
This would allow to separate the properties separately from the object which is being embedded.

```cs
    // This is the member I want to serialize, but it will never be shown to the user.
    [AnnotationIgnore]
    public EmbeddingClass[] Objects {get;set; }

    //the index of object can be selected
    public int SelectedObject{get;set:]

    [EmbedProperties]
    [XmlIgnore]
    public EmbeddingClass SelectedObject => Objects[SelectObject];

```

alnlarsen:
Closes #1327

### DIFF
--- a/Engine.UnitTests/EmbeddedPropertyTests.cs
+++ b/Engine.UnitTests/EmbeddedPropertyTests.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Linq;
+using System.Xml.Serialization;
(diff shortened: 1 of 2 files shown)


Oeh1995 (approved):
looks good

PR #1896, opened 2025-02-06
Fix null reference exception in gitversion calculation

### LINKED ISSUE #1895: null reference exception in `tap sdk gitversion

alnlarsen:
Closes #1895


StefanHolst (approved):
👍

PR #1900, opened 2025-02-20
Use RestartManager API for detecting files in use on windows

alnlarsen:
The Windows RestartManager API is intended for installers to be able to know which processes are preventing an installation from proceeding. By using this API in OpenTAP, we can give better hints to the user about which processes to close.

The RestartManager API also provides mechanisms for forcefully closing these applications, and restarting them again (if the application supports it), but I think we should be cautious about using these features due to the risk of data loss.

See screenshot of package manager with this feature enabled:

image


Package/Installer.cs
```diff
@@ -140,7 +152,18 @@ internal int UninstallThread()
...
         {
-            var verb = System.Globalization.CultureInfo.CurrentCulture.TextInfo.ToTitleCase(command.ToLower()) + "ed";
+            string titlecase(string str) => System.Globalization.CultureInfo.CurrentCulture.TextInfo.ToTitleCase(str);
+            
+            var verb = titlecase(command.ToLower()) + "ed";
+            // Example usages:
+            // "Tried to {commandFriendlyName} {pkg.Name}, but there was nothing to do."
+            // "There was an error while trying to {commandFriendlyName} '{pkg.Name}'."
+            var commandFriendlyName = command; 
+            if (string.Equals(command, PrepareUninstall, StringComparison.OrdinalIgnoreCase))
```
rmadsen-ks, line 162: RunCommand call with two different commands and inside the call there is a whole special case for one of them. Can we clean this up a bit and just hardcode the strings?

image

Package/Installer.cs
```diff
@@ -215,7 +239,11 @@ internal int RunCommand(string command, bool force, bool modifiesPackageFiles)
                     return (int)ExitCodes.UserCancelled;
 
                 log.Debug(ex);
-                return (int) ExitCodes.GeneralException;
+                return (int)ExitCodes.GeneralException;
+            }
+            finally
+            {
+                postInstallHook();
```
rmadsen-ks, line 246: It seems that postInstallHook is never actually used for anything, so maybe just delete it.

Package/Installer.cs
```diff
@@ -251,8 +279,51 @@ private FileInfo[] GetFilesInUse(string tapDir, List<string> packagePaths)
             return filesInUse.ToArray();
         }
 
+        private Action postInstallHook = () => { };
+
+        private void WaitForPacakgeFilesFreeWindows(List<string> packagePaths)
```
rmadsen-ks, line 284: "Pacakge"

Package/Installer.cs
```diff
@@ -251,8 +279,51 @@ private FileInfo[] GetFilesInUse(string tapDir, List<string> packagePaths)
...
+        {
+            var allfiles = packagePaths.SelectMany(PluginInstaller.FilesInPackage).ToArray();
+
+            retry:
+            var procs = RestartManager.GetProcessesUsingFiles(allfiles);
+            if (procs.Count == 0)
+                return;
+            var msg = new StringBuilder();
+            msg.AppendLine("The following applications are blocking the installation:");
```
rmadsen-ks, line 293: This is a bit counterintuitive when it happens during uninstall.

Maybe replace "installation with "The operation".

rmadsen-ks (approved):
Works well!

PR #1901, opened 2025-02-20
1899 keep dependency errors when overriding dependency

### LINKED ISSUE #1899: No dependency incompatibility errors when a package dependency is manually specified
Normally, when a plugin requires a dll which is incompatible with another installed plugin, we produce this kind of error:

```
OpenTap.Runner.dll in this package require assembly Newtonsoft.Json in version 12.0.0.0 while that assembly is already installed through package 'OpenTAP' in version 13.0.0.0.
Please align the version of Newtonsoft.Json (13.0.0.0 vs 12.0.0.0)  to ensure interoperability with package 'OpenTAP' or uninstall that package.
```

However, if e.g. the OpenTAP dependency version is manually specified, compatibility checks are skipped. In this case, the plugin will end up shipping Newtonsoft 12 without any warnings.

```xml
  <Dependencies>
    <PackageDependency Package="OpenTAP" Version="^9.22.1" />
  </Dependencies>
```

alnlarsen:
Closes #1899

### DIFF
--- a/Package/CreatePackage/Package.cs
+++ b/Package/CreatePackage/Package.cs
@@ -482,6 +482,36 @@ namespace OpenTap.Package
 
             public void Clear(PackageDef pkg) => packageAssemblies.Invalidate(pkg);
         }
+        
+        private static void VerifyDependenciesInstalled(this PackageDef pkg)
+        {
+            var currentInstallation = Installation.Current;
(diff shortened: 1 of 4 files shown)


rmadsen-ks (approved):
Looks like a nice cleanup and improvement!

PR #1909, opened 2025-03-03
1852 nullable property types

### LINKED ISSUE #1852: Open tap can't parse tap plans (xml) for test that have properties of type Nullable<T>
**Describe the bug**

Tests that have properties of type Nullable<T> or whose properties have nested Nullable<T> within them throw an error when loading the tapplan xml

 - Windows 11
Latest stable 9.27.0 and main branch

Open a test plan with a Nullable

Example         

Code:
public int? MyVariable{ set; get; }

TestPlan:
<MyVariable>2</MyVariable>

**Actual behavior**

Load the property correctly from the test plan and assign value 2

**Expected behavior**

Throws an error. TapEngine can't handle Nullables it seems.

11:02:06.944  Serializer   XML Line 15: Unable to read Nullable`1. Value cannot be null.
Parameter name: newobj

**Additional context**

I have added the following code to tap.Engine-> ObjectSerializer.cs

image

Opentap can now handle my Nullable properties even if nested across other objects. Test plan opened without issue and property was assigned correctly. Seems to work fine but it only work for known SDT types (such as int.

rmadsen-ks:
Close #1852

### DIFF
--- a/Engine.UnitTests/SerializerTests.cs
+++ b/Engine.UnitTests/SerializerTests.cs
@@ -448,5 +448,33 @@ namespace OpenTap.Engine.UnitTests
                 Assert.IsFalse(wasUsed);
             }
         }
+
+        public class ObjectWithNullable
+        {
+            public double? NullableDouble { get; set; }
+        }
+
+        [Test]
+        public void SerializeDeserializePropertyWithNullable()
+        {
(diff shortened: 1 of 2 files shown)


Oeh1995 (approved):
Looks good

PR #1911, opened 2025-03-07
run linux regression tests on ubuntu-latest

alnlarsen:
Ubuntu 20.04 jobs will soon start failing. Let's just use latest.


rmadsen-ks (approved):
Looks great!

PR #1914, opened 2025-03-10
1913-AbandonedMutexFix

### LINKED ISSUE #1913: AbandonedMutex can cause OpenTAP to not being able to start
Somehow the AddRecent log code can stop without releasing the mutex. this is a problem because its a system wide mutex. The fix is to ignore the exception and continue.

More info: https://learn.microsoft.com/en-us/dotnet/api/system.threading.abandonedmutexexception?view=net-9.0

rmadsen-ks:
- Fixed the issue by catching the AbandonedMutexException. When this exception is caught we still get the mutex and the opportunity to release it, but if its not released the same issue will occur for the next application trying to use it.

- Due the complexity of the bug, this cannot really be reproduced using a unit test. It should be reproducible by attaching a debugger and stopping the application without releasing the mutex.

Close #1913

### DIFF
--- a/Engine/SessionLogs.cs
+++ b/Engine/SessionLogs.cs
@@ -283,8 +283,15 @@ namespace OpenTap
             {
                 if (!recentFilesValid) 
                     return Array.Empty<string>();
-                
-                recentLock.WaitOne();
+
+                try
+                {
+                    recentLock.WaitOne();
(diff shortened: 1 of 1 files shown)


alnlarsen (approved):
Looks good to me

PR #1917, opened 2025-03-14
Change invariant string comparisons to ordinal

### LINKED ISSUE #1819: Cleanup: Prefer ordinal string comparisons over invariant string comparisons
In many places, we are currently using `StringComparer.InvariantCulture [IgnoreCase]` when doing string comparisons. From briefly looking through the code, I don't think we ever do this for a good reason. 

For most caching purposes, ordinal comparisons would be much faster because invariant comparisons do a lot more work, compared to ordinal comparisons which just compare the literal value of code points.

For sorting purposes, invariant sorting makes great sense when considering different locales. E.g. when displaying a list of test steps, it is natural that e.g. `ß` will be ordered close to `ss`. So any collection being sorted before being presented to the user should probably continue using `InvariantCulture`.

In other places, we are using `string.{ToLower,ToUpper}()` before doing a case insensitive string comparison. This is also much slower than just using `OrdinalIgnoreCase`, so let's improve this as part of the cleanup as well.

Sources:
* Performance: [String Equality and Performance in C#](https://rhale78.wordpress.com/2011/05/16/string-equality-and-performance-in-c/)
* Semantics: [Difference between InvariantCulture and Ordinal string comparison](https://stackoverflow.com/questions/492799/difference-between-invariantculture-and-ordinal-string-comparison)

alnlarsen:
It is (almost) always correct to use ordinal comparisons, but OpenTAP mostly uses invariant comparisons.

See linked issue for relevant resources

Closes #1819

### DIFF
--- a/Engine.UnitTests/TestTestSteps/PipingProcessStep.cs
+++ b/Engine.UnitTests/TestTestSteps/PipingProcessStep.cs
@@ -53,7 +53,7 @@ namespace OpenTap.Engine.UnitTests.TestTestSteps
                     var act = DescribeBytes(r.Answer);
                     log.Info($"Expected: {exp}");
                     log.Info($"  Actual: {act}");
-                    if (r.Answer.Equals(ExpectedAnswers[i], StringComparison.Ordinal) == false)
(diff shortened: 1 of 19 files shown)


rmadsen-ks (approved):
Looks great!

PR #1920, opened 2025-03-14
Implement equality members in example object

alnlarsen:
This fixes a warning when multi-editing this example


rmadsen-ks (changes requested):
Since this is example code, should we add a comment about why this is done?

rmadsen-ks (approved):
looks good

PR #1925, opened 2025-03-19
fix payload folder selection logic

alnlarsen:
This misplaced parenthesis caused us to compare the literal string 'x64.ToUpper()' to the $(PlatformTarget) variable, which obviously does not work well.


rmadsen-ks (approved):
Looks great

PR #1930, opened 2025-03-20
1929 Package Validation

rmadsen-ks:
Added support for checking if a file exists for package validation. This is mostly intended to be used by system-wide packages which can be uninstalled/upgraded through other paths than package install/uninstall.

Added unit tests to ensure that it works.

Tested that it works with system wide packages.

### DIFF
--- a/Package/PackageActions/List.cs
+++ b/Package/PackageActions/List.cs
@@ -197,10 +197,24 @@ namespace OpenTap.Package
                 var installedPackage = installed.FirstOrDefault(p => p.Name == plugin.Name);
                 var latestPackage = packages.Where(p => p.Name == plugin.Name).OrderByDescending(p => p.Version).FirstOrDefault();
 
-                var installedString = installedPackage == null ? "" : installedPackage.IsValid() ? " - installed" : " - needs reinstall";
-                
-                if (installedPackage != null && installedPackage.IsSystemWide())
-                    installedString += " system-wide";
+                bool isInstalled = installedPackage != null;
+                bool isSystemWide = installedPackage?.IsSystemWide() ?? false;
+                bool isValid = installedPackage?.IsValid() ?? false;
+                string installedString = "";
+                if (isInstalled)
+                {
+                    installedString = "-";
+                    if (isValid)
+                        installedString += " installed";
+                    if(isSystemWide)
+                        installedString += " system-wide";
+                    if (!isValid)
+                    {
+                        if (installedString != "-")
+                            installedString += " -";
+                        installedString += " needs reinstall";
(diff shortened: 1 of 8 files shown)


Package/PackageActions/List.cs
```diff
@@ -197,7 +197,7 @@ private void PrintReadable(List<PackageDef> packages, HashSet<PackageDef> instal
                 var installedPackage = installed.FirstOrDefault(p => p.Name == plugin.Name);
                 var latestPackage = packages.Where(p => p.Name == plugin.Name).OrderByDescending(p => p.Version).FirstOrDefault();
 
-                var installedString = installedPackage == null ? "" : " - installed";
+                var installedString = installedPackage == null ? "" : installedPackage.IsValid() ? " - installed" : " - needs reinstall";
                 
                 if (installedPackage != null && installedPackage.IsSystemWide())
                     installedString += " system-wide";
```
alnlarsen, line 203: I think this looks pretty bad:

```
PathWave License Manager         - 7.4.1+efeba870                                   - needs reinstall system-wide
```

Could we make it something like
```
PathWave License Manager         - 7.4.1+efeba870                                   - system-wide - needs reinstall
```

I'm not sure how I feel about the wording of "needs reinstall". I would prefer something like "not validated". What do you think?

Package/PackageFile.cs
```diff
@@ -539,6 +539,8 @@ public string Hash
         [DefaultValue("package")]
         public string Class { get; set; }
 
+        public List<Validation> Validation { get; set; }
+
```
alnlarsen, line 543: Add xml docs

Package/PackageSchema.xsd
```diff
@@ -130,4 +130,24 @@
...
+
+  <xs:element name="Validation">
+    <xs:complexType>
+    <xs:sequence>
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:element name="FileExists" type="FileExists"/>
+      </xs:choice>
+    </xs:sequence>
+    </xs:complexType>
+  </xs:element>
```
alnlarsen, line 152: Will this schema change make older versions of OpenTAP unable to parse the content of new OpenTAP packages using this feature?

alnlarsen (changes requested):
Overall this looks great, but there are a lot of XML comments missing. Can you take a look at those?

Package/PackageSchema.xsd
rmadsen-ks: This should not be a problem. At least old versions of KS8400 can list the packages just fine.

PR #1940, opened 2025-03-28
Fix resource annotation

alnlarsen:
This fixes an issue with available values for resources in some inheritance scenarios where the base type is an IResource, but not an IInstrument or an IDut.


rmadsen-ks (approved):
Looks great!

PR #1942, opened 2025-03-30
Multi language support

alnlarsen:
* Add a new settings object with support for language selections
* Add new cli action for creating translation files
* Add translation support in the Annotation layer


Oeh1995 (approved):
Very cool

Oeh1995 (dismissed):

Engine/Annotations/Annotation.cs
```diff
@@ -1316,17 +1317,22 @@ public IEnumerable Selected
             }
         }
 
-
         class InputStepAnnotation : IAvailableValuesSelectedAnnotation, IOwnedAnnotation, IStringReadOnlyValueAnnotation
         {
+            public class Strings : StringLocalizer<Strings>
+            {
+                public readonly string None = "None";
```
rmadsen-ks, line 1324: Can we do ```public string None => GetTranslation("None");```? I think that would be a bit less magical. 

Also, a translator can then be used to get all the existing translation by calling the property getters through reflection.

sdk/OpenTap.Sdk.New/TranslateAction.cs
```diff
@@ -0,0 +1,226 @@
...
+using OpenTap.Package;
+using OpenTap.Translation;
+
+namespace OpenTap.Sdk.New;
+
+/// <summary>
+/// This class contains helpful utilities for creating translations of packages.
+/// </summary>
+[Display("translate", Group: "sdk", Description: "Create a new translation template for a package.")]
+public class TranslateAction : ICliAction
```
rmadsen-ks, line 25: lets add an argument to specify language.

rmadsen-ks (changes requested):
I had a few remarks.

Engine/TapSerialization.cs
```diff
@@ -127,41 +128,42 @@ public object Deserialize(XDocument document, ITypeData type = null, bool autoFl
             var prevSer = currentSerializer.Value;
             currentSerializer.Value = this;
             ClearErrors();
-            using (ParameterManager.WithSanityCheckDelayed())
+            // We set the language to english during serialization / deserialization to ensure nothing culture-specific
+            // is written / loaded from files.
+            using var _ = withNeutralLanguageContext();
```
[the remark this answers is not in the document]
alnlarsen: I don't actually think it is relevant for OpenTAP, but my thinking was that it could be an issue for serializers defined in other plugins

sdk/OpenTap.Sdk.New/TranslateAction.cs
alnlarsen: This doesn't really make sense since we can only generate a template with neutral strings. I think creating files for other languages is in the domain of tooling such as [ResXManager](https://marketplace.visualstudio.com/items?itemName=TomEnglert.ResXManager) where the neutral language is used as the baseline for creating new translations.

Engine/Translation/ITranslationProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+internal class ResXTranslationProvider : ITranslationProvider
+{
+    public CultureInfo Culture { get; }
+    private readonly Dictionary<string, DateTime> UpdateTime = [];
+    public ResXTranslationProvider(CultureInfo culture, IEnumerable<string> files)
+    {
+        Culture = culture;
+        foreach (var file in files)
+        {
+            // Mark all files as
```
rmadsen-ks, line 32: as what?

Engine/Translation/ITranslationProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+    public ResXTranslationProvider(CultureInfo culture, IEnumerable<string> files)
+    {
+        Culture = culture;
+        foreach (var file in files)
+        {
+            // Mark all files as 
+            UpdateTime[file] = DateTime.MinValue;
+        }
+
+        TapThread.WithNewContext(() =>
```
rmadsen-ks, line 36: This class should probably be an IDisposable and clean up the thread when disposed.

Engine/Translation/ITranslationProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+    private ImmutableDictionary<string, string> _stringLookup = ImmutableDictionary<string, string>.Empty;
+
+    bool NeedsUpdate(string key)
+    {
+        if (!UpdateTime.TryGetValue(key, out var lastUpdate)) return false;
+        if (DateTime.Now - lastUpdate < TimeSpan.FromSeconds(1)) return false;
+        if (!File.Exists(key)) return false;
+        if (lastUpdate == DateTime.MinValue) return true;
+        var lastWrite = new FileInfo(key).LastWriteTime;
+        if (lastWrite > lastUpdate) return true;
```
rmadsen-ks, line 106: Actually, I would not say it is important the the last write is at a new date. If its just different from the last time we checked, that should prompt for a change.

lets encapsulate the cache validation in a function like:
```
string CacheValidationMarker(string file){
    return FileInfo(key).LastWriteTime.ToString();
}

```

We can then decide inside the function what triggers an update. It could be LastWriteTime, or LastWriteTime + Size, ...

Engine/UserInput.cs
```diff
@@ -132,6 +133,13 @@ public class SubmitAttribute : Attribute { }
     /// <summary> Standard implementation of UserInputInterface for Command Line interfaces</summary>
     public class CliUserInputInterface : IUserInputInterface
     {
+        class Strings : IStringLocalizer
+        {
+            public string PleaseEnterNumberOrName => this.Translate("Please enter a number or name ");
```
rmadsen-ks, line 138: Should this pattern be like:

```cs
class Strings : IStringLocalizer
        {
           static Strings strings = new Strings();
           public static string PleaseEnterNumberOrName => strings .Translate("Please enter a number or name ");
```

Then it can be used as this:
```Strings.PleaseEnterNumberOrName```

rmadsen-ks (changes requested):
Can you write a how to test guide in the PR?

Engine/UserInput.cs
alnlarsen: That's a great idea!

rmadsen-ks (changes requested):
Lets make sure to clean up threads and not read translations unless they are needed.

rmadsen-ks (approved):
Works perfectly!

PR #1944, opened 2025-04-01
Add hackish workaround to avoid upgrade errors

### LINKED ISSUE #1943: Fix issue preventing upgrades from older versions of OpenTAP
Older versions of OpenTAP will attempt to load plugins while self-upgrading. This causes issues in the new serializer type added in version 9.28.

By catching TypeLoadException and falling back to the old behavior, we avoid this issue in specific upgrade scenarios.

alnlarsen:
This is a workaround for a bug in older versions in OpenTAP

Closes #1943


Oeh1995 (approved):
Looks good

PR #1956, opened 2025-04-28
prepend package name to eula identifier

### LINKED ISSUE #1955: EULA: Avoid name clashes
It is currently pretty easy for eula identifiers to clash. We should prepend the package name to the eula identifier to guard against this.

alnlarsen:
This prevents a likely source of EULA name clashes.

A EULA is not worth much if OpenTAP considers it to be accepted due to an incidental name clash, and the user is given no opportunity to review it.

Closes #1955


Oeh1995 (approved):
Nice

PR #1968, opened 2025-05-02
Ensure SRQ delegate is not garbage collected

### LINKED ISSUE #1963: ScpiInstrument SRQ's crash KS8400 editor
_Seen today in **9.28**. I didn't test older versions._

A customer is trying to use SRQ's because they have code in LabVIEW that is using viWaitOnEvent to wait for an SRQ to be triggered (its getting triggered when a measurement completes).

I tried to implement an example with ScpiInstrument in KS8400. This works mostly, but I am running into an issue where the unmanaged eventhandler is triggering a crash in KS8400. To reproduce, use the following code to create an instrument / test step that should get an SRQ when the *OPC is sent. In general it will work, but will fail when:

1. Run the testplan and it succeeds, then the connection to the instrument is closed automatically
2. Click the "open connection" button
3. Click run and Editor.exe will crash
4. Sometimes you might need to run it multiple times

This fails both in Debug and Release

You can use my HiSLIP instrument at TCPIP0::156.140.94.238::hislip0::INSTR

Its possible I am doing something incorrect with the SRQ, but I don't see any other examples on how to use it. 

```
using OpenTap;
using System.Threading;

namespace WaitOnEventExample
{
    [Display("InstrumentWithSRQ", Group: "WaitOnEventExample", Description: "Insert a description here")]
    public class InstrumentWithSRQ : ScpiInstrument
    {
        private static AutoResetEvent _eventSignal = new AutoResetEvent(false);
        private static bool _readyForSrq = false;
        private static byte _srqStatusByte = 0;

        public InstrumentWithSRQ()
        {
            Name = "MyINST";
        }

        public override void Open()
        {
            SRQ += Instrument_SRQ;
            base.Open();
        }

        public override void Close()
        {
            base.Close();
            SRQ -= Instrument_SRQ;
        }
        public int MeasureAndWaitForSrq()
        {

            // tell the SRQ handler to do its job now
            _readyForSrq = true;
            // setup the SRQ on the instrument side
            SetupSrq();

            // do the thing that causes the SRQ to be fired
            this.ScpiCommand("*WAI;*OPC");

            // wait for the SRQ to be fired
            _eventSignal.WaitOne();

            int esr = int.Parse(this.ScpiQuery("*ESR?"));
            Log.Debug($"*ESR?: {esr}, Status Byte: {_srqStatusByte}");

            return esr;
        }

        private void SetupSrq()
        {
            // clear the status
            this.ScpiCommand("*CLS");

            // trigger an event when *OPC is sent
            this.ScpiCommand("*ESE 1");

            // enable the SRQ system
            this.ScpiCommand("*SRE 32");
        }

        private static void Instrument_SRQ(ScpiInstrument sender)
        {
            if (_readyForSrq)
            {
                // Reset the flag to prevent re-entrance
                _readyForSrq = false;
                // Read the SRQ status byte
                ((IScpiInstrument)sender).IO.ReadSTB(ref _srqStatusByte);
                // Disable the SRQ system so we don't get another signal
                sender.ScpiCommand("*SRE 0");

                _eventSignal.Set(); // Signal the waiting thread
            }
        }
    }

    [Display("WaitForEvent", Description: "Wait for averaging to complete", Group: "Wait On Event Example")]
    public class WaitOnEventTestStep : TestStep
    {
        [Display("Instrument")]
        public InstrumentWithSRQ Instrument { get; set; }

        public override void Run()
        {
            Log.Debug($"srq result = {Instrument.MeasureAndWaitForSrq()}");

            RunChildSteps(); //If step has child steps.
            UpgradeVerdict(Verdict.Pass);
        }
    }
}

```

alnlarsen:
This fixes an issue causing OpenTAP to crash under some circumstances when using SRQ

Closes #1963


Oeh1995 (approved):
Looks good

PR #1973, opened 2025-05-09
Package platform dependent runtime config

alnlarsen:
This fixes an issue causing a lot of type resolution errors to occur when installing packages which depend on WPF / PresentationCore APIs


rmadsen-ks (approved):
Looks great and seems to work!

PR #1980, opened 2025-05-12
Remove dependency on resx writer

alnlarsen:
This removes the dependency on ResXResourceWriter in favor of a XDocument


rmadsen-ks (approved):
Works well! Nice fix.

PR #1981, opened 2025-05-12
Add viwaitonevent to visa

### LINKED ISSUE #1964: OpenTAP does not support viWaitOnEvent and customers would be better off using that an SRQ event handler in most cases
Historical experience says that people really like viWaitOnEvent, rather than using an SRQ event handler, because they can use it in a synchronous manner, rather than adding in asynchronousness to their app/sequence. viWaitOnEvent is much simpler to implement, especially in a test sequence. 

viWaitOnEvent is NOT currently exposed through IVisa/ScpiIO/ScpiInstrument, so this would be a somewhat breaking change?

frederikja163:
Closes #1964

### DIFF
--- a/Engine/IVisa.cs
+++ b/Engine/IVisa.cs
@@ -47,6 +47,8 @@ namespace OpenTap
         int viInstallHandler(int vi, int eventType, viEventHandler handler, int UserHandle);
         /// <summary>Uninstall handler</summary>
         int viUninstallHandler(int vi, int eventType, viEventHandler handler, int userHandle);
+        /// <summary> Waits for an occurrence of the specified event for a given session. </summary>
+        int viWaitOnEvent(int vi, int eventtype, int timeout, out int outeventtype, out int outcontext);
         /// <summary>Read data from device</summary>
         unsafe int viRead(int vi, ArraySegment<Byte> buffer, int count, out int retCount);
         /// <summary>Write data to device</summary>
@@ -61,5 +63,6 @@ namespace OpenTap
         int viUnlock(int vi);
         /// <summary>Event handler prototype</summary>
         delegate int viEventHandler(int vi, int eventType, int context, int userHandle);
+
     }
 }
--- a/Engine/Visa.cs
+++ b/Engine/Visa.cs
@@ -43,6 +43,8 @@ namespace OpenTap
         private delegate int viUninstallHandlerDelegate(int vi, int eventType, IVisa.viEventHandler handler, int userHandle);
         private delegate int viInstallHandlerDelegate2(int vi, int eventType, IVisa.viEventHandler handler, int UserHandle);
         private delegate int viUninstallHandlerDelegate2(int vi, int eventType, IVisa.viEventHandler handler, int userHandle);
+
+        private delegate int viWaitOnEventDelegate(int vi, int eventType, int timeout, out int outEventType, out int outContext);
         private unsafe delegate int viReadDelegate(int vi, ArraySegment<byte> buffer, int count, out int retCount);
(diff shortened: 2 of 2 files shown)


alan-copeland-keysight:
I think I like #1 the best - 

```Require the user to call EnableEvent and DisableEvent on the scpiinstrument before firing the events to wait on.

var scpi = new ScpiInstrument();
scpi.EnableEvent(ScpiEvent.EVENT_SERVICE_REQ);
// ...
scpi.WaitOnEvent(ScpiEvent.EVENT_SERVICE_REQ, 1000, out _, out _);
```

However, the user might need receive the context variable, and then to explicitly call viClose on it. Its a ViSession and if it doesn't get closed, you will eventually run out of ID's (_behavior is dependant on VISA implementation I think)_. The lack of a way to close the context in the viEventHandler for SRQ's is also a similar bug (_should I submit an issue on this?_).

In many cases with simple testplans, you won't get an ViSession/context exhaustion issue. However if you are using viWaitOnEvent in a loop, e.g. waiting for a trace from a spectrum analyzer, then plotting it on a UI, you could run out of ViSession/context ID's. In some implementations it was limited to 256 ID's, in other it was upped to 1024, or the limit could be higher now.

frederikja163:
@alan-copeland-keysight Im not sure i fully understand what context you want to close that cannot be closed. Did you look at the code? It might also be possible you are talking about something only tangentially related, in which case it should maybe be a seperate pull request.

rmadsen-ks (changes requested):
WaitOnEvent outContext needs to be closed somehow. Or alternatively we just dont support it.

Would it work with:

  public ScpiIOResult WaitOnEvent(ScpiEvent eventType, int timeout, out ScpiEvent outEventType,
            out int? outContext); ?

alan-copeland-keysight:
If the context is not closed, you will leak a context handle, and eventually you will run out of those and things will crash. This could definitely happen if you use viWaitOnEvent in a loop (leak reading a waveform over and over, which is what the customer is trying).

The Keysight VISA help specifies the following in the **viWaitOnEvent** documentation:
```
If the value VI_NULL is specified in the outContext parameter of viWaitOnEvent and the return value is successful, viClose is automatically invoked on the event context rather than returning it to the application. Important: C/C++ programs do this, see Note on Context Return in visa32.cs and visa32.vb below.
	
Since system resources are used when waiting for events (viWaitOnEvent), the viClose function must be called to free up event contexts (outContext).


........ Note about visa32.cs / visa32.vb ........

Note on Context Return in visa32.cs and visa32.vb
 This section contains information specific to the Windows product.

It is not possible to pass a 'null' pointer for the 'ref context' parameter in viWaitOnEvent as it is defined in visa32.cs and visa32.vb. This means that VISA will not close the context and when viWaitOnEvent is called in a loop, the system will eventually run out of resources.

The solution is to close the returned context after a successful return from viWaitOnEvent. Here is some C# sample code that works:

          ...
          int context = 0;
          err = visa32.viWaitOnEvent(vi, visa32.VI_EVENT_SERVICE_REQ, 500, ref eventType, ref context);
          if (err >= visa32.VI_SUCCESS)
           {
                visa32.viClose(context);
          }
          ...

```

* Could you always just close the context automatically? yes, however....

* When is the context actually useful information? when you have multiple events configured on different resources. It is possible to get a viWaitOnEvent for a different resource than the one you expect, so the context could be useful to query the device to say "who are you?" so that you can disambiguate which event actually fired....

So.........

In an OpenTAP context, the user doesn't really have access to VISA calls to use the **context** to check the identity, so it makes sense that you could just close the context automatically for the user. I think that would meet the core usecase of this feature.

frederikja163:
Not sure why package diff fails, but i dont think there are actual breaking changes

PR #1984, opened 2025-05-14
Set assembly version info for Tap.Upgrader.exe

alnlarsen:
This removes a warning about multiple different versions of Tap.Upgrader being present when installing OpenTAP 9.29.0-rc.1


Oeh1995 (approved):
Great!

PR #1988, opened 2025-05-15
Narrow down translation type discovery

alnlarsen:
This removes a lot of unnecessary types from translations by pruning types during the recursive search.

Because we did not previously prune in the type discovery step, we would pick up a lot of unneeded types if they were referenced by private properties or internal types.


rmadsen-ks (approved):
Looks good

PR #1989, opened 2025-05-16
Add install script for getting KS8400 on Mac and Linux

alnlarsen:


alnlarsen:
We now have a beautiful gui installer!

PR #1990, opened 2025-05-18
Specify distinctName when querying all package names

### LINKED ISSUE #1999: HttpPackagRepository: GetPackageNames does not use distinctName
The query which just fetches package names should use distinctName since this significantly reduces the amount of data the repository sends, is much faster, and puts less load on the repository.

alnlarsen:
This should improve query performance

Closes #1999


rmadsen-ks (approved):
nice!

PR #2005, opened 2025-05-27
increase retry rate in tap upgrader

alnlarsen:
This reduces the likelihood that another process using tap.exe after an upgrade is started. The consequences of not updating tap.exe are much more serious after updating to .NET 9 than previously


rmadsen-ks (approved):
Looks good!

PR #2017, opened 2025-06-02
2008 test plan run pause api.

### LINKED ISSUE #2008: TestPlanRun.Pause
Add a pausing api to the test plan run.

This should work the same way as when clicking Pause in a UI.

rmadsen-ks:
close #2008

### DIFF
--- a/Engine/TestPlanExecuteArgs.cs
+++ b/Engine/TestPlanExecuteArgs.cs
@@ -3,9 +3,7 @@ using System.Collections.Generic;
 using System.Threading;
 namespace OpenTap;
 
-/// <summary>
-/// Arguments struct for TestPlan.Execute
-/// </summary>
+/// <summary> Arguments for TestPlan.Execute </summary>
 public sealed class TestPlanExecuteArgs
 {
     /// <summary> The result listeners used in the run. If not set, this will get the default value of ResultSettings.Current.</summary>
@@ -15,13 +13,11 @@ public sealed class TestPlanExecuteArgs
     public IEnumerable<ResultParameter> Parameters { get; set; }
 
     /// <summary> This is set when it is only wanted to run a subset of the steps in the test plan.</summary>
-    public ISet<ITestStep> StepsOverride { get; set; }
+    public ISet<ITestStep> SelectedTestSteps { get; set; }
 
-    /// <summary> Callback for when something wants to pause the test plan execution.</summary>
+    /// <summary> Callback invoked when <see cref="OpenTap.TestPlanRun.Pause" /> is called from within the test plan.</summary>
     public Action<TestPlanPauseEventArgs> OnPauseRequestedCallback { get; set; }
 
-    /// <summary>
-    /// This cancellation token allows canceling the test plan execution.
-    /// </summary>
+    /// <summary> This cancellation token allows cancelling  the test plan execution. </summary>
     public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
 }
--- a/Engine/TestPlanExecuteArgs.cs
+++ b/Engine/TestPlanExecuteArgs.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+namespace OpenTap;
+
+/// <summary>
+/// Arguments struct for TestPlan.Execute
+/// </summary>
+public sealed class TestPlanExecuteArgs
+{
+    /// <summary> The result listeners used in the run. If not set, this will get the default value of ResultSettings.Current.</summary>
+    public IEnumerable<IResultListener> ResultListeners { get; set; } = ResultSettings.Current;
+
+    /// <summary> Additional Parameters for the run.</summary>
+    public IEnumerable<ResultParameter> Parameters { get; set; }
+
+    /// <summary> This is set when it is only wanted to run a subset of the steps in the test plan.</summary>
+    public ISet<ITestStep> StepsOverride { get; set; }
+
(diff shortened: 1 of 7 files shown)


Engine/TestPlanExecuteArgs.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+namespace OpenTap;
+
+/// <summary>
+/// Arguments struct for TestPlan.Execute
```
alnlarsen, line 7: The comment says struct, but the declaration says class. Can we align this?

Engine/TestPlanExecuteArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+public sealed class TestPlanExecuteArgs
+{
+    /// <summary> The result listeners used in the run. If not set, this will get the default value of ResultSettings.Current.</summary>
+    public IEnumerable<IResultListener> ResultListeners { get; set; } = ResultSettings.Current;
+
+    /// <summary> Additional Parameters for the run.</summary>
+    public IEnumerable<ResultParameter> Parameters { get; set; }
+
+    /// <summary> This is set when it is only wanted to run a subset of the steps in the test plan.</summary>
+    public ISet<ITestStep> StepsOverride { get; set; }
```
alnlarsen, line 18: I know we are already using the parameter "StepsOverride", but I never liked it. Can we call it something even more descriptive such as "StepsToExecute" or "SelectedTestSteps"?

Engine/TestPlanExecution.cs
```diff
@@ -478,14 +478,30 @@ public Task<TestPlanRun> ExecuteAsync(CancellationToken abortToken)
...
+                StepsOverride = stepsOverride,
+                CancellationToken = cancellationToken
+            });
+        }
+
+        /// <summary>
+        /// Execute the TestPlan asynchronously.
+        /// </summary>
+        /// <param name="args">Execution arguments.</param>
+        /// <param name="cancellationToken">Cancellation token to abort the testplan</param>
```
alnlarsen, line 495:
```suggestion
        /// <param name="cancellationToken">Cancellation token to abort the testplan</param>
```

Delete this

Engine/TestPlanExecution.cs
```diff
@@ -507,20 +523,46 @@ public Task<TestPlanRun> ExecuteAsync(IEnumerable<IResultListener> resultListene
...
+                StepsOverride = stepsOverride
+            });
         }
 
-        TestPlanRun executeInContext(IEnumerable<IResultListener> resultListeners,
-            IEnumerable<ResultParameter> metaDataParameters, HashSet<ITestStep> stepsOverride)
+        /// <summary>
+        /// Execute the TestPlan as specified. Blocking.
+        /// </summary>
+        /// <param name="resultListeners">ResultListeners for result outputs.</param>
```
alnlarsen, line 537: Remove comments for removed parameters

Engine/TestPlanExecuteArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary> Callback for when something wants to pause the test plan execution.</summary>
```
alnlarsen, line 20:
```suggestion
    /// <summary> Callback invoked when <see cref="OpenTap.TestPlanRun.Pause" /> is called.</summary>
```

Engine/TestPlanExecuteArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+    public Action<TestPlanPauseEventArgs> OnPauseRequestedCallback { get; set; }
+
+    /// <summary>
+    /// This cancellation token allows canceling the test plan execution.
```
alnlarsen, line 24:
```suggestion
    /// This cancellation token allows cancelling the test plan execution.
```

Engine/TestPlanPauseEventArgs.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace OpenTap;
+
+/// <summary>
+/// Event args when a pause has been requested.
+/// </summary>
+public sealed class TestPlanPauseEventArgs
```
alnlarsen, line 6: I'm wondering if the test plan start args should be available in this as well, but I'm okay with not including it

Engine/TestPlanRun.cs
```diff
@@ -95,6 +95,14 @@ public string TestPlanName
         /// <summary> The thread that started the test plan. Use this to abort the plan thread. </summary>
         public TapThread MainThread { get; }
 
+        internal TestPlanExecuteArgs ExecuteArgs { get; set; }
+
+        /// <summary> Requests that the test plan run gets paused. If UI cannot pause the test plan this will be silently ignored. </summary>
```
alnlarsen, line 100:
```suggestion
        /// <summary> Request pausing the test plan by invoking the pause callback provided when the test plan was started. If the execution context does not support pausing, the request will be ignored. </summary>
```

Engine/TestPlanRun.cs
```diff
@@ -95,6 +95,14 @@ public string TestPlanName
...
+        public void Pause(TestStepRun stepRun = null)
```
alnlarsen, line 101: This effectively blocks the calling thread until the test plan is resumed, right? Should the doc comment mention this?

Also, I am wondering if this should return a bool indicating whether the test plan paused or not to avoid steps with weird behavior if they expect pausing to be available.

alnlarsen (changes requested):
This looks really promising! Please see remarks

rmadsen-ks:
Actually it turns out that this api was not needed. Closing this.

PR #2034, opened 2025-06-17
re-add opentap.dll reference in nuget package

alnlarsen:
This change only worked in some build types, but caused issues especially in multi-stage builds and transitive nuget dependencies


rmadsen-ks (approved):
Looks good

PR #2042, opened 2025-06-19
Create a migration guide to opentap 9.29 (.net9)

frederikja163:


alnlarsen:
We currently have this in our project templates:

```xml
	<PropertyGroup Condition="'$(OS)' == 'WINDOWS_NT' AND '$(Configuration)' == 'Debug'">
		<!--
      We target .NET Framework in debug builds when debugging with the WPF Editor due to a bug in Visual Studio's debugger.
      The debugger assumes that netstandard projects should be debugged as .NET Core apps, and thus launches a .NET Core debugger
      which fails to attach because tap.exe is a .NET Framework application.
      To ensure maximum compatibility, we recommend targetting netstandard2.0 in release builds, unless you need specific
      APIs that are not available in netstandard2.0.
    -->
		<TargetFramework>net472</TargetFramework>
	</PropertyGroup>
```

1. Remove this from our templates
2. Add a migration section about removing this from .csproj files if debugging doesn't work with Editor 9.29+

PR #2054, opened 2025-07-02
1820 Mixin Execution Order: Fix

### LINKED ISSUE #1820: Execution Order of Mixins
In some cases we might need to add multiple mixins to a test step. In such cases giving orders to added mixin executions might be useful. 

For example:
I have a mixin that is used to skip a test step called `SkipMixin`. Another mixin is used to assign a pre condition to a property (`PreconditionMixin`). 

I add `SkipMixin `to a test step and then I want to execute the pre condition I assigned with another mixin to the skipboolean property of the `SkipMixin`. However, skip mixin is executed first, thus the pre condition I try to execute does not affect the skip boolean of `SkipMixin`.

`SkipMixin`:
```csharp
[MixinBuilder(typeof(ITestStepParent))]
[Display("Skip Mixin", Description: "Skips the test step.")]
public class SkipMixin : ITestStepPreRunMixin
{
    [Display("Skip Step", Description: "Skip or not skip")]
    public bool SkipStep { get; set; }

    public void OnPreRun(TestStepPreRunEventArgs eventArgs)
    {
        eventArgs.SkipStep = SkipStep;
    }
}
```

`PreconditionMixin`'s pre run function:

```csharp
/// <summary>
/// Update the value of the member that calculated precondition assigned before test step is executed.
/// </summary>
/// <param name="eventArgs"></param>
public void OnPreRun(TestStepPreRunEventArgs eventArgs)
{
    PreConditionProvider.UpdateMemberValue(eventArgs.TestStep);
}
```

In this specific example, I need the `PreconditionMixin` to be executed before `SkipMixin`

rmadsen-ks:
- Added a way to control the execution order of TestStep PostRun and PreRun mixins.

- Added unit test to verify the fix.

Close #1820

### DIFF
--- a/Engine.UnitTests/MixinTests.cs
+++ b/Engine.UnitTests/MixinTests.cs
@@ -225,6 +225,82 @@ namespace OpenTap.UnitTests
                 Assert.AreEqual("123", aMemberValue);
             }
         }
+
+        public class MixinOrder1 : ITestStepPreRunMixin,ITestStepPostRunMixin
+        {
+            public string Value = "";
+
+            public void OnPreRun(TestStepPreRunEventArgs eventArgs)
+            {
+                (eventArgs.TestStep as TestStepWithMixins).preOrders.Add(Value);
+            }
+            public void OnPostRun(TestStepPostRunEventArgs eventArgs)
+            {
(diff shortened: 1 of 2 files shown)


alnlarsen (approved):
Looks good

PR #2057, opened 2025-07-03
2048 Allow Edit Test Plan While Paused.

### LINKED ISSUE #2048: Add API to allow editing the test plan while paused
Currently, plugins just checks if the test plan is locked, but actually it would be nice if this could be customized by a UI.

rmadsen-ks:
- Added API to allow editing the test plan while its paused.

- Added support for parameters and mixins.

Close #2048


Engine/ParametersMenuModel.cs
```diff
@@ -83,21 +83,22 @@ public void ParameterizeOnParent()
...
             {
-                var plan2 = source
-                    .Select(step => step is TestPlan plan ? plan : step.GetParent<TestPlan>()).FirstOrDefault();
-                return (plan2?.IsRunning ?? false) || (plan2?.Locked ?? false);
+                return source
+                    .Select(static step => step as TestPlan ?? step.GetParent<TestPlan>())
+                    .FirstOrDefault(static x => x != null)
+                    ?.AllowEdit ?? false;
             }
         }
```
alnlarsen, line 95: Maybe:
1. Undo the rename
2. Create a new property, `TestPlanAllowsEdit` with the current code
3. Update TestPlanLocked implementation:
```csharp
public bool TestPlanLocked => !TestPlanAllowsEdit;
```

alnlarsen (changes requested):
Looks good now, but there are a lot of unrelated things in the diff. Can you squash this?

rmadsen-ks:
This was superseeded by a squashed PR.

PR #2058, opened 2025-07-03
Use more recent smartinstaller

alnlarsen:
Use the most recent smartinstaller

Note that because the smart installer no longer runs rooted, I had to change the default diretory


rmadsen-ks (approved):
looks great!

PR #2060, opened 2025-07-03
2059 Allow jump to parent

### LINKED ISSUE #2059: Allow jumping to the parent step even if its the last step in the plan.
Currently, it works if its not the last step, because break offered is raised on every step. This causes the jump behavior to work because it can go to the next step and then jump back.

rmadsen-ks:
Close #2059


alnlarsen (dismissed):
Looks great!

alnlarsen (approved):
Works well!

PR #2063, opened 2025-07-07
2062 add dotnet to path

### LINKED ISSUE #2062: Package install does not work when dotnet is not in PATH
This happens because our logic for creating isolated child processes assumes dotnet is in PATH. This will usually be the case, but a notable exception is .desktop shortcuts on Linux which may or may not be aware of user PATH configurations, depending on how they are launched. 

We have a similar problem on Mac. Mac app shortcuts also don't look at PATH configurations

alnlarsen:
This is kind of a workaround, but I think it is the simplest solution which works reliably across systems.

To verify this solution:

1. Remove dotnet from PATH
2. Run `/path/to/dotnet tap.dll package install CSV`
3. Observe that it doesn't work

Closes #2062

### DIFF
--- a/Shared/ExecutorInterop.cs
+++ b/Shared/ExecutorInterop.cs
@@ -101,6 +101,15 @@ namespace OpenTap
                 RedirectStandardOutput = true,
                 RedirectStandardError = true
             };
+            var dotnet = ExecutorClient.Dotnet;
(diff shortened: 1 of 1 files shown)


rmadsen-ks (approved):
Looks great!

PR #2080, opened 2025-07-22
Translation: don't log errors for abstract types

### LINKED ISSUE #2079: Translation: don't log errors for abstract IStringLocalizers
Currently, running `tap sdk translate OpenTAP` gives this warning:

String localizer 'OpenTap.Translation.IStringLocalizer' does not have an empty constructor, and will not be translated.

We should not log htis warning for abstract types

alnlarsen:
Closes #2079


rmadsen-ks (approved):
Looks good!

PR #2082, opened 2025-07-23
Improve templates slightly less

alnlarsen:
* Make .vscode launch configs cross-platform (possible now because of .NET9)
* Set default editor to Editor (thanks to community edition)
* Remove Visual Studio workaround from .csproj file (not needed because of .NET 9)
* Add Rider launch config for Linux/MacOS


rmadsen-ks (approved):
Looks great!

PR #2087, opened 2025-07-25
Include translations for all display attributes

### LINKED ISSUE #2086: Translation: Always add all DisplayAttributes
Our current translation generator scans all types that could possibly be shown in the UI by the current plugin, but it does not capture objects which could be shown in the UI by other plugins, such as an Enum. This is to avoid bloating the translation file with irrelevant entries since this would include every public type in the plugin.

But if a type or member is using a Display Attribute, we should take that as a hint that the type is intended for display.

Let's include all types and members with display attributes in addition to what we currently include

alnlarsen:
* Include translations for all display attributes, on the assumption they were added because they can be displayed
* Relax SkipType logic to properly translate e.g. derived types with no new properties (SequenceStep was previously not translated for example)

With this change, we now include a bunch of 
Closes #2086


rmadsen-ks (approved):
Looks great!

PR #2088, opened 2025-07-25
2083 Multiple IEnumerable<> interface implementations bug: Fix

### LINKED ISSUE #2083: TypeData exceptions when type implements IEnumerable multiple times
Consider this type:
```csharp
public class MyCustomType : List<int>, IEnumerable<int>, IEnumerable<double>
{
    IEnumerator<double> IEnumerable<double>.GetEnumerator()
    {
        return (this as List<int>).Select(x => Convert.ToDouble(x)).GetEnumerator();
    }
}
```

Attempting to get the typedata of this with TypeData.FromType(typeof(MyCustomType)) results in an `AmbiguousMatchException`:

```
System.Reflection.AmbiguousMatchException: Ambiguous match found for ' System.Collections.Generic.IEnumerable`1[System.Int32]'.
   at System.RuntimeType.GetInterface(String fullname, Boolean ignoreCase)
   at System.Type.GetInterface(String name)
   at OpenTap.ReflectionHelper.GetEnumerableElementType(Type enumType)
   at OpenTap.TypeData.PostLoad()
   at OpenTap.TypeData..ctor(Type type, PluginSearcher searcher)
   at OpenTap.TypeData.FromType(Type type)
   at TestExample.Run() in C:\git\ks8400\gui\TestTestSteps\DemoFunctions.cs:line 43
```

Thanks @brandon-tan-keysight for the example

rmadsen-ks:
- Fixed  the issue by catching the ambiguous match exception.

- Added unit test to verify the fix.
Close #2083

### DIFF
--- a/Engine.UnitTests/TypeDataTest.cs
+++ b/Engine.UnitTests/TypeDataTest.cs
@@ -1,5 +1,6 @@
 ﻿using NUnit.Framework;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


alnlarsen (changes requested):
This leads to null reference exceptions when trying to serialize such types:

image

Can we get this to fail with a friendlier error?

PR #2093, opened 2025-07-29
serializer: defer property setter

### LINKED ISSUE #2084: Unable to merge Input<T> settings
If I create a sequence step and add two child steps using `Input<T>` and merge them on the sequence step, the settings are not correctly loaded when saving and reloading the test plan,
even with a simple T such as `Input<int>`.

alnlarsen:
This fixes an issue where parameter members are cloned before they are fully deserialized if the parameterized member uses deferred setters

Also adds a unittest verifying the behavior

Closes #2084


rmadsen-ks (approved):
Great!

PR #2094, opened 2025-07-30
tap.sh: don't suggest adding user to root group

### LINKED ISSUE #1971: Unfortunate error message
If OpenTAP is installed in a directory, where the executing user does not have write permissions, the `tap` command exits with exit code 0 along with an error message. If, for instance, the OpenTAP installation directory group owner is `root`, then the error message becomes:

`This installation belongs to the group 'root'. The user can be added to this group with the command 'usermod -a -G root <user>'.`

This is a scary command to run, especially for novice users. I suggest two changes:
- We change the wording a bit so we do not recommend regular users to become member of the `root` group.
- We use another exit code than 0.

alnlarsen:
This message was intended to help the user add themselves to the 'opentap' group, which is created by some version of SmartInstaller.

But if the installation is created in some other way, this suggestion is potentially dangerous

Closes #1971

### DIFF
--- a/LinuxInstall/tap.sh
+++ b/LinuxInstall/tap.sh
@@ -52,6 +52,9 @@ else
   # If the user cannot write to the installation, OpenTAP will not work correctly.
   # Instead, we should give a hint about how to resolve the issue.
   TapDllGroupOwner="$(stat -c "%G" "$TapDllDir")"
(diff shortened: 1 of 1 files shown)


rmadsen-ks (changes requested):
Maybe we should just remove this. The new installer makes a user-local installation.

PR #2095, opened 2025-07-30
1945 merge complex types

### LINKED ISSUE #1945: Serializer error when merging list member from embedded class
I didn't investigate this too thoroughly yet, but I hear from reliable sources that something like this causes issues during serialization

```csharp
public class MyStep : TestStep
{
    public List<ComplexType> ComplexList { get; set; } = [];
    [EmbedProperties] public EmbeddedClass Emb { get; set; } = new();

    public override void Run()
    {
    }
}

public class EmbeddedClass : ValidatingObject
{
    public List<ComplexType> ComplexList { get; set; } = [];
}

public class ComplexType
{
    public int X { get; set; }
    public string Y { get; set; }
}

```

alnlarsen:
This allows merging lists of complex types if the element type can be cloned.

The current parameterization system already allows and supports this, but throws errors during serialization.

This also adds a unittest verifying everything is still correctly connected after serialization, and ensuring that updates to individual properties of elements in lists are propagated correctly.

Closes #1945

### DIFF
--- a/Engine.UnitTests/AnnotationTest.cs
+++ b/Engine.UnitTests/AnnotationTest.cs
@@ -1889,23 +1889,6 @@ namespace OpenTap.UnitTests
             Assert.AreNotEqual(obj1.Items, obj2.Items);
         }
 
-        [Test]
-        public void AddExternalParameterWarning()
-        {
-            var obj1 = new TestStepWithLists();
-            var obj2 = new TestStepWithLists();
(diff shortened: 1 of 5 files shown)


rmadsen-ks:
Can we figure out if this issue is at all related to embeded properties? If not, it should not be included the test.

rmadsen-ks (approved):
great!

PR #2119, opened 2025-09-01
Package install rework 2

### LINKED ISSUE #2111: Package install rework
I recently discovered that while open files cannot be deleted on Windows, they can be renamed and moved.

A lot of the complexity in our package installer logic comes from starting an isolated process and copying relevant plugins to the new location, then communicating file commands over an IPC socket. By using the rename strategy, this would no longer be required.

In addition, our startup sequence is jumping through a lot of hoops to avoid loading dlls for the same reason.

And lastly, we have a TapUpgrader.exe which exists so we can update tap.dll and tap.exe, which would also no longer be required.

This is a big refactor, but it would also remove a lot of complexity in different parts of the code.

Historically, it would be necessary to shut down any OpenTAP processes before initiating a plugin update, but this will no longer be needed. This of course breaks with user expectation, where e.g. an Editor will be loaded with older plugin versions than the user expects. We can leverage the existing logic which notifies the Editor of new versions to notify the user that the Editor should be restarted to use the new plugin versions.

In the long term, I think it would be possible to load the updated plugins, but that would require using AssemblyLoadContext to manage different versions of the same plugins.

alnlarsen:
* Completely remove the Isolation feature
* Migrate install / uninstall actions to move files out of the way before an upgrade instead of deleting them
* Update Tap.Upgrader to use the same strategy for upgrading OpenTAP in-place
* Update Tap.Upgrader to also update tap.dll
* Greatly simplify the startup sequence to just call the CliActionExecutor with no tricks
* Include hash for tap.exe
* Delete RestartManager

Closes #2111

### DIFF
--- a/Package/PackageActions/IsolatedPackageAction.cs
+++ b/Package/PackageActions/IsolatedPackageAction.cs
@@ -30,7 +30,7 @@ namespace OpenTap.Package
         /// </summary>
         [Browsable(false)]
         [CommandLineArgument("no-isolation", Description = "Avoid starting an isolated process.")]
-        [Obsolete("Package actions are no longer starting insolated.")]
+        [Obsolete("Package actions are no longer starting isolated.")]
         public bool NoIsolation { get; set; }
         
 


rmadsen-ks (commented):
In general this works shockingly well. 

One thing I noticed was that weird stuff happens when a plugin is deleted, even though the references to that DLL are still alive.

Installed Demonstration 9.0.7:

Image

Removed Demonstration 9.0.7:

Image

Installed Demonstration 9.0.6:

Some things works, but not everything.

Downgrading from Demonstration 9.0.7 to 9.0.6 seems to work flawlessly, although I think its still technically 9.0.7 that is installed.

It makes me wonder, what is the problem with removing - shouldn't it just keep the assembly reference alive in the background?

alnlarsen:
I have created a separate issue for this problem: https://github.com/opentap/opentap/issues/2144

rmadsen-ks (approved):
Reviewed the code and it works well. 

We need to fix #2144 before releasing 9.31, but that can come separately.

PR #2123, opened 2025-09-04
translations: provide neutral display attribute

### LINKED ISSUE #2122: Add an a way for DisplayAttribute to get the neutral language variant.

alnlarsen:
This adds a way to get the neutral variant of a display attribute based on a translated display attribute. This is needed for some plugins which make decisions based on string values in display attributes.

Closes #2122


rmadsen-ks (changes requested):
lets make the new property internal set.

PR #2125, opened 2025-09-08
2105 Custom Step Names: Fix

### LINKED ISSUE #2105: Custom step name formatting
Currently all step formatting has to be done with the format macro. But some steps could do something more user friendly.

Add an IFormatStepName interface:

```C#
public interface IFormatStepName : ITestStep{
    string GetFormattedName(); // this returns the formatted name, but may include macros {} in the output.
}

rmadsen-ks:
Added an interface that enables a test step to do custom name formatting.

Added unit tests to verify the feature.

Close #2105


alnlarsen (approved):
Looks great!

PR #2163, opened 2025-10-17
2162 scan more attributes

### LINKED ISSUE #2162: PluginSearcher: Scan more attributes
It is easy to accidentally load the assembly backing a TypeData by calling e.g. `td.GetAttribute<FooAttribute>()`. 

Today, PluginSearcher detects common attributes such as `BrowsableAttribute` and `DisplayAttribute` so they can be read without loading the assembly, because they are commonly used by UIs.

Other common attributes are `HelpLinkAttribute` and `SupportedOSPlatformAttribute`. Let's add these two attributes to the scanner. 

We can consider if we should just scan all attributes, though that might impact startup time.

alnlarsen:
* Fix a couple of build warnings
* Scan HelpLinkAttribute during search
* Scan Supported supported OS during search
* TypeData: Return scanned helplink
* PluginSearcher: Avoid returning unsupported plugins
* Add unittests verifying the typedata is not loaded prematurely

Closes #2162

### DIFF
--- a/.gitversion
+++ b/.gitversion
@@ -4,7 +4,7 @@
 # This is the version number that will be used. Prerelease numbers are calculated by 
 # counting git commits since the last change in this value.
 
-version = 9.31.0
+version = 9.31.1
 
 # A version is determined to be a "beta" prerelease if it originates from the default branch
 # The default branch is the first branch that matches the following regular expession.


sdk/OpenTap.Sdk.MSBuild/OpenTap.Sdk.MSBuild.csproj
```diff
@@ -6,6 +6,14 @@
...
     <PlatformTarget>AnyCPU</PlatformTarget>
+    <!-- We have had this warning for many years and never seen any issues, and I'm tired of seeing it
+        in the build output. It looks something like this:
+
+        warning NU1701: Package 'Microsoft.IO.Redist 6.0.0' was restored
+        using '.NETFramework,Version.....,
+        instead of the project target framework '.NETStandard,Version=v2.0'.
+        This package may not be fully compatible with your project. -->
+    <NoWarn>$(NoWarn);NU1701</NoWarn>
   </PropertyGroup>
```
rmadsen-ks:
```suggestion
    <!-Quiet a low-grade warning in the style of:
        warning NU1701: Package 'Microsoft.IO.Redist 6.0.0' was restored
        using '.NETFramework,Version.....,
        instead of the project target framework '.NETStandard,Version=v2.0'.
        This package may not be fully compatible with your project. 
        
        This is not seen to be an important issue in our case. 
        -->
    <NoWarn>$(NoWarn);NU1701</NoWarn>
</PropertyGroup>
```

PR #2167, opened 2025-10-21
updatecheck: prevent check from bricking cli

### LINKED ISSUE #2166: CLI: UpdateCheck bricks CLI when not connected to the internet
Because we wrap the update check in `CliUserInputInterface.AcquireUserInputLock()`, if the update check does not complete, it will indefinitely block all output.

This is the case when using Airplane Mode on a Windows laptop. Instead of timing out, the action hangs indefinitely.

I propose making the following changes:

1. Only acquire the lock when we are ready to write, and not while querying repositories
2. Make repository queries time out during update

alnlarsen:
This prevents the update check from indefinitely bricking the CLI when a repository is not reachable.

Closes #2166

### DIFF
--- a/Package/UpdateCheck.cs
+++ b/Package/UpdateCheck.cs
@@ -64,6 +64,7 @@ namespace OpenTap.Package
                 return;
 
 
+            using var _ = CliUserInputInterface.AcquireUserInputLock();
             if (updates.Any())
             {
                 Log.Info("Updates available for:");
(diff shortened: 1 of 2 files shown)


Package/UpdateCheck.cs
```diff
@@ -44,11 +44,14 @@ private void CheckForUpdatesAsync(CancellationToken cancellationToken)
 
             var installation = new Installation(ExecutorClient.ExeDir);
             IPackageIdentifier[] installedPackages = installation.GetPackages().ToArray();
-            Parallel.ForEach(PackageManagerSettings.Current.GetEnabledRepositories(null), repo =>
+
+            var linkedToken = CancellationTokenSource.CreateLinkedTokenSource([cancellationToken]);
+            linkedToken.CancelAfter(TimeSpan.FromSeconds(10));
```
rmadsen-ks: lets remove this timeout.

alnlarsen: Agreed

PR #2181, opened 2025-11-11
2180 fix teestream

### LINKED ISSUE #2180: MemoryStream artifacts are not saved correctly
The TeeStream implementation appears to not propagate its data correctly to all clients, dependning on the number of listeners and size of the stream.

alnlarsen:
This fixes a bug causing TeeStream to incorrectly propagate some artifacts because the buffer offset in the Read() implementation was not respected.

Closes #2180


rmadsen-ks (approved):
Great fix!

PR #2197, opened 2025-11-27
Catch exceptions thrown when instanciating annotators

### LINKED ISSUE #2196: Failed license checks can break the AnnotationResolver
If a plugin performs a license check in it's constructor and the check fails, then an error will be thrown by the AnnotationResolver causing dialogs and components to break in KS8400. 

This can be reproduced by installing a plugin that requires a license such as PTEM Battery DUT

Oeh1995:
If an exception is thrown in the constructor of a plugin then that should not crash the entire annotation resolving. I've added handling for exceptions thrown and then we just skip that particular instance and move on. 

Closes #2196

### DIFF
--- a/Engine/Annotations/Annotation.cs
+++ b/Engine/Annotations/Annotation.cs
@@ -3599,6 +3599,8 @@ namespace OpenTap
     /// </summary>
     internal class AnnotationResolver
     {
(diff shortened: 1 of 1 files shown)


alnlarsen (changes requested):
Minor nitpick

alnlarsen (approved):
Looks good

PR #2204, opened 2025-12-01
fix: CustomPackageData not loaded during install

### LINKED ISSUE #2202: New plugin types not loaded during package install
After the recent PluginSearcher race-condition fix https://github.com/opentap/opentap/pull/2171, SearchAsync() no longer completes before installation continues. Plugins that add ICustomPackageData types aren’t loaded in time, causing XML deserialization to produce MissingPackageData.

alnlarsen:
This fixes an issue causing new ICustomPackageData to not be loaded during package installation. After the refactoring in #2171, PluginManager.GetPlugins() was no longer invalidated after searching.

Closes #2202

### DIFF
--- a/Engine/PluginManager.cs
+++ b/Engine/PluginManager.cs
@@ -79,13 +79,11 @@ namespace OpenTap
         {
             public static ReadOnlyCollection<Type> GetPlugins(Type pluginBaseType)
             {
-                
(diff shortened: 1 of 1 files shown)


Oeh1995 (approved):
This works great!

PR #2205, opened 2025-12-02
Removed check that caused things not to work token domain had a port.

### LINKED ISSUE #2198: Allow package resolution on arbitrary port
**Describe the bug**

If a repository is hosted on a special port, then OpenTAP throws an exception with this message: `Domain should only be the host part of a URI and not a full absolute URI.`

 - OS: Any
 - OpenTAP Version 9.31.1-beta.9+65918661

**To Reproduce**

1. Host a repository on any port that is not 80 or 443
2. Run `tap package list -r "http://localhost:<port>;token=<token>"`

**Expected behavior**

If the token is valid and grants the correct permissions, `tap` should list the available package(s) in that repository.

rmadsen-ks:
We still probably want to support that domains can differ by port. So the same host can contain multiple token domains.

Another check could probably be made, but I am not sure how to robustly check for that.

Close #2198

### DIFF
--- a/Engine/Authentication/AuthenticationSettings.cs
+++ b/Engine/Authentication/AuthenticationSettings.cs
@@ -111,8 +111,6 @@ namespace OpenTap.Authentication
         /// <returns>A preconfigued HttpClient object</returns>
         public HttpClient GetClient(string domain = null, bool withRetryPolicy = false, string baseAddress = null)
         {
-            if (Uri.IsWellFormedUriString(domain, UriKind.Absolute))
-                throw new ArgumentException("Domain should only be the host part of a URI and not a full absolute URI.", nameof(domain));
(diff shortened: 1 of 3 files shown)


AsgerIversen (dismissed):
I guess the intent of this check it to guide users to use only the host part of the URI to pick a token. If GetClient can now also be called with `https://demo.automation.keysight.com/api/repository` as the `domain` argument, or a token can be created to only work for that URI. This intent was to only distinguish tokens on the domain/host and not on the entire URI. It might work to do that though, I'm not entirely sure. 

Anyway, the issue here is just that the IsWellFormedUriString misunderstands the colon that preseeds the port with the colon that comes after http in "http://", I think. Perhaps we could also make a smaller change to more manually check that the domain arg/property is only the host as the exception message expains says?

alnlarsen (approved):
Great!

PR #2212, opened 2025-12-09
authentication: support using self signed certificates

### LINKED ISSUE #2211: authentication settings: add API for allowing self-signed certificates
This is needed for some local https deployments where people would like to use authenticated package repositories with self signed certificates

alnlarsen:
Closes #2211


rmadsen-ks:
See issue comment

PR #2225, opened 2026-01-19
1979 Process Step NewLine Fix

### LINKED ISSUE #1979: ProcessStep adds an extra newline to output
ProcessStep uses the AppendLine method in the output data receive handler. This adds an extra unexpected newline sequence to the output which can be used as an input to another step. Would it be better to use the Append method instead?

You can simulate this by assigning the output from the process step to the _Log Message_ field of the log output step. The field wil expand with one extra row (the process should produce some output).

rmadsen-ks:
Fixed the issue by stream processing instead of using a StringBuilder.

Close #1979

### DIFF
--- a/BasicSteps/ProcessStep.cs
+++ b/BasicSteps/ProcessStep.cs
@@ -260,18 +260,19 @@ namespace OpenTap.Plugins.BasicSteps
                     
                     var mixedOutStream = new MemoryStream();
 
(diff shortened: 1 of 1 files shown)


rmadsen-ks:
Unfortunately this did not seem to have the wanted effect. It seems like the way Process works it that it calls ReadLine, so all the data will always have its natural newline removed. We could remove the final new line if any @alnlarsen, what do you think?

PR #2229, opened 2026-01-21
2216 parameterized embedded

### LINKED ISSUE #2216: Parameterized embedded propeties does not forward validation errors.
Validation error on a embedded property that has been forwarded to eg the test plan will not be shown at the parent step level.

This both applies to when it is parameterized to the test plan and a parent step.

rmadsen-ks:
Close #2216


alnlarsen (dismissed):
Works well!

PR #2265, opened 2026-02-09
2264 --log set between CliAction ctor and Execute

### LINKED ISSUE #2264: --log setting is set between CliAction construction and .Execute
**Describe the bug**

The log path set from --log is set between the ctor of CliAction and when its .Execute function is called. This means if  a CliAction wants to create external loggers, it has to either wait untill .Execute or manually parse the option to use the same log path.

 - OpenTAP Version 9.22.1 and 9.31.0

**Actual behavior**

Upon startup the session log is named/renamed 2-3 times

1) On setup in PluginManager.Load
2) CliActionExecutor from EngineSettings
3) ICliActionExtensions from the CLI

**Expected behavior**
We should set the SessionLog as few times as possible, especially since the rename logic is quite expensive with copying the entire file contents over.

frederikja163:
Closes #2264

Previously the log path was set 3 times.
First by PluginManager.Load to "tap <timestamp>"
Then by CliActionExecutor from EngineSettings (default "SessionLogs <timestamp>")
Finally by CliActionExtensions from the CLI

It was set in CliActionExtensions because that is where we parse Cli Arguments. However this gave issues if CliActions wanted to use this log in their constructor. Seperately if the application crashes before setting it the log would not be placed where the user would expect it to.

The new system sets it only twice.

First in PluginManager.Load from either arguments or to default "SessionLogs <timestamp>"
Second in TapEntry to the value in EngineSettings, unless the argument --logs was provided.

** Testing **
I ran it locally with and without the argument, with a breakpoint in SessionLogs.rename to see how many times it was called and from which location.

Everything seems to be working as expected.

### DIFF
--- a/Cli/TapEntry.cs
+++ b/Cli/TapEntry.cs
@@ -55,6 +55,7 @@ internal class TapEntry
         AppDomain.CurrentDomain.ProcessExit += (s, e) => cliTraceListener.Flush();
 
         PluginManager.Search();
+        SessionLogs.SetLogPathFromEngineSettings();
         DebuggerAttacher.TryAttach();
         CliActionExecutor.Execute();
     }
--- a/Engine/Cli/CliActionExecutor.cs
+++ b/Engine/Cli/CliActionExecutor.cs
@@ -190,27 +190,6 @@ namespace OpenTap.Cli
                 return (int)ExitCodes.UnknownCliAction;
             }
 
-            try
-            {
-                // setup logging to be relative to the executing assembly.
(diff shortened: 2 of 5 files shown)


alnlarsen (approved):
This seems to work as intended. One small thing though, Latest.txt is placed next to the indicated log file, and not in session logs. I don't necessarily think this is wrong, but it's also not exactly what I expected. Regardless, I think this is good enough to merge. It works and the new description is much clearer.

PR #2274, opened 2026-02-17
Use the official docker action

### LINKED ISSUE #1967: Update Docker Build method
We are currently using kaniko@master to build out docker images. There are a few issues with this

1. kaniko no longer has a maintainer
2. targetting the master branch is a risk
3. There are official docker endorsed actions for building docker images

I think we should prefer the more official, and more supported, docker actions over kaniko.

Oeh1995:
We should use the offical docker build action in our pipeline instead of the one from kaniko as it no longer is maintained. 

Closes #1967


alnlarsen (commented):
Does this work? Can we test it before merging?

alnlarsen (approved):
Looks great!

PR #2286, opened 2026-02-20
build with langversion 13

### LINKED ISSUE #2285: Unable to start Editor 9.28.2 with OpenTAP 9.32
This appears to be a regression introduced by building with the latest langversion.

Building with langversion 13 appears to fix this. Let's build with an older langversion until we can figure out what changed.

alnlarsen:
langversion 14 introduces some implicit behavior changes which has introduced some bugs. Specifically, it prevents Editor version 9.28 from starting.

Closes #2285


rmadsen-ks (dismissed):
Great!

PR #2313, opened 2026-03-23
libgit: add musl support

alnlarsen:
* add musl libgit binaries to Linux packages
* enhance arm64 detection on Linux
* add musl path to libgit loader

Closes #2312


rmadsen-ks (approved):
Lets add some testing on alpine.

PR #2315, opened 2026-03-24
make OpenTapPackageReference non-private

alnlarsen:
this prevents the C# compiler from copying them to the build root, which has a couple of negative consequences:

* causes opentap to emit warnings about duplicate assemblies
* general pluginsearcher slowdown
* breaks the Installation.Current.FindPackageContaining{Type,File} methods

Closes #2314


rmadsen-ks (approved):
Great!

PR #2321, opened 2026-04-01
2307 TestStepAbortIssue: fix

rmadsen-ks:
- Added catching exceptions from GetPropertiesFromObject and setting Error or exception on the test step run.
- Added unit test to verify the issue.

Close #2307


alnlarsen (approved):
looks good

PR #2322, opened 2026-04-01
pinned and updated all workflow actions.

rmadsen-ks:


alnlarsen (commented):
This is a good idea, but aren't we equally compromised if for example actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f is using an action with a floating tag? (or some other http resource)

I don't think we can do anything about it, just pointing out that it's not a complete solution.

alnlarsen (approved):
Looks good

PR #2323, opened 2026-04-13
2318 Members Lookup Memory Leak: Fix

### LINKED ISSUE #2318: Memory leak: `TestStepExtensions.membersLookup` permanently roots steps with DynamicMembers after Execute
## Summary

Any `ITestStep` that has a `DynamicMember` added via `DynamicMember.AddDynamicMember()` is permanently rooted in a static cache after `TestPlan.Execute()`. The objects can never be garbage collected, causing unbounded memory growth when plans are repeatedly loaded and executed.

This affects **all** consumers of `DynamicMember` — mixins, parameterized members, or any plugin that adds dynamic properties to steps.

## Root Cause

`TestStepExtensions` contains a static cache:

```csharp
private static readonly Cache<(TypeData target, ITypeData source), (IMemberData, bool hasEnabledAttribute)[]> membersLookup
    = new Cache<(TypeData, ITypeData), (IMemberData, bool)[]>(() => PluginManager.CacheState);
```

During `TestPlan.Execute()`, `GetSettingsLookup(TypeData targetType, ITypeData sourceType)` is called for each step to resolve resources, DUTs, and result sinks. For steps with `DynamicMember`s, `TypeData.GetTypeData(step)` returns a `DynamicTestStepTypeData` from `DynamicMemberTypeDataProvider`.

`DynamicTestStepTypeData` is **per-instance** — it holds a strong reference to the step via a `private readonly object target` field, and its `GetHashCode()`/`Equals()` are based on that instance reference:

```csharp
// DynamicTestStepTypeData (simplified from decompilation)
private readonly object target;

public override int GetHashCode()
{
    return (BaseType.GetHashCode() * 397) ^ target.GetHashCode();  // per-instance!
}

public override bool Equals(object obj)
{
    // ... compares target by reference
}
```

This means every step with a `DynamicMember` creates a **unique key** in the `ImmutableDictionary` inside `membersLookup`. The `Cache<K,V>` class only clears its dictionary when `PluginManager.CacheState` fires its `Updated` event (which only happens during plugin search — never during normal execution):

```csharp
private void CheckClear()
{
    if (!object.Equals(clearTrigger(), currentObj))  // never true during normal execution
    {
        dict = ImmutableDictionary<K, V>.Empty;
        currentObj = clearTrigger();
    }
}
```

**Result**: Each executed step with `DynamicMember`s gets a permanent entry. The `DynamicTestStepTypeData` key holds a strong reference to the step, which in turn holds the `TestPlan`, `TestPlanRun`, all child steps, result listeners, and everything connected to the execution graph.

## GC Root Chain (from `dotnet-dump analyze` → `gcroot`)

```
HandleTable (strong handle)
  → System.Object[]
    → OpenTap.Cache<(TypeData, ITypeData), (IMemberData, Boolean)[]>
      → ImmutableDictionary<(TypeData, ITypeData), (IMemberData, Boolean)[]>
        → ImmutableDictionary+HashBucket[] (tree nodes)
          → DynamicTestStepTypeData          ← per-instance, holds target
            → SimpleStep                     ← the step that should have been GC'd
              → TestPlan
                → TestPlanRunSummaryListener
```

## Reproduction

Minimal reproduction — create and execute plans in a loop, each with steps that have a `DynamicMember`:

```csharp
using System;
using System.Runtime.CompilerServices;
using OpenTap;

// Any trivial DynamicMember will trigger the leak
sealed class BareDynamicMember : DynamicMember
{
    static readonly ConditionalWeakTable<object, StrongBox<double>> _values = new();

    public BareDynamicMember()
    {
        Name = "BareValue";
        TypeDescriptor = TypeData.FromType(typeof(double));
        DeclaringType = TypeData.FromType(typeof(MyStep));
        Writable = true;
        Readable = true;
    }

    public override void SetValue(object owner, object value)
        => _values.GetOrCreateValue(owner).Value = value is double d ? d : 0.0;

    public override object GetValue(object owner)
        => _values.GetOrCreateValue(owner).Value;
}

[Display("MyStep")]
public class MyStep : TestStep
{
    public double Value { get; set; } = 42.0;
    public override void Run() => UpgradeVerdict(Verdict.Pass);
}

class Program
{
    static void Main()
    {
        PluginManager.SearchAsync().Wait();

        GC.Collect(2, GCCollectionMode.Aggressive, true, true);
        long baseline = GC.GetTotalMemory(true);
        Console.WriteLine($"Baseline: {baseline / (1024.0 * 1024):F2} MB");

        for (int i = 1; i <= 50; i++)
        {
            BuildAndExecute();

            if (i % 10 == 0)
            {
                GC.Collect(2, GCCollectionMode.Aggressive, true, true);
                GC.WaitForPendingFinalizers();
                GC.Collect(2, GCCollectionMode.Aggressive, true, true);
                long current = GC.GetTotalMemory(true);
                double growth = (current - baseline) / (1024.0 * 1024.0);
                Console.WriteLine($"After run {i,3}: {current / (1024.0 * 1024):F2} MB (growth: {growth:+0.00;-0.00} MB)");
            }
        }
    }

    [System.Runtime.CompilerServices.MethodImpl(
        System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
    static void BuildAndExecute()
    {
        var plan = new TestPlan();
        for (int i = 0; i < 500; i++)
        {
            var step = new MyStep();
            plan.ChildTestSteps.Add(step);
            DynamicMember.AddDynamicMember(step, new BareDynamicMember());
        }
        plan.Execute(Array.Empty<IResultListener>());
        // plan goes out of scope — should be GC'd, but never is
    }
}
```

**Expected**: Memory stays flat after warmup (~5-20 MB).
**Actual**: Memory grows ~15 MB/iteration indefinitely:

```
Baseline:       5.18 MB
After run  10: 115.00 MB (growth: +110.00 MB)
After run  20: 225.00 MB (growth: +220.00 MB)
After run  30: 340.00 MB (growth: +335.00 MB)
After run  40: 450.00 MB (growth: +445.00 MB)
After run  50: 560.00 MB (growth: +555.00 MB)
```

## Key observations

1. **No DynamicMembers → no leak** — Steps without `DynamicMember`s use shared `TypeData` instances as cache keys, which are interned and reused. Memory stays flat.
2. **DynamicMember without Execute → no leak** — Adding `DynamicMember`s and discarding the plan *without calling Execute* does not leak. The `DynamicMemberTypeDataProvider.dict2` (`ConditionalWeakTable`) correctly allows GC.
3. **Execute triggers it** — `GetSettingsLookup` is called during execution to resolve resources. This populates `membersLookup` with per-instance keys that are never evicted.
4. **Detach / RemoveDynamicMember does not help** — The cache retains the `DynamicTestStepTypeData` key even after the `DynamicMember` is removed from the step, since the key is already in the `ImmutableDictionary`.

## Impact

This leak affects any long-running OpenTap host that repeatedly loads and executes plans containing steps with `DynamicMember`s. Common scenarios:

- Automated test sequences that load → execute → discard plans in a loop
- CI/CD runners executing many plans per session
- Any plugin using the mixin/`DynamicMember` API (limit checking, parameterization, etc.)

Memory grows linearly and is never reclaimed until the process exits.

## Current Workaround

An `ITestPlanRunMonitor` that invalidates the cache's `currentObj` field via reflection after each plan execution, forcing the next `CheckClear()` to clear the `ImmutableDictionary`:

```csharp
using System;
using System.Reflection;
using OpenTap;

[Display("Dynamic Member Cache Monitor",
    Description: "Workaround for membersLookup memory leak with DynamicMembers.")]
public class DynamicMemberCacheMonitor : ComponentSettings<DynamicMemberCacheMonitor>,
                                          ITestPlanRunMonitor
{
    private static readonly FieldInfo? MembersLookupField;
    private static readonly FieldInfo? CurrentObjField;
    private static readonly PropertyInfo? CountProperty;
    private static readonly bool ReflectionAvailable;

    static DynamicMemberCacheMonitor()
    {
        try
        {
            var asm = typeof(TestStep).Assembly;
            var extType = asm.GetType("OpenTap.TestStepExtensions");
            MembersLookupField = extType?.GetField("membersLookup",
                BindingFlags.Static | BindingFlags.NonPublic);
            if (MembersLookupField != null)
            {
                var cache = MembersLookupField.GetValue(null);
                var cacheType = cache?.GetType();
                CurrentObjField = cacheType?.GetField("currentObj",
                    BindingFlags.Instance | BindingFlags.NonPublic);
                CountProperty = cacheType?.GetProperty("Count");
            }
            ReflectionAvailable = MembersLookupField != null
                               && CurrentObjField != null
                               && CountProperty != null;
        }
        catch { ReflectionAvailable = false; }
    }

    public void EnterTestPlanRun(TestPlanRun plan) { }

    public void ExitTestPlanRun(TestPlanRun plan) => ClearCache();

    public static void ClearCache()
    {
        if (!ReflectionAvailable) return;
        var cache = MembersLookupField!.GetValue(null);
        if (cache == null) return;
        CurrentObjField!.SetValue(cache, new object()); // invalidate
        CountProperty!.GetValue(cache);                  // trigger CheckClear()
    }
}
```

## Environment

- OpenTap 9.32.3
- .NET 8.0 / .NET 9.0
- Windows 11

rmadsen-ks:
Fixed the issue by not using the Cache class and instead using a more local fix using conditional weak table.

An additional benefit of conditional weak table is that we remove the need for invalidation.

1. Moved the core functionalityh into GetSettingsLookupNoCache.
2. Changed the caching strategy to use a layered conditional weak table relating to the dynamic types instead of a tuple as a key. The important insight here is that TypeData is not dynamic, so that has a very limited number of instances, while ITypeData is dynamic and can have an unlimited number of instances. Hence we use Dictionary for the TypeData and ConditionalWeakTable for ITypeData.

Verified that a leak occurs on main, but not in the feature branch.

Close #2318


alnlarsen (approved):
Code looks good, nice to get rid of some cache complexity

PR #2326, opened 2026-04-13
process step: enable AddToLog by default

### LINKED ISSUE #2324: Run Program step does not indicate missing command or syntax errors
The default settings for the Run Program step does not enable either the `Add To Log` option or the `Check Exit Code` option. This means that syntax errors in commands are not visible to the end user.

We can address this by always adding stderr output to the log. The current `Add To Log` option states that `the result of the query is added to the log`. stderr output is generally not considered a result, so I don't think it is a lie to include it.

alnlarsen:
The previous default behavior completely hides errors and warnings, which seems like a mistake. E.g. if the specified program is not installed, or did not parse arguments correctly, ProcessStep will pass without reporting any issues. The expected default behavior in this case would be that the step fails or errors.

I think there is a strong case for inverting the defaults; check exit codes unless explicitly disabled, and always forward output unless disabled.

Closes #2324


rmadsen-ks (approved):
Great!

PR #2328, opened 2026-04-14
Handle termination character in ScpiInstrument

### LINKED ISSUE #2327: ScpiInstrument incorrectly does a second Read on a SOCKET connection (or ASRL?) if the instrument returns the termination character as the only data
We have the case where an instrument (Keysight E6416A, Multi Transceiver RF Test Set) can return only ```\n``` for the response to a query ```INST:PLAYBACK:IDS?```

This is the case where there are no playback sources defined.

Image

Instrument SCPI manual says:
```
10.4.8.1 INSTrument:PLAYBack:IDS?
Get the list of playback source ids
Syntax:
INSTrument:PLAYBack:IDS?
Description:
Get the list of playback source ids.
Example:
// get playback id’s
inst:playback:ids?
1,2
```

The problematic code is here:
https://github.com/opentap/opentap/blob/main/Engine/ScpiInstrument.cs#L646

alan-copeland-keysight:
Handle the case in a SOCKET connection where an instrument might not have any data to return, but needs to send a response to a query (e.g. returning a list that is empty). The Termination Character is recognized as the end of the data and no further reads are sent.

This should have similar behavior on a serial instrument since it also does not have an out-of-band end indicator (EOI).

This only takes effect when the Termination Character is enabled (default is disabled)

### DIFF
--- a/Engine/ScpiInstrument.cs
+++ b/Engine/ScpiInstrument.cs
@@ -643,6 +643,10 @@ namespace OpenTap
 
                     return text;
                 }
+                else if (firstChar == TerminationCharacter)
+                {
+                    return firstChar.ToString();
+                }
                 else
                 {
                     string theRest = LockRetry(() => ReadString());


rmadsen-ks (approved):
This makes a lot of sense. lets get it in.

PR #2331, opened 2026-04-16
added note relating to .NET Framework vs .NET 9

### LINKED ISSUE #2283: Support Image syntax in 'tap package install'
The image install syntax is very convenient for installing specific versions of different packages, but it wipes out the currently installed packages unless `--merge` is specified.

The package install action is older, and uses a different syntax which is not as flexible as the image install syntax.

I propose updating the `package install` action to support image syntax, so `tap package install Editor:9.28.2,CSV` is equivalent to installing the same image string with the `--merge` option

rmadsen-ks:
Close #2283

### DIFF
--- a/doc/Developer Guide/Introduction/Readme.md
+++ b/doc/Developer Guide/Introduction/Readme.md
@@ -9,7 +9,7 @@ Development requires the following software:
 
 - Visual Studio 2022 or above
 - OpenTAP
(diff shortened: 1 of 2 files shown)


alnlarsen (dismissed):
Looks great! But the PR says it closes itself. Can you link the doc issue?

PR #2332, opened 2026-04-16
translate break conditions

alnlarsen:
this makes Break Conditions translatable

Cloes https://github.com/opentap/opentap/issues/2333


rmadsen-ks (approved):
Great!

PR #2335, opened 2026-04-17
allow parameterizing [Output] properties onto parent scopes

### LINKED ISSUE #2266: Allow promoting Output to parent scope
Currently, [Output] from test steps is only accessible within the subtree where the output is relevant. This makes sense because the availability of an output is hard to determine outside of the local scope. However, it also prevents some very simple usecases. Consider the following pattern for example:

```
Retry Step # Retries `n` times until child step succeeds
  Measure Step # Performs a measurement which can fail
Log Result Step # Does something interesting with the output from Measure Step
```

Right now, there is no obvious way to get the output from `Measure Step` to `Log Result Step`.

With this feature, it should be possible to simply parameterize the output from `Measure Step` onto a parent scope. This should probably follow the same rules as parameterization.

rmadsen-ks:
Outputs typically have non-public setters, which caused them to fail the Writable check used in parameterization and serialization. This made it impossible to expose an output at a parent scope so it could be consumed by sibling steps (e.g. Retry > Measure, then Log Result Step).

- IsValidParameter: accept [Output] members even when not writable
- ParameterMemberData: treat parameter as writable for outputs; SetValue is a safe no-op for non-writable inner members
- Serializer: emit Parameter/Scope attributes on parameterized outputs without serializing the value (outputs are written by Run)
- Deserializer: match non-writable outputs with a Parameter= attribute so the parameterization is restored

Close #2266

### DIFF
--- a/Engine.UnitTests/ScopeParametersTest.cs
+++ b/Engine.UnitTests/ScopeParametersTest.cs
@@ -1025,5 +1025,62 @@ namespace OpenTap.UnitTests
             Assert.AreEqual(123.0, consumer2.SeenValue,
                 "The consumer did not observe the value from the parameterized output after round-tripping through the serializer.");
         }
+        
+        class SelectOutputFromUserInput : IUserInputInterface, IUserInterface
+        {
+            public bool WasInvoked;
+            public string SelectName { get; set; }
+            public void RequestUserInput(object dataObject, TimeSpan Timeout, bool modal)
+            {
+                var datas = AnnotationCollection.Annotate(dataObject);
+                var selectedName = datas.GetMember("Output");
+                var avail = selectedName.Get<IAvailableValuesAnnotationProxy>();
+                Assert.That(avail.AvailableValues.Count(), Is.EqualTo(1));
+                WasInvoked = true;
+            }
+
+            public void NotifyChanged(object obj, string property) { }
+        }
+
+        [TestCase(IconNames.ParameterizeOnParent)]
+        [TestCase(IconNames.ParameterizeOnTestPlan)]
+        public void TestCanParameterizeOutput(string scope)
+        {
+            using var _ = Session.Create(SessionOptions.OverlayComponentSettings);
+            var singleInput = new SelectOutputFromUserInput();
+            UserInput.SetInterface(singleInput);
+            
+            var plan = new TestPlan();
+            var parentStep = new SequenceStep();
+            var outputStep = new OutputProducerStep();
+            var inputStep = new OutputConsumerStep();
+
+            plan.ChildTestSteps.Add(parentStep);
+            parentStep.ChildTestSteps.Add(outputStep);
+            plan.ChildTestSteps.Add(inputStep);
+
+            var a = AnnotationCollection.Annotate(outputStep);
+            var oMem = a.GetMember(nameof(outputStep.Measurement));
+            var menu = oMem.Get<MenuAnnotation>().MenuItems.ToLookup(m => m.Get<IIconAnnotation>().IconName);
+            Assert.That(menu[IconNames.ParameterizeOnParent].Count(), Is.EqualTo(1));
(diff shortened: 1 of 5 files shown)


Engine.UnitTests/ScopeParametersTest.cs
```diff
@@ -902,6 +902,105 @@ public void DynamicTypeDataEqualityTest()
...
+            Assert.IsInstanceOf<ParameterMemberData>(param2);
+
+            var outputMember2 = TypeData.GetTypeData(producer2).GetMember(nameof(producer2.Measurement));
+            Assert.IsTrue(DynamicMember.IsParameterized(outputMember2, producer2),
+                "The output was not restored as parameterized after deserialization.");
+
+            // After running, the parameter should reflect the output value of the producer.
+            producer2.Run();
+            Assert.AreEqual(42.0, (double)param2.GetValue(seq2));
+        }
```
alnlarsen, line 1027: Tests are incomplete.

An OutputConsumerStep is defined, but it is not used.

The tests need to close the loop by actually verifying the output is assigned to inputs.

Engine.UnitTests/ScopeParametersTest.cs
```diff
@@ -902,6 +902,105 @@ public void DynamicTypeDataEqualityTest()
...
+        /// <summary>
+        /// Reads a value via a public (writable) property - can be connected to an output via an
+        /// input/output relation, or simply have the value set by a parameter.
+        /// </summary>
+        public class OutputConsumerStep : TestStep
+        {
+            public double Value { get; set; }
+            public override void Run()
+            {
+                Log.Info("Consumed value: {0}", Value);
```
alnlarsen: this log goes nowhere?

Engine.UnitTests/ScopeParametersTest.cs
```diff
@@ -902,6 +902,105 @@ public void DynamicTypeDataEqualityTest()
...
+
+            // Should be considered a valid parameter now despite Writable==false.
+            Assert.IsTrue(ParameterManager.CanParameter(outputMember, new ITestStepParent[] { producer }));
+
+            const string paramName = "Exposed Measurement";
+            var parameter = outputMember.Parameterize(seq, producer, paramName);
+            Assert.IsNotNull(parameter);
+
+            // GetValue on the parameter should reflect the output value on the source.
+            producer.Run();
```
alnlarsen: We need to actually run the test plan for this to make sense. producer.Run() only verifies that C# methods work.

Engine.UnitTests/ScopeParametersTest.cs
```diff
@@ -902,6 +902,105 @@ public void DynamicTypeDataEqualityTest()
...
+            var param2 = TypeData.GetTypeData(seq2).GetMember(paramName);
+            Assert.IsNotNull(param2, "Parameter was not preserved during deserialization.");
+            Assert.IsInstanceOf<ParameterMemberData>(param2);
+
+            var outputMember2 = TypeData.GetTypeData(producer2).GetMember(nameof(producer2.Measurement));
+            Assert.IsTrue(DynamicMember.IsParameterized(outputMember2, producer2),
+                "The output was not restored as parameterized after deserialization.");
+
+            // After running, the parameter should reflect the output value of the producer.
+            producer2.Run();
```
alnlarsen: same problem

alnlarsen (changes requested):
Tests are low quality

alnlarsen (changes requested):
Other missing tests:

1. Verify the parameterize options are available on the Ouput member ("Parameterize...", "Parameterize on Parent", "Parameterize on Test Plan")
2. Verify that output can be parameterized on the *test plan* since this is the key reason the feature is needed

alnlarsen (dismissed):
Looks good to merge

alnlarsen (approved):
Looks awesome!

PR #2338, opened 2026-04-21
fix global test fixture

alnlarsen:
The old fixture was never invoked because it using static methods and was not in the right namespace.


rmadsen-ks (approved):
Great! Only one test fails on my machine

PR #2345, opened 2026-04-28
Add parallel sweep step

MatCapSen:
This is the fusion of Parallel and Sweep Parameter Range in a single step.

Disclaimer: part of this step was written by a LLM but it was reviewed and tested by a human.

### DIFF
--- a/BasicSteps/SweepParameterParallelStep.cs
+++ b/BasicSteps/SweepParameterParallelStep.cs
@@ -1,0 +1,303 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+
+namespace OpenTap.Plugins.BasicSteps
+{
+    [AllowAnyChild]
+    [Display(
+        "Sweep Parameter Parallel",
+        "Runs child steps for each sweep value simultaneously in separate threads.",
+        "Flow Control"
+    )]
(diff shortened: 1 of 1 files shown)


rmadsen-ks:
Hi @MatCapSen, thanks for the commit, this is interesting. I have been thinking about creating something like this, but the fact that you need to clone all the steps for each thread means supporting it from a UI perspective is non-trivial. For example, we want UI to show the progress of each step, but now each step is running in multiple threads in parallel, so its suddenly both running and completed at the same time.

What is your use case? How did you test it?

MatCapSen:
> Hi @MatCapSen, thanks for the commit, this is interesting. I have been thinking about creating something like this, but the fact that you need to clone all the steps for each thread means supporting it from a UI perspective is non-trivial. For example, we want UI to show the progress of each step, but now each step is running in multiple threads in parallel, so its suddenly both running and completed at the same time.
> 
> What is your use case? How did you test it?

Yes it's true that it's hard to represent for UI. In my opinion it would require additional UI developments to be able to visualize properly (maybe parallel lines in the flow). In the mean time I have opted for the simplest solution : display only the steps of one thread (the first iteration). In my use case this is sufficient as all iterations take roughly the same time.

My use case is testing several DUTs. I would usually sweep through DUTs but in this case I have one instrument per DUTs so I can parallelize it. I'm testing it on a real test bench testing PCBs.

PR #2346, opened 2026-04-28
2294 Multiple groups for result parameters

rmadsen-ks:
Added that more that 1 groups gets jointed with / for result parameters.

Also added an extremely minor cleanup in ifstep (dead code).

Close #2294

### DIFF
--- a/Engine/ResultListener.cs
+++ b/Engine/ResultListener.cs
@@ -574,7 +574,7 @@ namespace OpenTap
                         group = display.Group[0].Trim();
                     else if (display.Group.Length == 0)
                         group = "";
-                    else group = string.Join(" / ", display.Group.Select(grp => grp.Trim()));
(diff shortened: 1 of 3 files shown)


Engine/ResultListener.cs
```diff
@@ -570,7 +570,11 @@ class Box
                     var name = display.Name.Trim();
                     string group = "";
 
-                    if (display.Group.Length == 1) group = display.Group[0].Trim();
+                    if (display.Group.Length == 1) 
+                        group = display.Group[0].Trim();
+                    else if (display.Group.Length == 0)
+                        group = "";
+                    else group = string.Join(" / ", display.Group.Select(grp => grp.Trim()));
```
alnlarsen: Let's use '\' for consistency with DisplayAttribute

alnlarsen (dismissed):
Looks fine

alnlarsen (approved):
Better!

PR #2362, opened 2026-05-11
2352 2351 output parameter improvements

### LINKED ISSUE #2351: Handle output parameters in the CLI
Currently output (get-only) parameters are shown as writeable external parameters. 

Instead we should should them as read only and not allow setting them when the test plan runs. In additional, they should be shown in the log when the test plan has completed.

### LINKED ISSUE #2352: Handle multiple parameterized outputs.
When multiple outputs are parameterized i believe only the first is considered when getting the value.

I think we should either disallow parameterizing multiple outputs unto the same parameter or add some rule like latest value is the right one (However, i think that might be impossible).

rmadsen-ks:
- Excluded read-only parameters from being shown in the CLI.
- Added checks for read-only and output attribute for paramerization related code.
- Modified test to add error cases.

Close #2352, #2351

### DIFF
--- a/Engine/DynamicMemberTypeDataProvider.cs
+++ b/Engine/DynamicMemberTypeDataProvider.cs
@@ -301,6 +301,8 @@ namespace OpenTap
         {
             if (source == newSource && newMember == member)
                 throw new Exception("Member is already parameterized.");
+            if (Writable == false && member.HasAttribute<OutputAttribute>())
+                throw new Exception("Output parameters cannot be combined.");
             
             parameterMembers = parameterMembers.Add(newSource, newMember);
         }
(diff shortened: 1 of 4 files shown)


Engine/DynamicMemberTypeDataProvider.cs
```diff
@@ -301,6 +301,8 @@ internal void AddAdditionalMember(object newSource, IMemberData newMember)
         {
             if (source == newSource && newMember == member)
                 throw new Exception("Member is already parameterized.");
+            if (Writable == false && member.HasAttribute<OutputAttribute>())
+                throw new Exception("Output parameters cannot be combined.");
```
alnlarsen: I think we call this 'merged' other places in the UI? Correct me if I'm wrong
```suggestion
                throw new Exception("Output parameters cannot be merged.");
```

alnlarsen (changes requested):
Looks great, testing this!

alnlarsen (changes requested):
This doesn't work. In KS8400, I am still able to merge output parameters.

Also (unrelated to this PR), I am able to edit the value of [Output] members after parameterizing.

I tested this by creating a Delay step and adding an [Output] Number Mixin.

alnlarsen (approved):
Works perfectly! Disregard my comment about mixin outputs.

PR #2364, opened 2026-05-12
safe testplan save mechanism

### LINKED ISSUE #2365: Crashing during TestPlan.Save() can cause test plan corruption
We can prevent this by saving to a temporary path before overwriting the target file.

alnlarsen:
save to a temporary path before overwriting the target file to ensure crashes during serialization does not replace the target file with a partially written testplan


rmadsen-ks (approved):
This is great!

PR #2374, opened 2026-05-26
include image fallback hash in hash calculation

### LINKED ISSUE #2373: Image hash collision when packages lie about their hash
This can happen if a user unzips an OpenTAP package, modifies the content, and repackages it without updating hashes.

This is user error because it only happens when packages are zipped without using OpenTAP tooling, but it is easy for us to fix by taking package name, os, version, and architecture into consideration when computing the hash

alnlarsen:
this fixes image hash collisions when packages lie about their content hash

Closes #2373

### DIFF
--- a/Package/Image/ImageIdentifier.cs
+++ b/Package/Image/ImageIdentifier.cs
@@ -80,21 +80,33 @@ namespace OpenTap.Package
             List<string> packageHashes = new List<string>();
             foreach (PackageDef pkg in packageList)
             {
+                /* Always include the base hash in the hash calculation.
+                 * The values the base hash is derived from are normally included in pkg.Hash and pkg.ComputeHash().
+                 * This guards against hash collisions if a package contains an incorrect hash.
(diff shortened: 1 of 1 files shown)


Package/Image/ImageIdentifier.cs
```diff
@@ -80,21 +80,24 @@ private string CalculateId(IEnumerable<PackageDef> packageList)
             List<string> packageHashes = new List<string>();
             foreach (PackageDef pkg in packageList)
             {
+                var baseHash = $"{pkg.Name} {pkg.Version} {pkg.Architecture} {pkg.OS}";
```
rmadsen-ks, line 92: can you add a comment about why the baseHash is needed. Seems like its already included in pkg.Hash.

rmadsen-ks (changes requested):
Can you add a comment about why t

Oeh1995 (approved):

PR #2375, opened 2026-05-26
Workqueue deadlock

### LINKED ISSUE #2371: Deadlock in post run of composite runs
There is a very narrow window between a work queue for result listeners being disposed and new work items being enqueued where a work item can be assigned to a disposed worker thread, causing the run to hang indefinitely.

alnlarsen:
this fixes a race condition which occurs when a disposed workqueue is
reused in composite runs.

Closes #2371


rmadsen-ks (approved):
Great!

PR #2380, opened 2026-05-29
Add session log env vars and Log Files documentation

### LINKED ISSUE #2359: User would like more control over session logs
The user would like to be able to configure some more behaviors of SessionLog handling. The thing the user would like to be able to control:
1. Session log file length. They have some very long session logs. They would like the session logs to rollover to a new file if the file gets too large. 
2. Max number of session logs that are kept. For compliance, they would like to keep ALL session logs

I think this is an important concept especially when OpenTAP is used as a component in large solutions.

rmadsen-ks:
- Make session log limits configurable via OPENTAP_SESSION_LOG_MAX_FILES, OPENTAP_SESSION_LOG_MAX_TOTAL_SIZE and OPENTAP_SESSION_LOG_MAX_FILE_SIZE
- Increased max file count from 20 to 100.
- Increased max  total size from 2GB to 10GB.

For most users, they will hit the 100 file limit long before the 10GB total size limit.  The 10GB total size limit will mostly be used for very long running sessions that also generates a lot of log messages.

Close #2359

### DIFF
--- a/Engine.UnitTests/SessionLogTest.cs
+++ b/Engine.UnitTests/SessionLogTest.cs
@@ -52,10 +52,12 @@ namespace OpenTap.Engine.UnitTests
             
             var prev = SessionLogs.MaxTotalSizeOfSessionLogFiles;
             var prevLogFileMaxSize = SessionLogs.LogFileMaxSize;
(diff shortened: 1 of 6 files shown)


alnlarsen (approved):
Looks great

PR #2384, opened 2026-06-03
Avoid loading step type in call to AllowChild

### LINKED ISSUE #2383: TestStepList.AllowChild always loads the child type
This is not really necessary because we can check during the plugin scan whether or not a type defines [AllowAsChildIn].

alnlarsen:
By tracking whether or not a step has the AllowAsChildInAttribute defined, we can selectively load test steps only if it is actually required to determine if a step can be inserted as a child step of another step.

If a step actually defines the attribute it is non-trivial to determine without actually loading the type, so this optimization only applies when a step does not define the attribute, which is the vast majority of cases.

Closes #2383

### DIFF
--- a/Engine/PluginSearcher.cs
+++ b/Engine/PluginSearcher.cs
@@ -829,6 +829,9 @@ namespace OpenTap
                         plugin.Display = attrInstance;
                     }
                         break;
+                    case "OpenTap.AllowAsChildInAttribute":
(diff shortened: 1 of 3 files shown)


rmadsen-ks:
It seems this breaks some logic. Maybe we should wait with this for 9.34?

rmadsen-ks (approved):

PR #2390, opened 2026-06-09
fix: break conditions not triggered for steps using defer

### LINKED ISSUE #2387: Break Conditions don't work on test steps that use Results.Defer()
**Describe the bug**

When a test step uses Results.Defer(), the break condition setting doesn't work on it. The break condition is ignored and the test plan proceeds as normal.

Throwing an Exception after the Results.Defer() also doesn't stop the test plan.

The same behaviour is also observed if a test step mixin uses results.Defer.

 - OS: Windows 11
 - OpenTAP Version 9.33.0

**To Reproduce**

1. Using the following test step:
    The test step sets its break condition to BreakOnFail by default.
    The Run() sets its Verdict to Fail.
    The ResultsDefer property can be set to true to run a dummy Results.Defer() call.

        public class MyFailStep : TestStep
        {
            public MyFailStep() : base()
            {
                BreakConditionProperty.SetBreakCondition(this, BreakCondition.BreakOnFail | BreakCondition.BreakOnError);
            }

            public bool ResultsDefer { get; set; } = true;
        
            public override void Run()
            {
                this.Verdict = Verdict.Fail;
        
                if (ResultsDefer)
                {
                    this.Results.Defer(() =>
                    {
                        TapThread.Sleep(100);
                    });
                }
            }
        }

2. With a test plan like so:
- MyFailStep
- Delay Step 1s
- Delay Step 0.1s
- Delay Step 0.1s

Image

**Actual behavior**

Running the test plan with the ResultsDefer property set to true will run the plan fully.
With the option disabled (no Results.Defer), the test plan breaks at the first step as expected.

**Expected behavior**

The test plan should break instead of running fully.

alnlarsen:
this happened because Step verdicts are propagated to StepRun objects in the Defer queue if there are pending defers, meaning it is not yet set when we check for break conditions.

We can fix this by checking against the verdict on the Step object instead.

This also highlights another problem though, namely that we don't check break conditions if the verdict is upgraded during Defer(), but that seems like much more of an edge case than break conditions being completely ignored when using Defer.

Closes #2387

### DIFF
--- a/Engine.UnitTests/BreakConditionTest.cs
+++ b/Engine.UnitTests/BreakConditionTest.cs
@@ -337,6 +337,51 @@ namespace OpenTap.Engine.UnitTests
             }
         }
 
+        public class TestStepThatFails : TestStep
+        {
+            public bool ResultsDefer { get; set; } = true;
+
+            public override void Run()
+            {
+                UpgradeVerdict(Verdict.Fail);
+                if (ResultsDefer) 
+                {
+                    Results.Defer(Sleep);
+                }
+            }
+
+            static void Sleep()
+            {
+                TapThread.Sleep(1000);
+            }
+        }
+
+        public class TestStepThatSucceeds : TestStep
+        {
+            internal bool DidExecute = false;
+            public override void Run()
+            {
+                UpgradeVerdict(Verdict.Pass);
+                DidExecute = true;
+            }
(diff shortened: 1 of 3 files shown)


rmadsen-ks (approved):
Great!

PR #2397, opened 2026-06-15
Allow package actions to detect that its a non-interactive install.

### LINKED ISSUE #2435: Allow package actions to detect that its a non-interactive install
This would allow packages with installers to better be able to serve UIs when it makes sense.

rmadsen-ks:
Added documentation about OpenTAP environment variables.

Also added some debug log info in tap package create

Close #2435


alnlarsen (approved):
Looks great to me! Nice to document our variables

PR #2406, opened 2026-06-19
Switch to PosixFileLock implementation on MacOS

### LINKED ISSUE #2405: FileLock can brick an installation on MacOS if the process crashes
This happens because the MacOS FileLock implementation relies on `FileOptions.DeleteOnClose` to clean up the lock. Unfortunately, because this is a Windows kernel feature, it is implemented in software on MacOS and Linux, meaning that the file is not deleted if the process exits abnormally.

I suggest switching MacOS to use the Linux FileLock implementation instead, which uses `flock` to acquire the lock instead.

alnlarsen:
* Add error value checks to flock() and open() calls
* Make MacOS use Posix FileLock
* Delete custom MacOS FileLock
* Stop deleting .lock file on dispose, since it allows simultaneous lock acqusition
* Make LockingPackageAction actually acquire lock after initial failure
* Delete flaky unittest

Closes #2405

### DIFF
--- a/Package.UnitTests/CliTests.cs
+++ b/Package.UnitTests/CliTests.cs
@@ -522,18 +522,6 @@ namespace OpenTap.Package.UnitTests
         }
 
         [Test]
-        public void InstallFromRepoTest()
-        {
-            int exitCode;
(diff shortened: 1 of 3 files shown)


rmadsen-ks (dismissed):
Great!

PR #2414, opened 2026-07-01
Pages.yml fix plus CI cleanup

alnlarsen:
Various CI cleanup tasks

* Fix pages.yml not building since Github deprecated node v20
* Remove checks related to our old gitlab docker images
* Deleted pages build job from the main ci.yml since it doesn't work, and should be triggered manually anyway


Oeh1995 (approved):
looks good

PR #2416, opened 2026-07-01
2387 fix defer v2

### LINKED ISSUE #2430: Break Conditions don't work in RunChildSteps child test steps that use Results.Defer()
This was partially addressed in #2387, but that was missing some handling in other code paths, namely RunChildSteps() used in e.g. Sequence step.

See #2416

alnlarsen:
this fixes an issue with breaking from e.g. Sequence step, which uses
RunChildSteps() for executing child steps observed by @brandon-tan-keysight 

Closes https://github.com/opentap/opentap/issues/2430


rmadsen-ks (approved):
Looks great!

Issue #2377, opened 2026-05-27
Better Plugin Documentation


rmadsen-ks:
markdown is a familiar format to most developers. Its convenient and quick to write and is easily accesible by humans and AIs alike. 

I suggest we make markdown the common standard for all help material related to opentap. This is not really a suggestion to add any new code to OpenTAP, instead I want to form a convention that can be added to our templates, documentation and existing plugins.

A plugin package should have a Readme.md file that introduces the user to the plugin then subfolders with additional markdown files can be used as topics digging into details. 

A panel plugin could be implemented for KS8400 to show the help for all installed plugins.

The package repository package could support the format and allow users to read the documentation for each package in the browser.

## Outstanding questions

- Should it be possible to differentiate between user doc and developer doc or could this just be by convention?
- Should there be support for an optional(maybe not) /doc folder inside the package that contains all the documentation vs a /Readme.md file.

bogdan-achim:
Verified with OpenTAP v9.35.0-RC.3 & KS8400 v9.35.0-RC.3

PR #2422, opened 2026-07-22
Plugin documentation

frederikja163:
Related #2377 

Not sure if this fixes the issue or if there is more to be done. But it serves as a starting point for plugin documentation to have discussions about.


doc/Developer Guide/Plugin Packaging and Versioning/Readme.md
```diff
@@ -206,6 +206,23 @@ A package can also include a package icon. The **File** element inside the confi
 </Package>
 ```
 
+##### Package Readme
+A package can include a `README.md` file describing what the package does and how to use it. By convention, the file is placed in the package's folder (`Packages/<PackageName>/README.md`) and added to the package like any other payload file:
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<Package Name="MyPlugin" xmlns="http://opentap.io/schemas/package" InfoLink="http://myplugin.com"
```
alnlarsen: I think `example.com` is a more common placeholder.

alnlarsen (changes requested):
The suggested template is not going to build currently.

The SourcePath attribute in the package.xml assumes there is a file named `README.md` in bin/Release during package create.

Furthermore, the README.md one would typically put in the project root is not necessarily the README one would ship to end users in a packaged product.

I think it's good to add a project README by default, but it should be separate from the end user readme.

I would suggest:
1. Keep the project readme you added
2. Add another end user readme in e.g. `./doc/README.md`
3. Update the .csproj template to copy doc/ to output during build
4. Update the package.xml to bundle `doc/**`

I think OpenTAP already has the required globbing functionality to package doc/** into Packages/PackageName/doc/**, but we should verify that recursive directories are handled correctly.

frederikja163:
> Furthermore, the README.md one would typically put in the project root is not necessarily the README one would ship to end users in a packaged product.
I think it's good to add a project README by default, but it should be separate from the end user readme.

The readme.md i added here i supposed to be the readme.md for the end-user. But i guess it doesnt really read that way.

The project template creates a csproj not a solution. I dont think it is that common to have a README.md in the csproj file itself.

Lets discuss what the right solution here is.

rmadsen-ks (approved):
Great!

alnlarsen (approved):
Looks great!

PR #2426, opened 2026-08-03
2425 SecureString Capped To 16 bytes fix

### LINKED ISSUE #2425: SecureString capped to 16 bytes max size
This is caused by CryptoStream reader only reading each 16 byte block at a time instead of filling the output.

rmadsen-ks:
Close #2425


alnlarsen (dismissed):
Looks great!

alnlarsen (approved):
still good

PR #2436, opened 2026-08-07
Write warnings to stderr and respect OPENTAP_NO_UPDATE_CHECK

### LINKED ISSUE #2434: Warnings should be written to stderr and not stdout
Currently warnings are written to stdout and this may break some tools that try to parse data from stdout. For example MCP stdio SSE

rmadsen-ks:
- ConsoleTraceListener now writes warnings to stderr like errors, so with --quiet nothing is written to stdout.
- The update check can be disabled entirely with OPENTAP_NO_UPDATE_CHECK. The check is done before the environment variable is set for child processes, otherwise it would always be suppressed.
- Add unit tests verifying that --quiet writes nothing to stdout, and that a duplicate assembly of a different version (e.g. Newtonsoft.Json) produces a warning on stderr.

Close #2434


alnlarsen (approved):
Looks great

PR #2441, opened 2026-08-10
Fix timer cancellation race in TimeoutOperation

### LINKED ISSUE #2442: Race in TimeoutOperation
In particular, short waits sometimes don't trigger the wait() function has multiple exit conditions on the same timer, only one of which trigger the timeout callback.

This causes the TestEngine tests to sometimes hang, so let's fix it.

alnlarsen:
Due to the dipose token in TimeoutOperation using the same timeout as the actual wait timeout (the condition for calling the timeout function), it was always a coin-toss whether or not the timeout would actually be invoked for short waits.

I suspect this was never flagged because the test is disabled on Linux and MacOS, both of which have a much lower os scheduler granularity compared to Windows' default granularity of ~15ms.

This led to spurious unittest hangs when the race *did* occur on Windows.

### DIFF
--- a/Engine.UnitTests/ReflectionHelperTest.cs
+++ b/Engine.UnitTests/ReflectionHelperTest.cs
@@ -43,7 +43,6 @@ namespace OpenTap.Engine.UnitTests
         }
 
         [Test]
-        [Platform(Exclude="Unix,Linux,MacOsX")]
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Great!

PR #2444, opened 2026-08-11
fix globbing in package xml

### LINKED ISSUE #2443: Package globbing `**` does not work
**Describe the bug**
The package.xml feature with including multiple files using the `**` globbing pattern does not work. It should be recursive according to: https://doc.opentap.io/Developer%20Guide/Plugin%20Packaging%20and%20Versioning/Readme.html#wildcards

alnlarsen:
The previous globbing implementation never really worked. This fix fetches all relevant files and lets the glob implementation decide for each file, instead of expecting Directories.GetFiles() to understand the glob in the file argument.

Closes #2443

### DIFF
--- a/Package.UnitTests/PackageGlobTests.cs
+++ b/Package.UnitTests/PackageGlobTests.cs
@@ -1,0 +1,106 @@
+using System;
+using System.IO;
+using System.Linq;
+using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


rmadsen-ks (approved):
Great

PR #2447, opened 2026-08-12
Fix translation of BreakCondition dynamic properties

### LINKED ISSUE #2446: ComponentSettings startup deadlock
This happens in some cases when component settings are accessed during early startup.

It seems to specifically be a conflict involving TestStepTypeData static constructors, which uses the translation API, which needs EngineSettings to finish deserializing, which can deadlock if something else is also attempting to access component settings.

We should be able to whack this specific mole by lazy-initializing the TestStepTypeData typedatas, but it would probably be good to revisit the component settings locking mechanism at some point.

alnlarsen:
This fixes a startup deadlock observed with some ComponentSettings access patterns during initialization

Closes #2446

### DIFF
--- a/Engine/DynamicMemberTypeDataProvider.cs
+++ b/Engine/DynamicMemberTypeDataProvider.cs
@@ -14,11 +14,16 @@ namespace OpenTap
 {
     internal class BreakConditionStrings : IStringLocalizer
     {
+        public const string DisplayNameNeutral = "Break Conditions";
+        public const string StepDescriptionNeutral = "When enabled, specify new break conditions. When disabled conditions are inherited from the parent test step, test plan, or engine settings.";
+        public const string GroupNeutral = "Common";
+        public const string TestPlanDescriptionNeutral = "When enabled, specify new break conditions. When disabled conditions are inherited from the engine settings.";
+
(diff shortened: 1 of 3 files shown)


Engine/DynamicMemberTypeDataProvider.cs
```diff
@@ -750,6 +755,20 @@ public override int GetHashCode()
...
+        /* specialized DisplayAttribute implementation which can be dynamically translated. */
+        class TranslatableDisplayAttribute((string raw, StringGetter translate) name, (string raw, StringGetter translate) description, (string raw, StringGetter translate) group, double order) : DisplayAttribute(name.raw, description.raw, group.raw, order), ITranslatableDisplayAttribute
+        {
+            private readonly StringGetter name = name.translate;
+            private readonly StringGetter description = description.translate;
+            private readonly StringGetter group = group.translate;
+
+            public DisplayAttribute Translate()
+            {
+                return new DisplayAttribute(name(), description(), group(), Order);
```
rmadsen-ks: Lets add the neutral translation here as well.

rmadsen-ks (approved):

PR #2451, opened 2026-08-13
2450 make PackageDef constructor public

### LINKED ISSUE #2450: Add public constructor for PackageDef
There's not really a good reason to not allow constructing PackageDef, and it leads to weird workarounds in plugins such as:

```csharp
PackageDef pkg;
var xml = @"<?xml version=""1.0"" encoding=""utf-8""?>
                <Package xmlns=""http://opentap.io/schemas/package"">
                </Package>";

using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(xml)))
{
    pkg = PackageDef.FromXml(stream);
}
```

Since FromXml() is available, there's no reason not to make the constructor available.

alnlarsen:
Closes #2450

### DIFF
--- a/Package/PackageFile.cs
+++ b/Package/PackageFile.cs
@@ -529,7 +529,7 @@ namespace OpenTap.Package
         /// </summary>
         [XmlAttribute]
         [DefaultValue("tappackage")]
-        public string FileType { get; set; }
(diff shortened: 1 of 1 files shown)


StefanHolst (approved):
🙏🥳🚀

PR #2452, opened 2026-08-19
fix serializing lists of doubles losing precision (#2433)

### LINKED ISSUE #2433: Failed to serialize double min/max value in List correctly
**Describe the bug**
 
Unable to serialize the min/max value in List<double>.
When TapPlan is dirty, and performing a save operation, the information that being saved is different from what is expected.
 
- OS: Windows 11
- OpenTAP Version 9.32.3
 
**To Reproduce**
 
 
1 Create a custom test step with the following properties
   `public List<double> ListOfDouble { get; set; } = new List<double>() { double.MinValue, double.MaxValue };`
2 Override Run method to check is ListOfDouble value remains Min and Max values
```
var tempList = new List<double>() { double.MinValue, double.MaxValue };
if(ListOfDouble.SequenceEqual(tempList))
    UpgradeVerdict(Verdict.Pass);
else
{
    UpgradeVerdict(Verdict.Fail);
}
```
3 Open Editor
4 Add test step
5 Observed the list value in Test Step Settings show the correct min and max value
6 Run test step, the system return Verdict.Pass
7 Add "Generic SCPI Instrument"
8 Remove the added instrument
9 Click "OK"
10 Observed the value in Test Step Settings changed (This is unexpected)
11 Run test step, the system return Verdict.Fail
 
**Actual behavior**
Return Verdict.Fail when comparing List<double> after tap plan was changed.
 
**Expected behavior**
Return Verdict.Pass when comparing List<double> after tap plan was changed.
 
**Additional Context**
Some investigation findings, adding and removing instrument causes test plan becomes dirty, the value remains in memory will goes through a series of serializer and populate into step settings. 
Tested property double in serialization, no issues.
This issue only occurs when double.Min and double.Max were used.

rmadsen-ks:
Numeric collections are deserialized through NumberFormatter/BigFloat, and BigFloat converted to double as (double)Numerator / (double)Denominator. BigInteger to double conversions truncate instead of rounding to nearest, so values needing all 17 significant digits changed value, e.g. double.MaxValue became 1.7976931348623155E+308. Large numerators/denominators additionally overflowed to infinity, turning double.Epsilon into 0.

BigFloat.ToDouble now rounds to the nearest double (ties to even), exactly like double.Parse does, and handles subnormals, overflow and underflow.

Close #2433

### DIFF
--- a/Engine.UnitTests/ListSerialization.cs
+++ b/Engine.UnitTests/ListSerialization.cs
@@ -124,6 +124,39 @@ namespace OpenTap.Engine.UnitTests
             Assert.AreEqual(15, step.Dict[""]);
         }
 
+        public class DoubleListStep : TestStep
+        {
+            public List<double> ListOfDouble { get; set; } = new List<double> { double.MinValue, double.MaxValue };
+            public double[] ArrayOfDouble { get; set; } = { double.MinValue, double.MaxValue };
+            public override void Run()
+            {
+            }
+        }
+
+        /// <summary>
+        /// Lists and arrays of doubles must survive a serialization round trip without losing precision,
+        /// also for values that need all 17 significant digits (e.g. double.MinValue/MaxValue). Issue #2433.
+        /// </summary>
+        [Test]
+        public void DoubleListSerializationKeepsPrecision()
+        {
+            var values = new List<double>
+            {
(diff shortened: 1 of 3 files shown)


alnlarsen (commented):
The tests are good

alnlarsen (approved):
Looks good, let's merge

PR #2455, opened 2026-08-24
Fix race condition in RepeatStep with deferred child results

### LINKED ISSUE #2439: RepeatStep's Clear Verdict intermittently doesn't work when a child test step uses Results.Defer
**Describe the bug**

When a child test step uses Results.Defer(), the Repeat test step's verdict sometimes isn't cleared, even using the Clear Verdict option.

 - OS: Windows 10
 - OpenTAP Version 9.34.0

**To Reproduce**

Using the following test step:

    /// <summary>
    /// Resets a count to 0 in the PrePlanRun()
    /// Each Run() will increment the count
    /// Pass when count == 3
    /// </summary>
    public class DeferStep : TestStep
    {
        private int count = 0;

        public override void PrePlanRun()
        {
            base.PrePlanRun();
            this.count = 0;
        }

        public override void Run()
        {
            this.count++;

            if (this.count == 3)
                this.Verdict = Verdict.Pass;
            else
                this.Verdict = Verdict.Fail;

            this.Results.Defer(() =>
            {
                TapThread.Sleep(1);
            });
        }
    }

1. Create a test plan with a Repeat step, and the DeferStep as a child
2. Repeat set to: 
 - Until
 - Verdict Of DeferStep
 - Equals Pass
 - Clear Verdict = true
 - Max Count = 10

Image

3. Set the Test Plan repeat option to repeat until fail, with a count of 100 (Or just manually re-run the test plan)

Image

**Actual behavior**

The test plan repeat will never reach its count, because it always fails at some point. The failure is intermittent.

Image

**Expected behavior**

The test plan should always pass.

rmadsen-ks:
Close #2439

## Summary
Fixes a race condition in `RepeatStep` where verdicts from child steps with deferred results could be applied after the next iteration cleared the verdict, causing incorrect final verdicts.

## Key Changes
- **ResultProxy.cs**: Added `WaitForDeferredActions()` method to wait for all deferred actions and verdict propagations to complete. This method:
  - Enqueues a sentinel work item to the defer worker and waits for it to complete
  - Ensures all previously deferred actions have finished executing
  - Can only be called from the test step's thread
  - Safely handles cases where the defer worker has already been disposed

- **RepeatStep.cs**: Added call to `Results.WaitForDeferredActions()` after each iteration completes to ensure child step verdicts are fully propagated before proceeding to the next iteration or checking the current verdict

- **BasicStepsTest.cs**: Added comprehensive test case `RepeatUntilPassWithDeferredResults` that:
  - Creates a test step that defers work while setting its verdict
  - Runs the test plan 30 times to catch the intermittent race condition
  - Verifies that the repeat step ends with the correct verdict despite deferred result propagation

## Implementation Details
The fix ensures proper synchronization between verdict propagation and iteration clearing by waiting for the defer worker queue to drain after each iteration. This prevents a scenario where a deferred verdict update from iteration N could be applied after iteration N+1 cleared the verdict, resulting in an incorrect final verdict.

https://claude.ai/code/session_01BEhLPRM1zEQnhhn79jD1kL


Engine/ResultProxy.cs
```diff
@@ -379,6 +379,23 @@ internal void DeferNoCheck(Action action)
             });
         }
 
+        /// <summary>
+        /// Waits for the deferred actions of this test step run to complete.
+        /// This includes actions added with <see cref="Defer"/>, but also the propagation of verdicts from
+        /// child test steps which deferred their results.
+        /// This may only be called from the same thread as the test step.
+        /// </summary>
+        public void WaitForDeferredActions()
```
alnlarsen: @rmadsen-ks do we really want to add a public API for this?

BasicSteps/RepeatStep.cs
```diff
@@ -114,7 +114,11 @@ Verdict Iterate()
             {
                 r.WaitForCompletion();
             }
-            
+
+            // when a child step defers its results, the verdict propagation to this step is deferred too.
+            // it has to be done before the next iteration clears the verdict.
+            Results.WaitForDeferredActions();
```
alnlarsen: I checked out the branch and tested the following alternative:

```csharp
var evt = new ManualResetEventSlim();
Results.Defer(evt.Set);
evt.Wait();
```

This works without adding a new public API

alnlarsen (changes requested):
I'm not really sold on adding a new public API for this.

The only problem I can think of with my proposed alternative is if a deferred action enqueues new deferred work.

alnlarsen (dismissed):
Adding an API for this is fine. It's better to offer a way to do this than the defer hack I suggested, and much more efficient in the optimistic case where nothing is deferred.

alnlarsen (approved):
Looks great!

PR #2457, opened 2026-08-25
improve incremental build speeds

alnlarsen:
This significantly speeds up incremental builds in multi-project solutions.

I think this was required in the before-times, but since 3cf57f1045641bac7eb197d9b3ca578e5f6ae4b7 it is no longer necessary


rmadsen-ks (approved):
Great!

PR #2472, opened 2026-09-08
2471 ScpiInstrument Ctor avoid calling GetResourceManager

### LINKED ISSUE #2471: ScpiInstrument constructor should not call libvisa GetResourceManager
On some platforms this causes hangs

rmadsen-ks:
Close #2471

Removed the call to GetResourceManager from ScpiInstrument constructor.

This causes that the error about a missing visa resource is not shown upon xml load, but the benefit is that we do not risk corrupting settings just because visa is not installed or broken.

The warning will still be emittet when trying to open the resource.

In some cases the users are using simulated ScpiInstrument on systems that does not have a resource manager and in those situations they should also not get warnings and hangs.


alnlarsen (approved):
Looks great!

